Issue #226, opened 2015-04-01
Format expressions within string interpolation


zoechi:
``` dart
'Bad Request${(detail!='' ? ': ' : '')}$detail'
```

should be

``` dart
'Bad Request${(detail != '' ? ': ' : '')}$detail'
```

munificent:
Ugh, I knew someone was going to ask for this. :)

It's a rathole, but at the very least it should probably fix up whitespace.

zoechi:
At least such simple ifs should be supported. More complex expressions are usually calculated outside the string. I find it annoying for example to create a block body for a function instead of `=>` for such a simple case.

munificent:
I did some more poking around at this based on @BlackHC's PR, but wasn't able to find any straightforward way to get it working. The problem is basically:

1. We likely *don't* want to add soft splits inside an interpolated string. In most corpus's I've looked at, it looks much better to just let the line run long than add a newline in the middle of the interpolation.

2. But we don't want to disable *all* newlines either. Hard lines (think a function literal inside an interpolation) should certainly be preserved. And the user's hand-authored ones maybe.

So that leads us towards some weird mode where newlines are handled really idiosyncratically inside interpolation. It might be tractable, but I don't think it will be simple.

nex3:
I'd like to +1 this at least for cleaning up whitespace. I have a simple transform to get rid of calls to `await`, and I'm ending up with a lot of strings that look like `"(${ expression})"`, with an extra space after the beginning of the interpolation. Whitespace outside the expression seems like a pretty trivial case to clean up, and one that will help a lot of simple transformations like the one I'm doing.

PR #512, opened 2016-07-13
Format expressions within string interpolation #226

BlackHC:
See cl/126996083 for some examples in our codebase.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -1167,7 +1167,7 @@ class SourceVisitor implements AstVisitor {
   }
 
   visitInterpolationString(InterpolationString node) {
-    token(node.contents);
+    _writeStringLiteral(node.contents.lexeme, node.offset);
   }
 
   visitIsExpression(IsExpression node) {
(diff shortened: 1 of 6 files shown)


munificent:
I tried this out on a big corpus and my impression was:

* Places where it fixed whitespace within a line were greatly improved. Stuff like `"${a-1}"` -> `"${a - 1}"`.

* Any time it actually added a newline, though, was way worse.

Given the latter, I don't think we should take this as-is. I think we'd want to also disable inserting newlines within an interpolation. That will be a bit more work, but I think it's doable. I totally understand if that's a little too much for you to tackle, though. If not, I'll see if I can find some time to hack on it.

If nothing else, thank you for sending me the PR because it helped me work through some examples and think about how to approach it.

BlackHC:
Thanks! I was talking to David de Kloet about this the other day and our feeling was that all the places where it really has to add newlines should be refactored to extract the expression into a variable probably (that's nothing though that dart_style can do).

How can we prevent additional splits? Do I have to create a rule for this? If you could sketch out the steps, I could give it a go.

(This could block dart_ast_writer because I need to make dart_style independent of the source text :()

Thanks,
 Andreas

munificent:
Basically what I'm thinking is that ChunkBuilder will keep track of whether or not we're in a string interpolation. (Or, more specifically, it will keep track of the number of interpolations we're in, since they can nest.)

If we are in one, then `split()` just writes a space or nothing instead of writing a soft split. Hard splits will still get split, so things like this will still have newlines:

```dart
"${() {
  statement;
}}";
```

I'm not sure if this would actually work, but that's what I'd try first.

PR #553, opened 2016-11-10
Implement visitor method to be added in the analyzer package

bwilkerson:
We'll want to add a type for the parameter and remove the explicit type on the list literal after updating the dependency constraints.


bwilkerson:
@munificent

munificent:
Can you revert the change that reorders all the members? That's not exactly diff friendly.

bwilkerson:
Sorry about that! Sorting changes reverted.

munificent:
Thanks!

PR #554, opened 2016-11-10
Replace use of deprecated API

bwilkerson:
We plan on removing deprecated APIs as part of the next breaking change of analyzer, and dart_style is using one of those APIs. This PR proactively replaces the use of the API. (In other words, this isn't urgent.) Unfortunately, this requires narrowing the version constraints on analyzer.

I'm not sure how to handle the pubspec.lock file in a situation like this. Guidance welcome.

Also, there is a strong-mode error being reported. I'd be happy to fix it, either as part of this PR or separately. Let me know if you'd like me to do that.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -603,7 +603,7 @@ class SourceVisitor implements AstVisitor {
 
     token(node.rightParenthesis);
     space();
-    visit(node.libraryUri);
+    visit(node.uri);
   }
 
   visitConstructorDeclaration(ConstructorDeclaration node) {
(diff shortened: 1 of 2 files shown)


bwilkerson:
@munificent

munificent:
> I'm not sure how to handle the pubspec.lock file in a situation like this. Guidance welcome.

Just run `pub upgrade` on your checkout of dart_style and then add the changed lockfile to the pull request. :)

> Also, there is a strong-mode error being reported.

The one on `math.max`? Ah, yes. I had thought it was a mistake in the implementation of LUB and would go away, but now I see that it's because `node` is untyped. (Ugh, silent `dynamic`. <shakes fist>.) If you want to go ahead and annotate that, that would be rad.

Thanks!

munificent:
I went ahead and fixed it and merged it in with an updated lockfile.

bwilkerson:
@munificent Ping.

bwilkerson:
Sorry, I didn't see your reply.

munificent:
:)

PR #555, opened 2016-11-12
Add tests for new language features:

munificent:
- "=" as separate for named optional parameters. This was conveniently
  already doing the right thing.
- Asserts in constructor initializers. This was broken. :( Fixed it.

This patch works, but the tests don't pass unless you use an unreleased
version of analyzer. So, for now this won't be merged in until after
that's released and I upgrade the pubspec in this branch.


munificent:
cc @bwilkerson

PR #559, opened 2016-11-24
Update tests to use correct Windows file paths

whesse:
Uri.toFilePath() rather than Uri.path should be used.


whesse:
Two tests are failing on the Windows bot at https://build.chromium.org/p/client.dart.packages/builders/packages-windows-dart_style
This change should fix them.

PR #574, opened 2017-01-10
Update SourceVisitor to extend ThrowingAstVisitor

bwilkerson:
@munificent

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -23,7 +23,7 @@ import 'whitespace.dart';
 
 /// Visits every token of the AST and passes all of the relevant bits to a
 /// [ChunkBuilder].
-class SourceVisitor implements AstVisitor {
+class SourceVisitor extends ThrowingAstVisitor {
   /// The builder for the block that is currently being visited.
   ChunkBuilder builder;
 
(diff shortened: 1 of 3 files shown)


bwilkerson:
No, I forgot that I hadn't run the tests, but they all fail without output like
```
00:00 +0 -1: loading test/command_line_test.dart                                                   
  Failed to load "test/command_line_test.dart":
  Unable to spawn isolate: Unhandled exception:
  Could not import "package:test/src/backend/suite.dart" from "package:test/src/runner/remote_listener.dart": Cannot open file, path = '/Users/brianwilkerson/.pub-cache/hosted/pub.dartlang.org/test-0.12.17/lib/src/backend/suite.dart' (OS Error: Too many open files, errno = 24)
  null
```

bwilkerson:
@munificent Any idea what I'm doing wrong? If not, could you test and/or land these changes? I really need to get this in so I can make progress on 1.22.

munificent:
> @munificent Any idea what I'm doing wrong?

The test package tries to spawn a bunch of isolates and/or open a bunch of files really quickly and can easily overload OS X's low default max open file limit. You can tweak it in the OS or limit test's concurrency using "-j 1".

> If not, could you test and/or land these changes? I really need to get this in so I can make progress on 1.22.

I'll do it. :)

PR #583, opened 2017-02-03
Don't drop type argument on method calls with closure arguments.

### LINKED ISSUE #582: Eats method type parameters?
dartfmt 0.2.14 ([having trouble using 0.2.15](https://groups.google.com/a/dartlang.org/forum/#!topic/misc/ttUczfcdycc))

```
$ cat xxy.dart
void fn() {
  List<String> a;
  // NOTICE THE TYPE PARAMETER
  a.fold<int>(0, (memo, s) {
    return max(memo, s.length);
  });
}
$ dartfmt xxy.dart
void fn() {
  List<String> a;
  // NOTICE THE TYPE PARAMETER
  a.fold(0, (memo, s) {
    return max(memo, s.length);
  });
}
```

munificent:
Fix #582.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.2.16
+
+* Don't discard type arguments on method calls with closure arguments (#582).
+
 # 0.2.15
 
 * Support `covariant` modifier on methods.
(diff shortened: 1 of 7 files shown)


srawlins:
Love it 👍

PR #598, opened 2017-03-17
Handle parenthesized and immediately invoked functions better.

### LINKED ISSUE #566: Bad formatting for multiline IIFE arguments
When I pass an immediately-invoked function expression, for example to be able to use `async` to create a future that I pass in to an API, the formatter produces strange output. For example:

```dart
void main() {
  return new CancelableOperation.fromFuture(() async {
    // Make the underlying suite null so that the engine doesn't start running
    // it immediately.
    engine.suiteSink.add(loadSuite.changeSuite((runnerSuite) {
      engine.pause();
      return runnerSuite;
    }));

    var suite = await loadSuite.suite;
    if (canceled || suite == null) return;

    debugger = new _Debugger(engine, reporter, suite);
    await debugger.run();
  }(), onCancel: () {
    canceled = true;
    // Make sure the load test finishes so the engine can close.
    engine.resume();
    if (debugger != null) debugger.close();
  });
}
```

is formatted to:

```dart
void main() {
  return new CancelableOperation.fromFuture(
      () async {
        // Make the underlying suite null so that the engine doesn't start running
        // it immediately.
        engine.suiteSink.add(loadSuite.changeSuite((runnerSuite) {
          engine.pause();
          return runnerSuite;
        }));

        var suite = await loadSuite.suite;
        if (canceled || suite == null) return;

        debugger = new _Debugger(engine, reporter, suite);
        await debugger.run();
      }(), onCancel: () {
    canceled = true;
    // Make sure the load test finishes so the engine can close.
    engine.resume();
    if (debugger != null) debugger.close();
  });
}
```

munificent:
Fix #566.

### DIFF
--- a/example/format.dart
+++ b/example/format.dart
@@ -16,13 +16,34 @@ void main(List<String> args) {
   // Enable debugging so you can see some of the formatter's internal state.
   // Normal users do not do this.
   debug.traceChunkBuilder = true;
-  debug.traceLineWriter = true;
-  debug.traceSplitter = true;
+//  debug.traceLineWriter = true;
+//  debug.traceSplitter = true;
   debug.useAnsiColors = true;
 
-  runTest("regression/0000/0068.stmt", 14);
+  formatUnit("""
+void main() {
+  return new CancelableOperation.fromFuture(
+      () async {
+        // Make the underlying suite null so that the engine doesn't start running
+        // it immediately.
+        engine.suiteSink.add(loadSuite.changeSuite((runnerSuite) {
+          engine.pause();
+          return runnerSuite;
+        }));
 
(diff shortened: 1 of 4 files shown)


example/format.dart
```diff
@@ -16,13 +16,34 @@ void main(List<String> args) {
   // Enable debugging so you can see some of the formatter's internal state.
   // Normal users do not do this.
   debug.traceChunkBuilder = true;
-  debug.traceLineWriter = true;
-  debug.traceSplitter = true;
+//  debug.traceLineWriter = true;
```
kevmoo: Huh?

munificent: Oops, temp code! I'll fix.

PR #604, opened 2017-03-21
Correct report unchanged formatting when reading from stdin.

munificent:


munificent:
Super annoying to test, and not likely to break again, so... 🤷‍♂️

PR #626, opened 2017-05-09
Support URIs in part-of directives.

### LINKED ISSUE #615: Allow using URI strings in "part of" declarations
According to the 1.23 CHANGELOG this is now supported in the language, but dartfmt doesn't seem to be happy with it.

dartfmt 0.2.16 on the command line removes a "part of" URI from the file if I use one.

Formatting via the dart_style library seems to refuse to format files using a URI "part of" -- seen in this pull request for source_gen trying to change the generated code: https://github.com/dart-lang/source_gen/pull/138

munificent:
Fix #615.

### DIFF
--- a/test/splitting/function_types.unit
+++ b/test/splitting/function_types.unit
@@ -1,0 +1,103 @@
+40 columns                              |
+>>> many parameters
+Function(first, second, third, fourth, fifth, sixth, seventh, eighth, ninth,
+    tenth, eleventh, twelfth) f;
+<<<
+Function(
+    first,
+    second,
+    third,
+    fourth,
+    fifth,
+    sixth,
+    seventh,
+    eighth,
+    ninth,
+    tenth,
+    eleventh,
+    twelfth) f;
+>>> parameters fit but ) does not
+Function(int firstArgume, int argumentTo) f;
+<<<
+Function(
+    int firstArgume, int argumentTo) f;
+>>> keep mandatory and positional on same line
+Function(param, [foo, bar]) f;
+<<<
+Function(param, [foo, bar]) f;
+>>> keep mandatory and named on same line
+Function(param, {T foo, T bar}) f;
+<<<
+Function(param, {T foo, T bar}) f;
+>>> move just optional positional to second line even though all fit on second
+Function(parameter, [int foo, String bar]) f;
+<<<
+Function(parameter,
+    [int foo, String bar]) f;
+>>> move just named to second line even though all fit on second
+Function(parameter, {int foo, String bar}) f;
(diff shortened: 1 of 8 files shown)


test/splitting/function_types.unit
```diff
@@ -0,0 +1,103 @@
...
+Function(parameter, {int foo, String bar}) f;
+<<<
+Function(parameter,
+    {int foo, String bar}) f;
+>>> avoid splitting in function type parameters
+Function(parameter1, void printFn(param1, param2)) f;
+<<<
+Function(parameter1,
+    void printFn(param1, param2)) f;
+>>>
```
natebosch: [nit] no name on this one? what is interesting about having a shorter first parameter?

test/splitting/function_types.unit
```diff
@@ -0,0 +1,103 @@
...
+    parameter2,
+    parameter3]) f;
+>>> split between type and name
+Function(VerylongParameterType parameterName) f;
+<<<
+Function(
+    VerylongParameterType
+        parameterName) f;
+>>> split in function type and on variable name
+Function(VeryVeryVeryVeryLongParameterType) veryLongVariableName;
```
natebosch: [nit] this isn't a VeryVeryVeryVeryLongParameterType it's a veryVeryVeryVeryLongParameterName

test/splitting/function_types.unit
```diff
@@ -0,0 +1,103 @@
...
+<<<
+Function(
+        VeryVeryVeryVeryLongParameterType)
+    veryLongVariableName;
+>>> split in nested function type forces outer split
+Function(int, String, Function(parameter1, parameter2, parameter3)) f;
+<<<
+Function(
+    int,
```
natebosch: [nit] this is misleading... IIUC this is a parameter named 'int' of type 'dynamic'. Or am I wrong?

test/splitting/function_types.unit
munificent: Yeah, not all of the tests have descriptions. I try to add them when it's worthwhile, but it's often obvious from the test.

In this case, it was an accidental copy/paste error and is redundant with the above one. Removed it.

test/splitting/function_types.unit
munificent: It is actually a type (though it doesn't matter either way for this test). Unlike the old function type syntax, the new syntax wisely assumes if you omit one identifier, you're omitting the name, not the type.

test/splitting/function_types.unit
munificent: See above.

PR #632, opened 2017-05-30
replace .toString() with .lexeme

danrubel:
Minor fix. Something I bumped into while integrating the fasta scanner into analyzer.


danrubel:
Good point. Is this what you were thinking should be done?

PR #634, opened 2017-06-15
Format multiline strings as block arguments.

### LINKED ISSUE #570: scheduled_test descriptor pattern is indented very strangely
The formatter doesn't play nicely at all with the following pattern from `scheduled_test`'s `descriptor` library:

```dart
void main() {
  d.file("test.dart", """
    import 'package:test/test.dart';
    
    void main() {
      test("failure", () => throw new TestFailure("oh no"));
    }
  """).create();
}
```

is formatted as:

```dart
void main() {
  d
      .file(
          "test.dart",
          """
    import 'package:test/test.dart';
    
    void main() {
      test("failure", () => throw new TestFailure("oh no"));
    }
  """)
      .create();
}
```

This is definitely a lot harder to read.

nex3:
Closes #570

Note that I'm not wedded to the nomenclature here at all.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -78,8 +78,8 @@ class SourceVisitor extends ThrowingAstVisitor {
   /// the block body is indented like an expression or a statement.
   ///
   /// Before a block argument is visited, [ArgumentSublist] binds itself to the
-  /// left bracket token of each collection literal it controls. When we later
-  /// visit that literal, we use the token to find that association.
+  /// beginning token of each block it controls. When we later visit that
+  /// literal, we use the token to find that association.
   final Map<Token, ArgumentSublist> _blockArgumentLists = {};
 
   /// Initialize a newly created visitor to write source code representing
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -68,20 +68,19 @@ class SourceVisitor extends ThrowingAstVisitor {
   /// in unison.
   final List<MetadataRule> _metadataRules = [];
 
-  /// The mapping for collection literals that are managed by the argument
-  /// list that contains them.
+  /// The mapping for blocks that are managed by the argument list that contains
+  /// them.
   ///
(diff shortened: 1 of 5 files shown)


lib/src/source_visitor.dart
```diff
@@ -68,20 +68,19 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+  /// also tracks the chunk before the argument that determines whether or not
+  /// the block body is indented like an expression or a statement.
   ///
-  /// Before a collection literal argument is visited, [ArgumentSublist] binds
-  /// itself to the left bracket token of each collection literal it controls.
-  /// When we later visit that literal, we use the token to find that
-  /// association.
-  final Map<Token, ArgumentSublist> _collectionArgumentLists = {};
+  /// Before a block argument is visited, [ArgumentSublist] binds itself to the
+  /// left bracket token of each collection literal it controls. When we later
```
munificent: "left bracket token of each collection literal it controls" → "beginning token of each block it controls".

test/splitting/strings.stmt
```diff
@@ -53,4 +53,24 @@ someMethod("""first line does not fits here
...
+>>> format multiline a block with a trailing argument
+someMethod("""
+  some
+  text
+""", "foo");
+<<<
+someMethod("""
+  some
+  text
+""", "foo");
```
munificent: Can you test:

* Some negative cases, like a multiline string with non-block arguments before and after.
* A sequence of multiline strings in an argument list.
* A mixture of multiline strings and other block arguments like collection literals.

lib/src/source_visitor.dart
nex3: Done.

test/splitting/strings.stmt
nex3: > Some negative cases, like a multiline string with non-block arguments before and after.

Did one. I'm not sure what other negative cases are meaningfully different, though.

> A sequence of multiline strings in an argument list.

Done.

> A mixture of multiline strings and other block arguments like collection literals.

Done.

PR #635, opened 2017-06-15
Don't add extra indentation when using a decorator pattern.

### LINKED ISSUE #382: Special-case decorator-style methods
I use the following pattern occasionally to simulate decorators; that is, to wrap the body of a method in some block or another to modify its behavior. It avoids extra indentation and looks nice, without being too complex:

``` dart
Future doThing() => pool.withResource(() {
  // ...
});
```

Currently, the formatter formats this like so:

``` dart
Future doThing() => pool.withResource(() {
      // ...
    });
```

It would be nice if it recognized the specific pattern of a `=>` method with a block-format body and indented it like a normal method body.

nex3:
Closes #382

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -886,17 +886,35 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     // If the body is a binary operator expression, then we want to force the
     // split at `=>` if the operators split. See visitBinaryExpression().
-    if (node.expression is! BinaryExpression) builder.endRule();
+    var expression = node.expression;
+    if (expression is! BinaryExpression) builder.endRule();
 
     if (_isInLambda(node)) builder.endSpan();
 
-    builder.startBlockArgumentNesting();
-    builder.startSpan();
-    visit(node.expression);
-    builder.endSpan();
-    builder.endBlockArgumentNesting();
+    // If the body is a block-formatted function invocation, don't add extra
+    // nesting to preserve normal indentation for decorators that modify the
+    // functionality of a method body:
+    //
+    //    void foo() => decorate(() {
+    //      "the body";
+    //    });
+    var isDecorator = expression is InvocationExpression &&
+        expression.argumentList.arguments.any((argument) {
+          if (argument is FunctionExpression) {
+            var body = argument.body;
+            return body is BlockFunctionBody && body.block.statements.isNotEmpty;
+          } else {
+            return false;
+          }
+        });
 
-    if (node.expression is BinaryExpression) builder.endRule();
+    if (!isDecorator) builder.startBlockArgumentNesting();
+    builder.startSpan();
+    visit(expression);
+    builder.endSpan();
+    if (!isDecorator) builder.endBlockArgumentNesting();
+
+    if (expression is BinaryExpression) builder.endRule();
 
     token(node.semicolon);
   }
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -899,14 +899,8 @@ class SourceVisitor extends ThrowingAstVisitor {
     //      "the body";
     //    });
     var isDecorator = expression is InvocationExpression &&
-        expression.argumentList.arguments.any((argument) {
-          if (argument is FunctionExpression) {
-            var body = argument.body;
-            return body is BlockFunctionBody && body.block.statements.isNotEmpty;
-          } else {
-            return false;
-          }
-        });
+        new ArgumentListVisitor(this, expression.argumentList)
+            .hasBlockArguments;
 
     if (!isDecorator) builder.startBlockArgumentNesting();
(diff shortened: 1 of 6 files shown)


nex3:
I personally wouldn't mind broadening the logic for this, but I kept it as narrow as possible to avoid changing much existing formatting.

lib/src/source_visitor.dart
```diff
@@ -886,17 +886,35 @@ class SourceVisitor extends ThrowingAstVisitor {
...
-    builder.startBlockArgumentNesting();
+    // If the body is a block-formatted function invocation, don't add extra
+    // nesting to preserve normal indentation for decorators that modify the
+    // functionality of a method body:
+    //
+    //    void foo() => decorate(() {
+    //      "the body";
+    //    });
+    var isDecorator = expression is InvocationExpression &&
+        expression.argumentList.arguments.any((argument) {
```
munificent: This logic is similar to, but not the same as the logic for handling indentation of function literals in other argument lists. In particular, I don't think you'd want to apply the special decorator logic if there are non-function arguments before and after the function in the function list, or between functions.

The logic for determining when functions can be un-indented in an argument list is pretty sensitive, so it would be good to reuse that code if at all possible.

test/whitespace/functions.unit
```diff
@@ -165,4 +165,12 @@ main() {
 >>> generic function typed parameter
```
munificent: Can you also add a regression test for the corresponding bug?

test/whitespace/functions.unit
```diff
@@ -165,4 +165,12 @@ main() {
...
\ No newline at end of file
+function(int foo<T, S>(T t, S s)) {}
+>>> decorator-style function
+function() => decorator(() {
+  body;
+});
+<<<
+function() => decorator(() {
+  body;
+});
```
munificent: What does the output look like if:

* The parameter list for the function has to wrap.
* The argument list for the decorator call before the function has to wrap.
* The argument list for the decorator call after the function has to wrap.

lib/src/source_visitor.dart
nex3: I'm not sure where that logic exists or how I should re-use it.

test/whitespace/functions.unit
nex3: Done.

test/whitespace/functions.unit
nex3: Done, but it turns out this doesn't work—the logic above doesn't see the comment as a statement. How should I check for something like that?

lib/src/source_visitor.dart
munificent: Oh, sorry. It's in ArgumentListVisitor. Look around for mentions of "function" and "block".

lib/src/source_visitor.dart
```diff
@@ -886,17 +886,35 @@ class SourceVisitor extends ThrowingAstVisitor {
...
 
-    builder.startBlockArgumentNesting();
+    // If the body is a block-formatted function invocation, don't add extra
+    // nesting to preserve normal indentation for decorators that modify the
+    // functionality of a method body:
+    //
+    //    void foo() => decorate(() {
+    //      "the body";
+    //    });
+    var isDecorator = expression is InvocationExpression &&
```
munificent: Should we also check that the ExpressionFunctionBody is for a named declaration too? I'm guessing we don't want this logic to kick in for a `=>` lambda, right?

test/whitespace/functions.unit
munificent: Did you get this figured out?

test/whitespace/functions.unit
munificent: Don't forget the first bullet point: the function declaration itself wraps, not the decorated one:

```dart
function(long, stuff, here) => ...
```

lib/src/source_visitor.dart
munificent: I did some testing locally, and I think using it for lambdas generates too many false positives, so I think we should only support this for named declarations.

munificent:
Following up on what we talked about:

I did some local testing of this on a big corpus. It definitely really helps in the decorator cases. But there are a lot of false positives where you have a method body that's something like a call to `map()` or `then()` where the user doesn't think of it as a decorator. In those cases, I think getting rid of the indentation makes the code a little harder to read and more confusing.

One thing that reduces the number of false positives is *not* using ArgumentListVisitor. That has some special logic to allow "wrapped" function literals for things like:

```dart
method(wrap(() {
  ...
});
```

That, combined with this PR means you can sort of get two "levels" of decoration and those cases are almost always false positives.

Even without that, though, there are still more false positives than I feel comfortable with. For now, I'm just going to sit on this until I get some more time to poke at it and maybe ask users what they think about various before/after examples.

munificent:

Issue #21, opened 2014-11-10
Force newline between adjacent strings


alan-knight:
from intl/test/message_extraction/verify_messages.dart

before:
```
  verify(
      "Caractères qui doivent être échapper, par exemple barres \\ "
      "dollars \${ (les accolades sont ok), et xml/html réservés <& et "
      "des citations \" "
      "avec quelques paramètres ainsi 1, 2, et 3");
```

after:
```
  verify("Caractères qui doivent être échapper, par exemple barres \\ "
      "dollars \${ (les accolades sont ok), et xml/html réservés <& et "
      "des citations \" " "avec quelques paramètres ainsi 1, 2, et 3");
```
It's possible that the escaped quote in there might also be involved.

munificent:
It doesn't *force* one now, but it will *preserve* one if you have it.

PR #636, opened 2017-06-15
Improve the handling of adjacent strings.

nex3:
This essentially makes the formatter match the style we used on pub
before it got formatted. Also, by ignoring the user's splits, it helps
call out cases where adjacent strings are unnecessary and can be
merged into single literals.

### DIFF
--- a/lib/src/nesting_builder.dart
+++ b/lib/src/nesting_builder.dart
@@ -80,11 +80,20 @@ class NestingBuilder {
   /// If omitted, [spaces] defaults to [Indent.block].
   void indent([int spaces]) {
     if (spaces == null) spaces = Indent.block;
+
+    // Indentation should only change outside of nesting.
+    assert(_pendingNesting == null);
+    assert(_nesting.indent == 0);
+
     _stack.add(_stack.last + spaces);
   }
 
   /// Discards the most recent indentation level.
   void unindent() {
+    // Indentation should only change outside of nesting.
+    assert(_pendingNesting == null);
+    assert(_nesting.indent == 0);
+
     // If this fails, an unindent() call did not have a preceding indent() call.
     assert(_stack.isNotEmpty);
 
--- a/lib/src/nesting_builder.dart
+++ b/lib/src/nesting_builder.dart
@@ -80,20 +80,11 @@ class NestingBuilder {
   /// If omitted, [spaces] defaults to [Indent.block].
   void indent([int spaces]) {
     if (spaces == null) spaces = Indent.block;
-
-    // Indentation should only change outside of nesting.
-    assert(_pendingNesting == null);
-    assert(_nesting.indent == 0);
-
     _stack.add(_stack.last + spaces);
   }
 
   /// Discards the most recent indentation level.
   void unindent() {
-    // Indentation should only change outside of nesting.
-    assert(_pendingNesting == null);
-    assert(_nesting.indent == 0);
-
     // If this fails, an unindent() call did not have a preceding indent() call.
     assert(_stack.isNotEmpty);
 
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -113,10 +113,38 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     var chunk = zeroSplit();
     rule.imply(chunk.rule);
-    visitNodes(node.strings, between: splitOrNewline);
+
+    visit(node.strings.first);
+    var previous = node.strings.first;
+    for (var string in node.strings.skip(1)) {
+      // Always split after a string that ends in a newline. Otherwise, indent a
+      // little bit so it's clear that the next string is a continuation rather
+      // than a new line.
+      if (_endsWithNewline(previous)) {
+        builder.writeWhitespace(Whitespace.nestedNewline);
+      } else {
+        builder.indent();
+        split();
+        builder.unindent();
+      }
+      visit(string);
+      previous = string;
+    }
 
     builder.endRule();
     builder.endSpan();
+  }
+
+  /// Returns whether [string] ends with a newline character.
+  bool _endsWithNewline(StringLiteral string) {
+    if (string is AdjacentStrings) {
+      return _endsWithNewline(string.strings.last);
+    } else if (string is StringInterpolation) {
+      var lastElement = string.elements.last;
+      return lastElement is InterpolationString && lastElement.value.endsWith("\n");
+    } else {
+      return (string as SimpleStringLiteral).value.endsWith("\n");
+    }
   }
 
   visitAnnotation(Annotation node) {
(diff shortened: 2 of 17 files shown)


lib/src/source_visitor.dart
```diff
@@ -108,12 +108,45 @@ class SourceVisitor extends ThrowingAstVisitor {
 
   visitAdjacentStrings(AdjacentStrings node) {
     builder.startSpan();
-    builder.startRule();
-    visitNodes(node.strings, between: splitOrNewline);
+    var rule = new Rule();
+    builder.startRule(rule);
+
+    var chunk = zeroSplit();
```
munificent: You don't want a split before the first string. It's the responsibility of the expression containing the string to put a split here if it needs it.

lib/src/nesting_builder.dart
```diff
@@ -80,20 +80,11 @@ class NestingBuilder {
...
-    assert(_nesting.indent == 0);
-
     _stack.add(_stack.last + spaces);
   }
 
   /// Discards the most recent indentation level.
   void unindent() {
-    // Indentation should only change outside of nesting.
-    assert(_pendingNesting == null);
-    assert(_nesting.indent == 0);
```
munificent: These sanity checks are useful. If they are tripping, then it's likely you're getting the formatter into a state where it will produce weird output in some cases. You may not have seen those cases in the examples you tested, but they're probably lurking out there.

If at all possible, you'll want to change the visitor code so that these don't fire.

lib/src/source_visitor.dart
```diff
@@ -108,12 +108,45 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+      // Always split after a string that ends in a newline. Otherwise, indent a
+      // little bit so it's clear that the next string is a continuation rather
+      // than a new line.
+      if (_endsWithNewline(previous)) {
+        builder.writeWhitespace(Whitespace.nestedNewline);
+      } else {
+        builder.indent();
+        split();
+        builder.unindent();
+      }
```
munificent: I think this policy is a little too specific for dartfmt. I agree it looks nice in the pub examples, but it's meaningless in lots of other examples where the strings in question are not lines of human readable text. If you look at the changed regression tests, there are strings of JSON and other stuff like that where the extra indentation is just confusing.

Also, this is what's causing the above asserts to fire. Indentation is for statement-level stuff and precedes expression nesting. Once you're inside an expression, messing with the indentation is likely to not do what you want.

lib/src/source_visitor.dart
nex3: If I don't add a split here, how do I ensure that there's always a split before the first string if the rest of the strings split?

lib/src/source_visitor.dart
nex3: > I think this policy is a little too specific for dartfmt. I agree it looks nice in the pub examples, but it's meaningless in lots of other examples where the strings in question are not lines of human readable text. If you look at the changed regression tests, there are strings of JSON and other stuff like that where the extra indentation is just confusing.

I strongly disagree. Based on the SDK, the vast majority of adjacent string literals are text where the distinction between a newline and a continued line is relevant to the reader. Even in cases where it's not relevant to the format itself, having this as a universal convention helps communicate to a reader that the string being constructed will exist all on one line.

Also, the current lack of a policy isn't any more meaningful. The regression test for #21 [wipes out the prior formatting entirely](https://github.com/dart-lang/dart_style/blob/master/test/regression/0000/0021.stmt#L13-L49). Having some policy that works well in the substantial majority of cases is better than having no policy at all—isn't that the entire point of the formatter?

> Also, this is what's causing the above asserts to fire. Indentation is for statement-level stuff and precedes expression nesting. Once you're inside an expression, messing with the indentation is likely to not do what you want.

I'm not sure I follow. What I want here is a little extra indentation for this particular split, which happens to be inside an expression. What else should I use to get that effect?

munificent: "Nesting" also causes indentation. The formatter internally uses "indent" to mean the sort of "block-level" indentation you get from statements and "nesting" to refer to the indentation you get when line wraps appear inside expressions.

A lot of the code assumes there will never be nesting in play at the point when indentation changes, because that implies you're still in the middle of an expression when crossing a statement boundary, which shouldn't be the case.

So instead of using `indent()` and `unindent()` here, you'll use `nestExpression()` and `unnest()`.

nex3: > OK. Can you indent +4 instead of +2? +2 is very rarely used at the expression level, which is where a series of adjacent strings are. Cascades are basically the only place where +2 happens there, because they are special magical snowflakes. Collection literals are +2 as well, but when they split, they split more like blocks than like expressions.

This looks a little odd to me, but maybe that's just because it's not what I'm used to. Done.

> So instead of using `indent()` and `unindent()` here, you'll use `nestExpression()` and `unnest()`.

Done.

lib/src/source_visitor.dart
munificent: Oof, that's hard. You don't actually always want to ensure there's a split before the first string. Consider:

```dart
{
  "long adjacent" "long adjacent";
}
```

That should not get formatted to:

```dart
{
  // <-- split added here.
  "long adjacent"
  "long adjacent";
}
```

Likewise:

```dart
var x = foo ? "short" : "long adjacent" "long adjacent";
```

Should maybe not become:

```dart
var x = foo
    ? "short"
    :
        "long adjacent"
        "long adjacent";
```

Or maybe it should? It may be that the best way to handle this is to have ArgumentListVisitor specifically handle arguments that are adjacent strings.

nex3: > ```dart
> {
>   "long adjacent" "long adjacent";
> }
> ```
> 
> That should not get formatted to:
> 
> ```dart
> {
>   // <-- split added here.
>   "long adjacent"
>   "long adjacent";
> }
> ```

That's probably true, but this seems like an extremely narrow edge-case considering that such a statement can't possibly do anything.

> Likewise:
> 
> ```dart
> var x = foo ? "short" : "long adjacent" "long adjacent";
> ```
> 
> Should maybe not become:
> 
> ```dart
> var x = foo
>     ? "short"
>     :
>         "long adjacent"
>         "long adjacent";
> ```
> 
> Or maybe it should?

I'd hand-format this as

```dart
var x = foo
    ? "short"
    : "long adjacent"
      "long adjacent";
```

I could try to add a special case to do that, but I don't think the formatting you listed is too bad. 

> It may be that the best way to handle this is to have ArgumentListVisitor specifically handle arguments that are adjacent strings.

This would make a fair number of `return` and `throw` cases look worse :confused:.

munificent: How do you intend to handle adjacent strings in `return` and `throw`? I would not expect it to push the first string to the next line there (and we definitely *don't* want it to if we're going to do optional semicolons ever).

nex3: > How do you intend to handle adjacent strings in `return` and `throw`? I would not expect it to push the first string to the next line there (and we definitely *don't* want it to if we're going to do optional semicolons ever).

That was my intention. It's the only way to keep the strings aligned, unless we try to match human formatting by adding `"return ".length`/`"throw ".length` spaces of indentation.

munificent: Aligning past the keyword would be nice, but I don't want to open that can of worms. If we do it there, users will (not wrongly) want it everywhere, and that gets very hard very quickly.

We also don't want to put a newline after the keyword. That will paint us into a very nasty corner when we later try to make semicolons optional. I think the output should be:

```dart
return "some string ..."
    "next line ...";
```

munificent:

PR #639, opened 2017-06-20
Support assert in initializer list experiment.

### LINKED ISSUE #522: Formatter should handle asserts in initializer list.
See dart-lang/sdk#27141 and dart-lang/sdk#27142 for context.

When the analyzer supports the syntax for assert initializers, the formatter needs to handle it gracefully.

munificent:
Fix #522.

### DIFF
--- a/lib/src/dart_formatter.dart
+++ b/lib/src/dart_formatter.dart
@@ -103,7 +103,7 @@ class DartFormatter {
 
     // Parse it.
     var parser = new Parser(stringSource, errorListener);
-    parser.parseGenericMethods = true;
+    parser.enableAssertInitializer = true;
 
     AstNode node;
     if (source.isCompilationUnit) {
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -171,22 +171,17 @@ class SourceVisitor extends ThrowingAstVisitor {
     builder.endSpan();
   }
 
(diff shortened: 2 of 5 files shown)


lib/src/dart_formatter.dart
```diff
@@ -103,7 +103,7 @@ class DartFormatter {
 
     // Parse it.
     var parser = new Parser(stringSource, errorListener);
-    parser.parseGenericMethods = true;
+    parser.enableAssertInitializer = true;
```
kevmoo: sure the min constraint on pkg/analyzer matches this?

munificent: I *believe* so, yes.

lib/src/source_visitor.dart
```diff
@@ -171,22 +171,17 @@ class SourceVisitor extends ThrowingAstVisitor {
     builder.endSpan();
   }
 
-  // TODO(rnystrom): Type annotate once analyzer publishes a version with the
-  // new AST type.
-  // TODO(rnystrom): Test.
-  visitAssertInitializer(node) {
-    _simpleStatement(node, () {
```
nex3: What was `_simpleStatement()` doing here? Why is it safe to remove?

munificent: Adding a trailing semicolon, mainly. It was a bug, uncaught because there were no tests for it since at the time the code was written, there wasn't a published version of analyzer that supported parsing the syntax.

PR #641, opened 2017-07-01
Always vertically align adjacent strings.

nex3:
This is an alternative to #636 which avoids adding a split before
adjacent stings, while also ensuring that they're always vertically
aligned. It currently doesn't support indenting continued strings, but
if you like this direction I can work on that.


munificent:
This is really nicely done, but I don't want to do alignment just for adjacent strings. I think it will annoy users when we don't align other kinds of expressions, and going *generally* in the direction of alignment could end up being a sweeping change to lots of code. I don't want to consider that right now (though it may be a good idea to do so later).

nex3:
I looked through the various feature requests for the formatter, and it looked like the only other requests for alignment are fundamentally different from this. They want alignment of different components within lines, like this:

```dart
var foo    = new      Qux(argument);
var barbaz = new Fizzbang(argument);
```

I didn't see any other requests for the type of alignment this CL adds, where the *indentation* matches between lines. Although they do both involve some sort of vertical alignment, they look very different. I think to the extent that users draw a parallel between them, it'll be easy to understand that aligning indentation is easier than aligning within lines.

munificent:
> I looked through the various feature requests for the formatter, and it looked like the only other requests for alignment are fundamentally different from this. 

The issue tracker may not reflect this, but some users have definitely asked for arbitrary alignment of parallel subexpressions like:

```dart
method(argument,
  second,
  function(third,
           fourth));
```

nex3:
That's fair, but there are compelling reasons to align adjacent strings that don't generally exist for arbitrary parameters. Adjacent strings represent a single block of text, and it's often important for inner portions of that text to be aligned—it helps ensure that text is readable both in the source code and wherever the string ends up.

munificent:

Issue #638, opened 2017-06-16
Drop use of "grinder" and bump args to <2.0.0


matanlurey:
Otherwise it's impossible to use `args 1.0.0` in some packages.

matanlurey:
I tried to fix this myself, but it's not possible because of `grinder`, which in turn depends on an abandoned package `unscripted`. I'd like to suggest we drop use of `grinder` in this package since `dart_style` is so important and adding this set of transitive dependencies is expensive to maintain.

munificent:
The grinder step is handy because it ensures that the version of dart_style printed by `--version` is the same as the package's actual version. We could get rid of grinder, but I would like some automated script — even a little hand-rolled one — that does what bump and validate do.

matanlurey:
That seems fair, maybe it makes sense to have some sort of shared tool to do this via pub global activate @munificent? /cc @kevmoo and @natebosch

natebosch:
I think we can hand roll something pretty easily. Would be nice to have in pub but probably high effort to payoff compared to other work we could do.

munificent:
> maybe it makes sense to have some sort of shared tool to do this via pub global activate

Uh, isn't that what grinder *is*?

natebosch:
Do we still use grind.dart to publish to npm?
Looks like it's been a while since it was published...

munificent:
@jacob314 is the only one who's published dart_style to npm. I don't consider it part of the regular publish step. He'd know more about that.

jacob314:
It isn't critical to publish. As we aren't actively working on auto-converting TS to Dart files at the moment. Until we can make all the runtimes support converting futures to promises automatically or we support extension methods there isn't much of a point as too many new APIs are using promises.

munificent:
Closing this because I don't think it's blocking any more.

PR #644, opened 2017-07-09
Bump use of pkg/args to <2.0.0.

matanlurey:
Partial towards https://github.com/dart-lang/dart_style/issues/638.

This is important because `dart_style` is used as a `dependency` in a few packages (think code-generation), and limiting `args` makes developing them (and using newer features) much harder.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
   sdk: ">=1.8.0 <2.0.0"
 dependencies:
   analyzer: '^0.30.0'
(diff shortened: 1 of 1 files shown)


matanlurey:
@munificent Thx! I'll make this change before merging.

kevmoo:
@matanlurey – FYI, already bumped to `1.0.8-dev` here https://github.com/dart-lang/dart_style/commit/3a04d154f8259563175d891cca30b1104a9ef103

PR #646, opened 2017-07-27
WIP: Analyzer 0 31 sanity

kevmoo:


pubspec.yaml
```diff
@@ -29,3 +29,6 @@ executables:
 transformers:
 - $dart2js:
     $exclude: 'test/**'
+
+dependency_overrides:
+  analyzer: 0.31.0-alpha.0
```
munificent: We don't want to commit this with dependency overrides.

lib/src/exceptions.dart
```diff
@@ -35,7 +35,7 @@ class FormatterException implements Exception {
         source += " " * (error.offset + error.length - source.length);
       }
 
-      var file = new SourceFile(source, url: error.source.fullName);
+      var file = new SourceFile.fromString(source, url: error.source.fullName);
```
munificent: Is this because of a change in the analyzer API? If so, you'll need to increase the minimum version constraint on analyzer.

kevmoo:
Closing this. Just an experiment

PR #651, opened 2017-09-21
Stop using pkg/scheduled_test

kevmoo:


kevmoo:
Ping, @munificent ?

.gitignore
```diff
@@ -1,5 +1,3 @@
-.idea
 .pub
-packages
 .packages
-build
\ No newline at end of file
+pubspec.lock
```
munificent: I do check in the lockfile. It doesn't affect consumers of dart_style, but it helps ensure everyone hacking on it gets the same experience.

pubspec.lock
```diff
@@ -1,335 +0,0 @@
-# Generated by pub
```
munificent: Let's not ditch this file.

test/io_test.dart
```diff
@@ -9,146 +9,126 @@ import 'dart:io';
...
-    var badBefore;
-    var goodBefore;
+    DateTime badBefore;
+    DateTime goodBefore;
 
-    schedule(() {
-      badBefore = modTime('bad.dart');
-      goodBefore = modTime('good.dart');
+    badBefore = modTime('bad.dart');
+    goodBefore = modTime('good.dart');
```
munificent: May as well unify these with the above two lines:

```dart
var badBefore = modTile('bad.dart');
var goodBefore = modTime('good.dart');
```

pubspec.lock
kevmoo: ack

test/io_test.dart
kevmoo: yup

PR #663, opened 2018-01-05
Add pubspec.lock to gitignore

grouma:


nex3:
I assume there's a reason `pubspec.lock` is checked in here, but I'm not sure what it is. Reassigning to @munificent.

munificent:
I actually do like having the lockfile checked in.

PR #671, opened 2018-02-21
Don't split before "." when the target has a trailing comma.

### LINKED ISSUE #548: Strange indentation after using DSL-like method call chain
Using trailing commas, we can build a nice DSL-like call:

``` dart
StatementBuilder assignConst(String variable, [TypeBuilder type]) {
    return new _AsAssign(
      this,
      variable,
      keyword: _AsAssignType.asConst,
      type: type,
    ).asStatement();
  }
```

However, this is formatted strangely, to:

``` dart
return new _AsAssign(
      this,
      variable,
      keyword: _AsAssignType.asConst,
      type: type,
    )
        .asStatement();
    // ^^^^^^^^^^^^^^^
    // Why a new line and then an indent?
```

Removing the trailing comma also results in a really weird output:

``` dart
return new _AsAssign(this, variable,
            keyword: _AsAssignType.asConst, type: type)
        .asStatement();
```

Or returning using '=>`:

``` dart
StatementBuilder assignConst(String variable, [TypeBuilder type]) =>
      new _AsAssign(
        this,
        variable,
        keyword: _AsAssignType.asConst,
        type: type,
      )
          .asStatement();
      // ^^^^^^^^^^^^^^^
      // Why a new line and then an indent?
```

### LINKED ISSUE #665: Function calls get formatted with an extra newline before a function call on the result.
This is from a test, and the unfortunate thing here is that the .thenReturn is on a new line and indented:

```dart
when(
 runner.call(
   typed(captureAny),
   typed(captureAny),
   workingDirectory: typed(captureAny, named: "workingDirectory"),
 ),
)
   .thenReturn(new Future<MockProcessResult>.value(new MockProcessResult()));
```

I would format it by hand like this:

```dart
when(
 runner.call(
   typed(captureAny),
   typed(captureAny),
   workingDirectory: typed(captureAny, named: "workingDirectory"),
 ),
).thenReturn(new Future<MockProcessResult>.value(new MockProcessResult()));
```

munificent:
Fix #548. Fix #665.

### DIFF
--- a/test/regression/0500/0548.stmt
+++ b/test/regression/0500/0548.stmt
@@ -1,0 +1,44 @@
+>>> (indent 2)
+  StatementBuilder assignConst(String variable, [TypeBuilder type]) {
+    return new _AsAssign(
+      this,
+      variable,
+      keyword: _AsAssignType.asConst,
+      type: type,
+    )
+        .asStatement();
+  }
+<<<
+  StatementBuilder assignConst(String variable, [TypeBuilder type]) {
+    return new _AsAssign(
+      this,
+      variable,
+      keyword: _AsAssignType.asConst,
+      type: type,
+    ).asStatement();
+  }
+>>> (indent 4)
+return new _AsAssign(this, variable,
+            keyword: _AsAssignType.asConst, type: type)
+        .asStatement();
+<<<
+    return new _AsAssign(this, variable,
+            keyword: _AsAssignType.asConst, type: type)
+        .asStatement();
+>>> (indent 2)
+StatementBuilder assignConst(String variable, [TypeBuilder type]) =>
+      new _AsAssign(
+        this,
+        variable,
+        keyword: _AsAssignType.asConst,
+        type: type,
+      )
+          .asStatement();
+<<<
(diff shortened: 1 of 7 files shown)


test/regression/0500/0548.stmt
```diff
@@ -0,0 +1,44 @@
...
+  }
+<<<
+  StatementBuilder assignConst(String variable, [TypeBuilder type]) {
+    return new _AsAssign(
+      this,
+      variable,
+      keyword: _AsAssignType.asConst,
+      type: type,
+    ).asStatement();
+  }
```
munificent: Each pair of snippets in these test cases is before followed by after.

a14n: What's the behaviour if the method chain need to wrap ?

munificent: Then it does split before the `.`:

```dart
// If it fits, don't split:
function(
  argument,
).method().method();

// If it splits, split before all of them, even the first:
function(
  argument,
)
    .method()
    .method()
    .method()
    .method()
    .method();
```

Thanks for asking this, it reminded me to add some more test coverage.

Issue #508, opened 2016-05-30
Vertical alignment


bergwerf:
So I worked a bit with Go again last week and one thing I really like is the automatic vertical alignment from the Go formatter like in this example snippet:

``` go
categoryHeader := WidgetCategoryHeader{
    Type:           "categoryheader",
    Title:          GetTranslation(categoryData, person.Language, "Name").(string),
    HeaderBgColor:  categoryTheme.HeaderBgColor,
    HeaderColor:    categoryTheme.HeaderColor,
    HeaderImageUrl: categoryData.ImgUrl,
}
```

Below is a snippet from one of my Dart projects:

``` dart
return new SnippetSelector(data['class'] as String,
  database: data['database'] as String,
  table: data['table'] as String,
  all: data['all'] as bool,
  attributes: data['attributes'] as Map<String, String>,
  parameters: data['parameters'] as Map<String, String>);
```

Now with some vertical alignment it could look like this:

``` dart
return new SnippetSelector(data['class'] as String,
  database:   data['database']   as String,
  table:      data['table']      as String,
  all:        data['all']        as bool,
  attributes: data['attributes'] as Map<String, String>,
  parameters: data['parameters'] as Map<String, String>);
```

I certainly like this much better than the previous snippet. It makes certain parts of my code look much cleaner. A Map could be aligned the same way.

``` dart
Map<String, dynamic> map = {
  'one':      1,
  'ten':      10,
  'hundred':  100,
  'thousand': 1000
}
```

`dartfmt` is part of the pre-commit hook in all my Dart projects (and I've never been so happy!), but hand formatting for this kind of cases is impossible.

I realize that this feature might be very difficult to implement. And probably not everyone agrees vertical alignment is a good thing. Even more so because `dartfmt` is not configurable, and changing this would mean everyone's code gets vertical alignment. The vertical alignment step could also be a 'post-processing' tool that you can run after `dartfmt`.

Anyway, I could not find an issue mentioning this subject.

munificent:
Yeah, it's good to have a tracking issue for this. Like you note, yes, it's very difficult to implement. gofmt doesn't do line breaking, so vertical alignment is a much simpler problem for it. :)

It's a deliberate choice in dartfmt to *not* do vertical alignment because:

* It's very complex to implement.
* Assuming you can implement it, it may be very slow.
* It's not obvious what the rules for which things to align should be.
[...]

PR #672, opened 2018-02-21
No new line before initializer list on constructor with trailling comma

### LINKED ISSUE #658: Constructor with parameters with trailing comma and initializer list inserts extra newline
What I get:

```
class A {
  A(
    int a,
  )
      : super();
}
```

What I expected :

```
class A {
  A(
    int a,
  ) : super();
}
```

a14n:
Closes #658

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -705,15 +705,28 @@ class SourceVisitor extends ThrowingAstVisitor {
   }
 
   void _visitConstructorInitializers(ConstructorDeclaration node) {
-    // Shift the ":" forward.
-    builder.indent(Indent.constructorInitializer);
+    final parameters = node.parameters?.parameters;
+    final hasTraillingComma = parameters?.isNotEmpty == true &&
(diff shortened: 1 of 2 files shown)


munificent:
Thank you! I made a few changes, mostly just cleaning up the code and reorganizing the tests. The only substantive change is that it does not do +5 indentation when the trailing comma is an optional parameter.

That does keep all of the initializers lined up, but it goes against all of the other formatting rules where we do not do odd indentation in order to line up elements. See also: #508.

a14n:
The +4/+5 indentation was done to be inline with what Flutter codebase does (named optional parameters are eavily used in Flutter).

> That does keep all of the initializers lined up, but it goes against all of the other formatting rules where we do not do odd indentation in order to line up elements.

One could argument that lining up elements in initializer list is like lining up members of a class. IMHO it's more readable (and less strange) to have +5 with named parameters. Given that initializer list can already have 2 indent length (+4 or +6) it doesn't seem a big issue to have +5 to *always* have elements line up in initializer.

PR #673, opened 2018-02-22
Tweak how initialization lists are handled after trailing commas.

munificent:
This gets avoids the pointless newlines of the old formatting rules,
but keeps it otherwise closer to the original behavior than
43832d24e7b86a165ec6a1765cbbb3644cae5f38 does.

In particular, initialization lists are always indented +6. This ensures
that initialization lists line up for constructors with trailing commas,
without, with optional parameters, mandatory, or none at all.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -716,37 +716,50 @@ class SourceVisitor extends ThrowingAstVisitor {
       //       parameter,
       //     ) : super();
       space();
+
+      var isOptional = node.parameters.parameters.last.kind != ParameterKind.REQUIRED;
+      if (node.initializers.length > 1) {
+        _writeText(isOptional ? " " : "  ", node.separator.offset);
+      }
+
+      // ":".
+      token(node.separator);
+
+      space();
+
+      builder.indent(6);
     } else {
       // Shift the itself ":" forward.
       builder.indent(Indent.constructorInitializer);
 
       // If the parameters or initializers split, put the ":" on its own line.
       split();
+
+      // ":".
+      token(node.separator);
+      space();
+
+
+      // Try to line up the initializers with the first one that follows the ":":
+      //
+      //     Foo(notTrailing)
+      //         : initializer = value,
+      //           super(); // +2 from previous line.
+      //
+      //     Foo(
+      //       trailing,
+      //     ) : initializer = value,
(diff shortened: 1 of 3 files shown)


lib/src/source_visitor.dart
```diff
@@ -716,37 +716,50 @@ class SourceVisitor extends ThrowingAstVisitor {
       //       parameter,
       //     ) : super();
       space();
+
+      var isOptional =
+          node.parameters.parameters.last.kind != ParameterKind.REQUIRED;
+      if (node.initializers.length > 1) {
+        _writeText(isOptional ? " " : "  ", node.separator.offset);
```
natebosch: [optional] It's hard for me to tell the difference between `" "` and `"  "` as I'm scanning the code.

Maybe make this `" "` and `" " * 2`?

lib/src/source_visitor.dart
```diff
@@ -716,37 +716,50 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+      token(node.separator);
+
+      space();
+
+      builder.indent(6);
     } else {
       // Shift the itself ":" forward.
       builder.indent(Indent.constructorInitializer);
 
       // If the parameters or initializers split, put the ":" on its own line.
```
natebosch: this comment feels strange to me.

Every `hasTrailingComma` would have split parameters, but not all split parameters or initializers are caused by `hasTrailingComma`.

Maybe "// Every parameter list with a trailing comma will have split initializers which puts the ":" on it's own line"

munificent: This section of code is for when the parameter list does *not* have a trailing comma, so I'm not sure we need to mention that here.

lib/src/source_visitor.dart
munificent: Hmm, that feels a little gratuitous.

lib/src/source_visitor.dart
natebosch: ugh yeah, I'm even more confused....

How do we know the parameters or intializers are split when we get here?

munificent: Sorry, this makes more sense if you're more familiar with the codebase. The `split()` function means "*allow* it to split here" but doesn't require it. It's inserting a potential split, not a mandatory one. That's why the preceding comment says "if". The `split()` call works for both cases.

PR #675, opened 2018-03-08
Clean up some analyzer imports

bwilkerson:
I am in the process of looking at clients of analyzer to determine what needs to be added to the public API in order to remove imports of internal code. In the process, I am sometimes cleaning up existing imports in order to make that task easier.

This replaces an import that exposes a very large API with more specific imports that expose a much smaller portion of the API.

I won't be offended if you'd prefer not to accept it.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -4,8 +4,10 @@
 
 library dart_style.src.source_visitor;
 
-import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/analyzer.dart' show ParameterKind;
+import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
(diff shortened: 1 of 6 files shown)


lib/src/source_visitor.dart
```diff
@@ -4,8 +4,10 @@
 
 library dart_style.src.source_visitor;
 
-import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/analyzer.dart' show ParameterKind;
```
munificent: Is the `show` clause needed to avoid collisions in names exported both by this and the imports below? If so, can you leave a comment to that effect. If it isn't needed to avoid a collision, let's go ahead and omit the `show`.

kevmoo:
@bwilkerson – are you going to run w/ this?

munificent:
Guessing this is stale...

PR #685, opened 2018-04-26
Fix an issue when running in Dart 2.

munificent:
Thanks for mentioning this! It was the only issue I found. All tests pass on the VM with `--preview-dart-2` now.


keertip:
Thanks Nate, Bob! Saw this and the other one that's already fixed in google3 today! Fixed before I could even report it 😀

PR #699, opened 2018-06-14
Dogfood the new "--fix" stuff.

munificent:
Use "=" for named parameter default separators. Remove unnecessary "new"
and "const". Reformat with the latest rules.

This does mean that dartfmt now only runs in --preview-dart-2, but I
think we're close to the point in time where that's OK. It has been
running in the Dart SDK with "--preview-dart-2" for a while.

### DIFF
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -57,7 +57,7 @@ class ArgumentListVisitor {
       _arguments._blocks.isNotEmpty || _functions != null;
 
   factory ArgumentListVisitor(SourceVisitor visitor, ArgumentList node) {
-    return new ArgumentListVisitor.forArguments(
+    return ArgumentListVisitor.forArguments(
         visitor, node.leftParenthesis, node.rightParenthesis, node.arguments);
   }
 
@@ -147,14 +147,8 @@ class ArgumentListVisitor {
 
     if (functionsStart == null) {
       // No functions, so there is just a single argument list.
-      return new ArgumentListVisitor._(
-          visitor,
-          leftParenthesis,
-          rightParenthesis,
-          arguments,
-          new ArgumentSublist(arguments, arguments),
-          null,
-          null);
+      return ArgumentListVisitor._(visitor, leftParenthesis, rightParenthesis,
+          arguments, ArgumentSublist(arguments, arguments), null, null);
     }
 
     // Split the arguments into two independent argument lists with the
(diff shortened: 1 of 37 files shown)


lib/src/argument_list_visitor.dart
```diff
@@ -147,14 +147,8 @@ class ArgumentListVisitor {
...
-      return new ArgumentListVisitor._(
-          visitor,
-          leftParenthesis,
-          rightParenthesis,
-          arguments,
-          new ArgumentSublist(arguments, arguments),
-          null,
-          null);
+      return ArgumentListVisitor._(visitor, leftParenthesis, rightParenthesis,
+          arguments, ArgumentSublist(arguments, arguments), null, null);
```
a14n: Unlike the [below `ArgumentListVisitor._` instantiation (line 160)](https://github.com/dart-lang/dart_style/pull/699/files#diff-0aa3c812b9d05341dfe503408b560ed6R160) it's on 2 lines instead of one argument per line. It's a little strange.

munificent: Yup. If an argument list fits on two lines, dartfmt packs it. More than two, and it goes to one argument per line.

This invocation, without the `new` is now just short enough to fit everything in two lines.

kevmoo:
Careful: should only publish this package w/ an SDK constraint >= the SDK where the VM has --preview-dart-2 enabled by default

natebosch:
> Careful: should only publish this package w/ an SDK constraint >= the SDK where the VM has --preview-dart-2 enabled by default

No need, the `dartfmt` binary is already run with the Dart 2 vm in the SDK. We don't have a supported external use case (other than developers on this package) for running form source where we'd be able to choose dart 2 or not.

kevmoo:
Lot's of things depend on pkg:dart_style – like source_gen

Don't want `pub upgrade` to start blowing up dart_style because they are
not on the Dart2-runtime version of the SDK

srawlins:
What about internal use cases?

natebosch:
ugh, yeah how did I forget out the people who import `dart_style`...

If we publish this today we'll break a lot of users and turn a lot of Travis status red...

munificent:
> Don't want `pub upgrade` to start blowing up dart_style because they are not on the Dart2-runtime version of the SDK

dart_style has an SDK constraint that requires a Dart SDK new enough where optional `new`/`const` is supported. Users will need to actually run their program with `--preview-dart-2` in order to get that behavior, at least until that's the default.

But they shouldn't find themselves in a situation where they have this version of dart_style but *can't* run with optional `new`/`const`.

But, yeah, I agree, publishing this will cause redness, which is why I'm starting to get some attention to this now. :)

PR #709, opened 2018-07-06
Don't split inside string interpolation.

munificent:
When it happens, it's usually a sign that you should split the string
yourself to avoid it. But it still looks terrible. And in multi-line
strings, you may *want* to have lines longer than the page width, in
which case the splitting is really bad.

### DIFF
--- a/lib/src/chunk_builder.dart
+++ b/lib/src/chunk_builder.dart
@@ -93,14 +93,15 @@ class ChunkBuilder {
   /// block.
   bool _firstFlushLeft = false;
 
-  /// Splitting is completely disabled inside string interpolation. We do still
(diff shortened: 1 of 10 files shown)


lib/src/chunk_builder.dart
```diff
@@ -93,6 +93,16 @@ class ChunkBuilder {
   /// block.
   bool _firstFlushLeft = false;
 
+  /// Splitting is completely disabled inside string interpolation. We do still
```
natebosch: [nit] blank line after the first sentence - maybe pick a different first sentence too?

munificent: Done.

PR #724, opened 2018-08-16
Ignore all non syntactic errors

danrubel:
The fasta parser produces some semantic errors (e.g. final variable not initialized) which the dart formatter can safely ignore.


munificent:
Thanks!

PR #725, opened 2018-08-16
Update map literal formatting test

danrubel:
This test case is invalid Dart syntax that the old analyzer parser let slide by.
From section 17.2 of the language spec...

> An expression statement consists of an expression other than a non-constant map literal (16.8) that has no explicit type arguments. The restriction on maps is designed to resolve an ambiguity in the grammar, when a statement begins with {.

### DIFF
--- a/test/splitting/invocations.stmt
+++ b/test/splitting/invocations.stmt
@@ -248,9 +248,9 @@ someVeryLongTargetFunction(
   element
 ].someLongMethod();
 >>> do not split on "." when target is map
-{"key": "value", "another": "another value"}.someLongMethod();
+<String,String>{"key": "value", "another": "another value"}.someLongMethod();
 <<<
-{
+<String, String>{
   "key": "value",
   "another": "another value"
 }.someLongMethod();


test/splitting/invocations.stmt
```diff
@@ -248,9 +248,9 @@ someVeryLongTargetFunction(
   element
 ].someLongMethod();
 >>> do not split on "." when target is map
-{"key": "value", "another": "another value"}.someLongMethod();
+<String,String>{"key": "value", "another": "another value"}.someLongMethod();
```
munificent: Is this change necessary? The expression in the expression statement isn't a map literal, it's a method invocation (`.someLongMethod()`) whose receiver is a map literal. Is that not valid?

danrubel: Correct. Although its valid as an expression, it's not valid as an expression statement.

bwilkerson:
@leafpetersen @lrhn For clarification of the spec.

leafpetersen:
The spec language doesn't seem to me to cover this case.  As best as I can tell, none of our release mode compilers (dart2js, dart) support this or any related things that I've tests (cascade, subscript, update), so I think it would be de facto non-breaking to make the spec reflect what seems to be the intention.  That is:

```
An {\em expression statement} consists of an expression other than a non-constant map literal (\ref{maps}) that has no explicit type arguments.
```

Should probably be:

```
An {\em expression statement} consists of an expression that does not begin with a non-constant map literal (\ref{maps}) that has no explicit type arguments.
```

leafpetersen:
@munificent Do you mind if this gets landed to unblock the analyzer/parser integration?  We'll have to sort out the spec decision and if necessary all of the implementations separately, and I don't see any reason to block on this now.

lrhn:
We should specify what the implementations do.
The grammar works either way - if a block ends with `}.foo();` then it's not a code block, so it must be a map, and the grammar is not ambiguous.  (It might not even be ambiguous without the trailing call, unless the block is empty, since any statement inside will be recursively distinguishable from an expression by, eventually, ending in a semicolon).

Allowing leading maps does make parsing need arbitrary look-ahead to determine whether something starting with `{foo: verylargeexpression` is a map or a code block starting with a labeled statement. (If it's followed by ';' it's a statement, if it's followed by '}', it's a map.

So, let's just keep the current behavior and codify it in the spec.
(The "begins with *expression*" is too vague for my part. I'd prefer to just say that an expression statement cannot start with a '{' character).

leafpetersen:
For the record, in SDK 1.13 (which is before the VM and the dart2js switched to the CFE, I believe), all of these examples were rejected by the VM and dart2js:

```dart
void main() {
  {"foo" : "bar"}.containsKey("foo");
  {"foo" : "bar"}..containsKey("foo");
  {"foo" : "bar"}["foo"];
  {"foo" : "bar"}["foo"] = "baz";
}
```

lrhn:
Then I see no problem in keeping them disallowed, let's just do that. I'll fix the specification.
https://dart-review.googlesource.com/c/sdk/+/71247

PR #734, opened 2018-09-04
add fix for doc comments

### LINKED ISSUE #730: Fix to convert from /** */ comments to ///
Since `///` comments is the style-guide recommendation for doc-comments, maybe add a fix to convert `/** */`-style doc-comments to `///`-style.

It's not absolutely trivial since `///` allows one less character per line, so converting existing documentation will end up with 81 character lines. The fix might need to print a notice with file and line-number for each line that is now too long, so the user can go back and fix that.

Maybe it's not really necessary because everybody already switched to `///` (and I'll just go back to my Perl script).

a14n:
Fix #730

### DIFF
--- a/lib/src/chunk_builder.dart
+++ b/lib/src/chunk_builder.dart
@@ -12,6 +12,7 @@ import 'nesting_builder.dart';
 import 'nesting_level.dart';
 import 'rule/rule.dart';
 import 'source_code.dart';
+import 'style_fix.dart';
 import 'whitespace.dart';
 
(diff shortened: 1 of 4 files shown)


munificent:
This looks great!

I ran it on a big corpus and the results look pretty good to me. I did make a few minor tweaks here: https://github.com/dart-lang/dart_style/tree/revise-doc-comment-fix. That builds on your PR. Once that gets reviewed and landed, this will be in too.

Thanks!

munificent:
I merged this in with some additional refinements. Thanks!

PR #735, opened 2018-09-11
Revise doc comment fix

munificent:


munificent:
@a14n has committed stuff plenty of times and is A-OK on the CLA.

lib/src/chunk_builder.dart
```diff
@@ -333,6 +341,48 @@ class ChunkBuilder {
     preserveNewlines(linesBeforeToken);
   }
 
+  /// Writes the text of [comment].
+  ///
+  /// If it's a JavaDoc comment that should be fixed to use `///`, fixes it.
+  void _writeCommentText(SourceComment comment) {
+    if (_formatter.fixes.contains(StyleFix.docComments)) {
+      var match = _javaDocComment.firstMatch(comment.text);
+      if (match != null) {
```
natebosch: [optional] invert the condition and short circuit return would reduce nesting in the rest of the method.

Same with the condition on line 348.

lib/src/chunk_builder.dart
```diff
@@ -333,6 +341,48 @@ class ChunkBuilder {
...
+        if (lines.isNotEmpty && lines.last.trim().isEmpty) lines.removeLast();
+
+        // Remove a leading "*" from the middle lines.
+        for (var i = 0; i < lines.length; i++) {
+          var line = lines[i];
+          var match = _javaDocLine.firstMatch(line);
+          if (match != null) {
+            line = match.group(1);
+          } else {
+            line = line.trimLeft();
```
natebosch: we're potentially changing indentation for lines that don't have a leading `*`. I think that's probably fine if that was the intention, but we can maybe call this out in a comment somewhere.

munificent: Done.

PR #740, opened 2018-09-14
Let users control name shown in errors when reading from stdin.

### LINKED ISSUE #739: Allow specifying the path shown in errors when reading from stdin
Right now, if a parse error occurs when you run dartfmt using stdin as its input source, it always shows the path for the error as `<stdin>`. Some tools drive dartfmt using stdin but do associate the source with a path and it would be helpful to show that path in the error.

Add a command-line option to let users specify the string shown.

munificent:
Fix #739.

### DIFF
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -173,7 +173,6 @@ void main(List<String> args) {
     }
   }
 
-  var stdinName = argResults["stdin-name"];
   if (argResults.wasParsed("stdin-name") && !argResults.rest.isEmpty) {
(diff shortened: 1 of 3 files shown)


bin/format.dart
```diff
@@ -170,14 +173,19 @@ void main(List<String> args) {
     }
   }
 
+  var stdinName = argResults["stdin-name"];
```
natebosch: [nit] consider adding a type on the left, or making it explicit with `as String` on the right. As is the implicit cast happens on line 188 and if it fails the error would be nicer here.

munificent: Done.

PR #741, opened 2018-09-17
Add more tests for doc comment fix and fix an edge case.

munificent:
A line that didn't have a leading `*` but did have a `*` elsewhere in
the line would have everything before the `*` trimmed.

cc @lrhn

### DIFF
--- a/test/fixes/doc_comments.stmt
+++ b/test/fixes/doc_comments.stmt
@@ -111,3 +111,49 @@ m() {}
  * STUFF
  *******/
 m() {}
+>>> nested comment
+/**
+ * Floo the grumshack.
+ *
+ * Example:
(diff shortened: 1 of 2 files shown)


test/fixes/doc_comments.stmt
```diff
@@ -110,4 +110,50 @@ m() {}
...
+ * ```
+*/
+m() {}
+<<<
+/// Floo the grumshack.
+///
+/// Example:
+/// ```dart
+/// /** Doc comment */
+/// var grumshack = getGrumshack();
```
natebosch: https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

😝

PR #743, opened 2018-09-19
Add --fix-typedefs to Dart formatter.

lrhn:
WDYT?
Would it be better to build a new node for the converted syntax and then visit that node, instead of physically building the output? It would ensure consistent formatting.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -1385,7 +1385,7 @@ class SourceVisitor extends ThrowingAstVisitor {
   /// Writes a `Function` function type.
   ///
   /// Used also by a fix, so there may not be a [functionKeyword].
-  /// In that case [syntheticFunctionPosition] should be the source position
+  /// In that case [functionKeywordPosition] should be the source position
   /// used for the inserted "Function" text.
   void _writeGenericFunctionType(
       AstNode returnType,
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -1315,19 +1315,37 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     if (_formatter.fixes.contains(StyleFix.typedefs)) {
       _simpleStatement(node, () {
+        // Inlined visitGenericTypeAlias
         token(node.typedefKeyword);
         space();
+
+        // If the typedef's type parameters split, split after the "=" too,
+        // mainly to ensure the function's type parameters and parameters get
+        // end up on successive lines with the same indentation.
+        builder.startRule();
+
         visit(node.name);
+
         visit(node.typeParameters);
+        split();
+
+        _writeText("=", (node.returnType ?? node.name).offset);
+        builder.endRule();
+
         space();
-        _writeText("=", node.returnType.offset);
-        space();
-        visit(node.returnType, after: space);
-        space();
+
+        // Inlined visitGenericFunctionType
+        builder.startLazyRule();
+        builder.nestExpression();
+
+        visit(node.returnType, after: split);
         _writeText("Function", node.name.offset);
-        // Recursively convert function-arguments to Function typed.
+
+        builder.unnest();
+        builder.endRule();
+        // Recursively convert function-arguments to Function syntax.
         _insideNewTypedefFix = true;
-        visit(node.parameters);
+        _visitParameterSignature(null, node.parameters);
         _insideNewTypedefFix = false;
       });
       return;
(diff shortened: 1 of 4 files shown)


lib/src/source_visitor.dart
```diff
@@ -1305,6 +1313,44 @@ class SourceVisitor extends ThrowingAstVisitor {
   visitFunctionTypeAlias(FunctionTypeAlias node) {
     visitMetadata(node.metadata);
 
+    if (_formatter.fixes.contains(StyleFix.typedefs)) {
+      _simpleStatement(node, () {
+        // Inlined visitGenericTypeAlias
```
munificent: All this copy/pasted code is a big concern for me. I try hard to minimize that because it tends to diverge over time and then similar code constructs end up formatting differently in different contexts.

Can you hoist this code out to private methods that are called in both places? If you scroll down past `visitYieldStatement()`, you'll see a lot of other methods that exist for this reason.

Here and below.

lib/src/source_visitor.dart
```diff
@@ -1337,8 +1407,13 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     builder.unnest();
     builder.endRule();
-
+    // Old style function parameters or single-identifier-meaning-name
+    // cannot occur inside the `Function` type expression.
+    // If they do anyway, retain them as an error.
+    bool wasInsideNewTypedefFix = _insideNewTypedefFix;
```
munificent: `bool` -> `var`

Is there a reason you don't just go ahead and fix them?

lib/src/source_visitor.dart
```diff
@@ -1836,11 +1911,16 @@ class SourceVisitor extends ThrowingAstVisitor {
       modifier(node.keyword);
 
       visit(node.type);
+      bool hasType = node.type != null;
+      if (_insideNewTypedefFix && !hasType) {
+        _writeText("dynamic", node.identifier.offset);
```
munificent: This needs a test for what happens to a comment before the identifier.

lib/src/style_fix.dart
```diff
@@ -18,11 +18,15 @@ class StyleFix {
   static const docComments = const StyleFix._(
       "doc-comments", 'Use triple slash for documentation comments.');
 
+  static const typedefs =
+      const StyleFix._("typedefs", "Use new typedef syntax for typedefs");
```
munificent: "." at end.

test/fixes/typedefs.unit
```diff
@@ -0,0 +1,49 @@
...
+>>> simple typedef
+typedef int Foo(int x);
+<<<
+typedef Foo = int Function(int x);
+>>> generic typedef
+typedef S Foo<S extends num>(S x);
+<<<
+typedef Foo<S extends num> = S Function(
+    S x);
+>>> named argument
```
munificent: This doesn't look like a named argument to me. :)

test/fixes/typedefs.unit
```diff
@@ -0,0 +1,49 @@
...
+typedef Foo = int Function(int);
+>>> simple typedef, no types
+typedef Foo(x);
+<<<
+typedef Foo = Function(dynamic x);
+>>> nested function types, no types
+typedef Foo(foo(x));
+<<<
+typedef Foo = Function(
+    Function(dynamic x) foo);
```
munificent: These look great. Also:

* Named parameters.
* Optional positional parameters.
* Metadata annotations on the typedef itself, and parameters.
* Comments before and after various tokens, in particular ones that move in the fix. It's not vital that the resulting comment location look *great*, we just need to make sure the comment isn't discarded.

lrhn:
Didn't address all comments yet, but refactored to avoid code duplication.

lib/src/source_visitor.dart
lrhn: Ick, good point.
Should it stay with the name, or should it get put before the type?
It's like metadata, so it could precede the entire typed parameter, but you *can* write `typedef F(foo(int /*c*/ bar));`. That's definitely where the current code puts it. 

Do you want to move the comment before the `dynamic` in this case?

test/fixes/typedefs.unit
lrhn: done

test/fixes/typedefs.unit
lrhn: Done.

lib/src/source_visitor.dart
lrhn: Done.

lib/src/source_visitor.dart
```diff
@@ -1835,14 +1863,24 @@ class SourceVisitor extends ThrowingAstVisitor {
...
 
-      visit(node.type);
+      bool hasType = node.type != null;
+      if (_insideNewTypedefFix && !hasType) {
+        // In function declarations and the old typedef syntax, you can have a
+        // parameter name without a type. In the new syntax, you can have a type
+        // without a name. Add "dynamic" in that case.
+
+        // Ensure comments on the identifier comes before the inserted type.
+        token(node.identifier.token, before: () {
```
lrhn: Here I am kind-of inlining `visitSimpleIdentifier`, which just calls `token(node.token)`.
Should I instead add an `{void before()}` or `{void beforeToken}` parameter to visitSimpleIdentifier and call that directly?

I do want to insert the dynamic between the comment and the token itself, which `token(..., before: ...)` does for me, but I can't put that extra aargument through the `visit`/`accept` procedure, and I don't want to store it in some object-global state.

lrhn:
All comments addressed. PTAL

PR #774, opened 2019-02-13
remove enableSetLiterals flag

danrubel:
Now that set literals are enabled by default, this is a step towards removing this feature's experimental flag from the analyzer/parser base.


danrubel:
Done

PR #780, opened 2019-03-11
Stop using deprecated AST classes/methods.

stereotype441:
As part of implementing the "UI as code" feature
(https://github.com/dart-lang/language/blob/master/accepted/future-releases/unified-collections/feature-specification.md)
the analyzer is deprecating the following classes and methods:

- ForStatement and ForEachStatement (ForStatement2 is now used for
  both kinds of loops)

- ListLiteral.elements (use elements2, which has a more general return
  type)

- AstVisitor.visitForEachStatement and AstVisitor.visitForStatement
  (visitForStatement2 is now used for both kinds of loops)

- AstVisitor.visitMapLiteral and AstVisitor.visitSetLiteral
  (visitSetOrMapLiteral is now used for both kinds of literals).

See deprecation CL: https://dart-review.googlesource.com/c/sdk/+/95665/

Analyzer clients need to be adapted to stop using the deprecated
classes and methods.

### DIFF
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -504,9 +504,7 @@ class ArgumentSublist {
     // TODO(rnystrom): Should we step into parenthesized expressions?
 
     if (expression is ListLiteral) return expression.leftBracket;
-    // TODO(rnystrom): should we return expression.leftBracket for sets as well?
-    if (expression is SetOrMapLiteral && !expression.isSet)
-      return expression.leftBracket;
+    if (expression is SetOrMapLiteral) return expression.leftBracket;
     if (expression is SingleStringLiteral && expression.isMultiline) {
       return expression.beginToken;
     }
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -504,7 +504,9 @@ class ArgumentSublist {
     // TODO(rnystrom): Should we step into parenthesized expressions?
 
     if (expression is ListLiteral) return expression.leftBracket;
-    if (expression is MapLiteral) return expression.leftBracket;
+    // TODO(rnystrom): should we return expression.leftBracket for sets as well?
+    if (expression is SetOrMapLiteral && !expression.isSet)
+      return expression.leftBracket;
     if (expression is SingleStringLiteral && expression.isMultiline) {
       return expression.beginToken;
     }
--- a/lib/src/call_chain_visitor.dart
+++ b/lib/src/call_chain_visitor.dart
@@ -311,8 +311,7 @@ class CallChainVisitor {
 
     // Don't split right after a collection literal.
     if (expression is ListLiteral) return false;
-    // TODO(rnystrom): should we return false for sets as well?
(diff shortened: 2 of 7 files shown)


lib/src/argument_list_visitor.dart
```diff
@@ -504,7 +504,9 @@ class ArgumentSublist {
     // TODO(rnystrom): Should we step into parenthesized expressions?
 
     if (expression is ListLiteral) return expression.leftBracket;
-    if (expression is MapLiteral) return expression.leftBracket;
+    // TODO(rnystrom): should we return expression.leftBracket for sets as well?
+    if (expression is SetOrMapLiteral && !expression.isSet)
```
munificent: Yes, just remove the `&& !expression.isSet` part and allow this to match sets too. (The fact that it doesn't was an oversight.)

lib/src/call_chain_visitor.dart
```diff
@@ -311,7 +311,8 @@ class CallChainVisitor {
 
     // Don't split right after a collection literal.
     if (expression is ListLiteral) return false;
-    if (expression is MapLiteral) return false;
+    // TODO(rnystrom): should we return false for sets as well?
```
munificent: Yes, you can remove the `isSet` check here too. (When I added support for set literals, I didn't think to search for `MapLiteral` everywhere and see where else sets should get special handling.)

lib/src/source_visitor.dart
```diff
@@ -503,7 +503,8 @@ class SourceVisitor extends ThrowingAstVisitor {
   ///     ]..addAll(numbers);
   bool _isCollectionLike(Expression expression) {
     if (expression is ListLiteral) return false;
-    if (expression is MapLiteral) return false;
+    // TODO(rnystrom): should we return false for sets as well?
+    if (expression is SetOrMapLiteral && !expression.isSet) return false;
```
munificent: Yep.

lib/src/source_visitor.dart
```diff
@@ -2173,14 +2177,18 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     // Don't allow a split between a name and a collection. Instead, we want
     // the collection itself to split, or to split before the argument.
-    if (node.expression is ListLiteral || node.expression is MapLiteral) {
+    // TODO(rnystrom): do we want the "if" test below to apply to sets as well
+    // as maps?
```
munificent: Yes.

lib/src/source_visitor.dart
```diff
@@ -2700,7 +2708,10 @@ class SourceVisitor extends ThrowingAstVisitor {
   ///         new SomeBuilderClass()..method()..method();
   int _assignmentCost(Expression rightHandSide) {
     if (rightHandSide is ListLiteral) return Cost.assignBlock;
-    if (rightHandSide is MapLiteral) return Cost.assignBlock;
+    // TODO(rnystrom): or do we want the "if" test on the line below to apply to
+    // both maps and sets?
```
munificent: Yes.

stereotype441:
@munificent PTAL to verify that the test changes are ok.  I think they should be but I'd rather have a second pair of eyes double check me :)

PR #783, opened 2019-03-22
Formatting support for spread and if control flow elements.

munificent:
(I'll do "for" in a separate commit.)


munificent:
Good idea. Added a few more tests around different expressions inside a spread.

PR #794, opened 2019-03-31
no extra-indent of expression function as argument with trailing comma

a14n:
Actual:

```dart
      onTap: () {
        Navigator.of(context).push(CupertinoPageRoute<void>(
          title: colorName,
          builder: (BuildContext context) => Tab1ItemPage(
                color: color,
                colorName: colorName,
                index: index,
              ),
        ));
      },
```

Expected:

```dart
      onTap: () {
        Navigator.of(context).push(CupertinoPageRoute<void>(
          title: colorName,
          builder: (BuildContext context) => Tab1ItemPage(
            color: color,
            colorName: colorName,
            index: index,
          ),
        ));
      },
```

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -1022,11 +1022,24 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     if (_isInLambda(node)) builder.endSpan();
 
-    builder.startBlockArgumentNesting();
+    // If this function invocation appears in an argument list with trailing
+    // comma, don't add extra nesting to preserve normal indentation.
+    var isArgWithTrailingComma = false;
+    var parent = node.parent;
+    if (parent is FunctionExpression) {
+      var argList = parent?.parent;
+      if (argList is NamedExpression) argList = argList.parent;
+      if (argList is ArgumentList &&
+          argList.arguments.last.endToken.next.type == TokenType.COMMA) {
+        isArgWithTrailingComma = true;
+      }
+    }
+
+    if (!isArgWithTrailingComma) builder.startBlockArgumentNesting();
     builder.startSpan();
     visit(node.expression);
     builder.endSpan();
-    builder.endBlockArgumentNesting();
+    if (!isArgWithTrailingComma) builder.endBlockArgumentNesting();
 
     if (node.expression is BinaryExpression) builder.endRule();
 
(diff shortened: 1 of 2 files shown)


kevmoo:
I *think* this is by design. @munificent will have thoughts

a14n:
>I think this is by design.

As this looks similar to #382 (that is not closed) this change seems doable.

(Context: I try to apply `dartfmt` on `flutter/examples` and see what's the smallest set of changes to be done to use it (or a small customized version) on flutter examples. Ideally those changes should be part of the current dartfmt to limit divergencies)

a14n:
The same behaviour could be done on function declarations and method declarations (I did it in https://github.com/a14n/dart_style/commit/7a51442c63fdd269b7421e0073256b3ab3ae4706 - happy to include it in this PR).

munificent:
I'm a little swamped with other stuff to take a look at these right now, but I'll try to circle back when I can.

I have some vague concerns that this will do bad things on more complex cases, so I'll need to run it on a bigger corpus and see what the results look like. The output you have here looks nice to me. It's mostly a question of making sure it doesn't regress other things.

a14n:
The only concern I have is when the function split after `=>`:

```dart
    onTap: () {
      Navigator.of(context).push(CupertinoPageRoute<void>(
        title: colorName,
        builder_______________________________________________: (BuildContext context) =>
            Tab1ItemPage(
          color: color,
          colorName: colorName,
          index: index,
        ),
      ));
    },
```

Ideally in this case I'd like to have:

```dart
    onTap: () {
      Navigator.of(context).push(CupertinoPageRoute<void>(
        title: colorName,
        builder_______________________________________________: (BuildContext context) =>
            Tab1ItemPage(
              color: color,
              colorName: colorName,
              index: index,
            ),
      ));
    },
```

But I don't manage to write the good code to handle this case and to produce what I'd like :-(

kasperpeulen:
I think best would be this:

```dart
    onTap: () {
      Navigator.of(context).push(CupertinoPageRoute<void>(
        title: colorName,
        builder_______________________________________________: (BuildContext context) =>
          Tab1ItemPage(
            color: color,
            colorName: colorName,
            index: index,
          ),
      ));
    },
```

a14n:
@munificent friendly reminder

munificent:
I tested this out on a big corpus locally. Overall, I like the output. However, like you note, the results are pretty weird when it needs to indent before the body of the function, as in:

```dart
  function(
    (parameter1, parameter2,
            parameter3) =>
        P(
      p,
    ),
  );
```

Cases like this seem pretty rare in the wild though, so I think it's still worth doing.

PR #800, opened 2019-04-04
Better indentation of adjacent strings in argument lists.

### LINKED ISSUE #799: adjacent  concat strings in arg lists lost indention - Regression with 1.2.5 (SDK 2.2.1-dev.3.0)
```diff
 @ShouldThrow(
   'Could not generate `toJson` code for `watch`.\n'
-      'None of the provided `TypeHelper` instances support the defined type.',
+  'None of the provided `TypeHelper` instances support the defined type.',
   configurations: ['default'],
 )
```

munificent:
Fix #799. See the CHANGELOG diff for details. The relevant examples are:

1.2.4:

```dart
function(
    "adjacent"
    "string");

function(
  "adjacent"
      "string", // Not great.
);

function(
    "adjacent"
    "string",
    "another");

function(
  "adjacent"
      "string",
  "another",
);
```

1.2.5:

```dart
function(
    "adjacent"
        "string"); // Regression, but consistent with trailing comma.

function(
  "adjacent"
      "string",
);

function(
    "adjacent"
        "string", // Uintentional, but an improvement.
    "another");

function(
  "adjacent"
      "string", // Unchanged.
  "another",
);
```

Now:

```dart
function(
    "adjacent"
    "string"); // No indent, since not needed to distinguish args.

function(
  "adjacent"
  "string", // Consistent with non-trailing.
);

function(
    "adjacent"
        "string", // Intentional, distinguishes from next string.
    "another");

function(
  "adjacent"
      "string", // Consistent with non-trailing.
  "another",
);
```

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -186,9 +186,72 @@ class SourceVisitor extends ThrowingAstVisitor {
   }
 
   visitAdjacentStrings(AdjacentStrings node) {
+    // We generally want to indent adjacent strings because it can be confusing
+    // otherwise when they appear in a list of expressions, like:
+    //
+    //     [
+    //       "one",
+    //       "two"
+    //       "three",
+    //       "four"
+    //     ]
+    //
+    // Especially when these stings are longer, it can be hard to tell that
+    // "three" is a continuation of the previous argument.
+    //
+    // However, the indentation is distracting in argument lists that don't
+    // suffer from this ambiguity:
+    //
+    //     test(
+    //         "A very long test description..."
+    //             "this indentation looks bad.", () { ... });
+    //
+    // To balance these, we omit the indentation when an adjacent string
+    // expression is the only string in an argument list.
+    var shouldNest = true;
+
+    isString(AstNode node) => node is StringLiteral || node is AdjacentStrings;
+
+    var parent = node.parent;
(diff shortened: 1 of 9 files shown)


lib/src/source_visitor.dart
```diff
@@ -186,9 +186,72 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+    //
+    //     test(
+    //         "A very long test description..."
+    //             "this indentation looks bad.", () { ... });
+    //
+    // To balance these, we omit the indentation when an adjacent string
+    // expression is the only string in an argument list.
+    var shouldNest = true;
+
+    isString(AstNode node) => node is StringLiteral || node is AdjacentStrings;
```
a14n: `AdjacentStrings` been a `StringLiteral` you can keep only the left hand side.

```
/// A string literal expression.
///
///    stringLiteral ::=
///        [SimpleStringLiteral]
///      | [AdjacentStrings]
///      | [StringInterpolation]
///
/// Clients may not extend, implement or mix-in this class.
abstract class StringLiteral implements Literal {
  /// Return the value of the string literal, or `null` if the string is not a
  /// constant string without any string interpolation.
  String get stringValue;
}
```

munificent: Fixed in 1.2.7. Thanks for pointing that out!

PR #805, opened 2019-04-08
handle instance creations and method calls with trailing comma like block in argument list

### LINKED ISSUE #804: method call with single instance creation (with trailing comma)
Acutal:

```dart
Navigator.push(
    context,
    MaterialPageRoute<DismissDialogAction>(
      builder: (BuildContext context) => FullScreenDialogDemo(),
      fullscreenDialog: true,
    ));
```

expected:

```dart
Navigator.push(context, MaterialPageRoute<DismissDialogAction>(
  builder: (BuildContext context) => FullScreenDialogDemo(),
  fullscreenDialog: true,
));
```

a14n:
Fixes #804

### DIFF
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -229,6 +229,18 @@ class ArgumentListVisitor {
       expression = (expression as NamedExpression).expression;
     }
 
+    // Allow instance creations and method invocations with trailing commas
+    if (expression is InstanceCreationExpression &&
+        expression.argumentList.rightParenthesis.previous.type ==
+            TokenType.COMMA) {
+      return true;
+    }
+    if (expression is MethodInvocation &&
+        expression.argumentList.rightParenthesis.previous.type ==
+            TokenType.COMMA) {
+      return true;
+    }
+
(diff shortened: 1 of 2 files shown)


munificent:
I don't think we want to special case this. It only kicks in when you have a trailing comma argument list inside a non-trailing comma argument list, which is already a "user failure". We want users to use trailing commas consistently to get good formatted code, and I'd rather not add complexity to dartfmt to handle cases where they don't do that.

I really appreciate you hacking on a fix for this. I'm sorry I didn't find the time to look at it. Dart 2.3.0 and Google IO kept me busy. :)

a14n:
>which is already a "user failure"

I don't think so. IMHO trailing commas are user hints to drive the formatter's output. There are tons of code (inside Flutter code and elsewhere) where trailing comma is used without outer trailing commas. 

A really commun case (from the example in #804) `Navigator.push` is almost never used with trailing comma.

munificent:
Sorry, I should have been clearer. I think it's a reasonable style to have an argument list without trailing commas inside one that has them, *when the inner argument list fits on one line*. In other words, any time you have line splitting in argument lists, they should all split the same way. If you have trailing commas inside a non-trailing comma arg list, the inner one will force both to split and the two will have different styles.

PR #806, opened 2019-04-11
handle method with trailing comma like block in method chain

a14n:
This PR will produce:

```dart
padding: MediaQuery.of(context).removePadding(
  removeTop: true,
  removeLeft: true,
  removeRight: true,
).padding,
```

instead of:

```dart
padding: MediaQuery.of(context)
    .removePadding(
      removeTop: true,
      removeLeft: true,
      removeRight: true,
    )
    .padding,
```

### DIFF
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -503,6 +503,16 @@ class ArgumentSublist {
       return expression.beginToken;
     }
 
+    // instance creation or method invocation with trailing comma are like block
+    if (expression is InstanceCreationExpression &&
+        expression.argumentList.endToken.previous.type == TokenType.COMMA) {
+      return expression.argumentList.beginToken;
+    }
+    if (expression is MethodInvocation &&
+        expression.argumentList.endToken.previous.type == TokenType.COMMA) {
+      return expression.argumentList.beginToken;
+    }
+
     // Not a collection literal.
     return null;
   }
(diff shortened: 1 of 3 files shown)


munificent:
Is this similar to #804 where it only comes into play when you have a trailing comma argument list inside a non-trailing comma one?

a14n:
>Is this similar to #804 where it only comes into play when you have a trailing comma argument list inside a non-trailing comma one?

No it's not depending on the presence of outer trailing comma.

Argument list with trailing commas is formatted with every arguments on their own lines. Visually it's really like block formatting and this PR try to make the outputs similar:

```dart
    // with the current dartfmt and trailing comma
    MediaQuery.of(context)
        .removePadding(
          removeTop: true,
          removeLeft: true,
          removeRight: true,
        )
        .padding;

    // if removePadding had taken a map as parameter, here's the dartfmt output
    MediaQuery.of(context).removePadding({
      removeTop: true,
      removeLeft: true,
      removeRight: true,
    }).padding;

    // with this PR and original argument list
    MediaQuery.of(context).removePadding(
      removeTop: true,
      removeLeft: true,
      removeRight: true,
    ).padding;
```

Like #804 this PR is a try to reduce the difference between how Flutter code is formatted and the current output of dartfmt.

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #816, opened 2019-05-21
Use featureSet parameter to configure the parser.

stereotype441:
This helps prepare for the next major version of the analyzer package.


stereotype441:
> Can you bump dart_style's version to 1.2.8-dev as well? After that, 👍 .

Done.

PR #829, opened 2019-07-08
widen analyzer constraint (for 0.37.0)

pq:
/cc @munificent


pq:
Dropping in favor of #828.

PR #832, opened 2019-08-05
Format null-assertion operators.

munificent:
In the process, I reorganized CallChainVisitor to more gracefully
handle the various combinations of selectors. It was getting a little
too hard to understand.

That also highlighted that we didn't have good support for invocation
expressions inside call chains. So I added better formatting for that
too.

### DIFF
--- a/lib/src/call_chain_visitor.dart
+++ b/lib/src/call_chain_visitor.dart
@@ -33,7 +33,7 @@ import 'source_visitor.dart';
 /// This means visiting the AST from top down visits the selectors from right
 /// to left. It's easier to format that if we organize them as a linear series
 /// of selectors from left to right. Further, we want to organize it into a
-/// two-tier hierarchy. We have an outer list of method calls and propery
+/// two-tier hierarchy. We have an outer list of method calls and property
 /// accesses. Then each of those may have one or more postfix selectors
 /// attached: indexers, null-assertions, or invocations. This mirrors how they
 /// are formatted.
@@ -251,7 +251,7 @@ class CallChainVisitor {
       _disableRule();
 
       for (var blockCall in _blockCalls) {
-        blockCall.writeBlockCall(_visitor);
+        blockCall.write(this);
       }
 
       // If there is a hanging call after the last block, write it without any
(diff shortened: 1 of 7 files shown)


lib/src/call_chain_visitor.dart
```diff
@@ -13,8 +13,30 @@ import 'rule/rule.dart';
...
+///       /  \
+///     obj   1
+///
+/// This means visiting the AST from top down visits the selectors from right
+/// to left. It's easier to format that if we organize them as a linear series
+/// of selectors from left to right. Further, we want to organize it into a
+/// two-tier hierarchy. We have an outer list of method calls and propery
+/// accesses. Then each of those may have one or more postfix selectors
+/// attached: indexers, null-assertions, or invocations. This mirrors how they
+/// are formatted.
```
nshahan: I love this comment! 🙏

pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
```
nshahan: How does the SDK constraint work with this package and the analyzer dep? Would everything here technically work SDK v2.3.0 installed?

lib/src/call_chain_visitor.dart
```diff
@@ -460,3 +401,197 @@ class CallChainVisitor {
...
+    visitor._beforeMethodArguments(this);
+
+    visitor._visitor.builder.nestExpression();
+    visitor._visitor.visit(_node.typeArguments);
+    visitor._visitor
+        .visitArgumentList(_node.argumentList, nestExpression: false);
+    visitor._visitor.builder.unnest();
+  }
+
+  void writeBlockCall(SourceVisitor visitor) {
```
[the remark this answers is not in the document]
munificent: Uh, apparently there's no good reason. This code was basically already in place before this change, I just moved it around. But I just tried calling `write()` and it does the same thing. Modulo a couple of bugs it uncovered, which I also fixed and added tests for.

pubspec.yaml
munificent: I *think* so, yes.

munificent:
> All the new syntax appears as I would expect. Is there any more ceremony required when there are changes to the way the existing syntax is formatted?

In general, no. Most tooling that requires dartfmt to be run only requires it on files that are actually touched by a given CL/commit/patch, so when formatting of existing code changes, we just let users incrementally passively migrate. So far, that's seemed to work fine.

For this case in particular, I don't know if much actual code in the wild will be affected. Aside from the new NNBD syntax (which obviously doesn't exist in the wild yet), the only changes involve subscripts nested inside method chains. The reason that case wasn't handled particularly gracefully is because I apparently never noticed any examples of it when I tested method chain formatting even on very large corpora. So I'd be surprised if this affects more than a tiny fraction of real code.

PR #834, opened 2019-08-09
Add --fix-single-cascade-statements flag.

cpelling:
Automatically cleans up violations of the lint avoid_single_cascade_in_expression_statements.

Before:
```
  o..m = x;
```
After:
```
  o.m = x;
```

Adds parentheses when necessary to avoid changing operator precedence, as in the following examples:

Before:
```
  a as A..x = 42;
```
After:
```
  (a as A).x = 42;
```

Before:
```
  await someFuture
    ..doThing();
```
After:
```
  (await someFuture).doThing();
```

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -5,6 +5,7 @@
 library dart_style.src.source_visitor;
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/standard_ast_factory.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/src/generated/source.dart';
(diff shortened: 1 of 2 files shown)


lib/src/source_visitor.dart
```diff
@@ -1102,7 +1103,183 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+  /// be a CascadeExpression) before removing the cascade.
+  void _fixCascadeByParenthesizingTarget(ExpressionStatement statement) {
+    CascadeExpression cascade = statement.expression;
+    assert(cascade.cascadeSections.length == 1);
+
+    // Write any leading comments and whitespace immediately, as they should
+    // precede the new opening parenthesis, but then prevent them from being
+    // written again after the parenthesis.
+    writePrecedingCommentsAndNewlines(cascade.target.beginToken);
+    _suppressPrecedingCommentsAndNewLines.add(cascade.target.beginToken);
```
munificent: Instead of suppressing comments on the old token, is it feasible to detach them and reattach them to the new `(` token?

munificent:
Actually, I can take a stab at writing tests for this.

munificent:
Hi, I added a new commit onto this with tests and a few tweaks to match the formatter's internal style here. https://github.com/dart-lang/dart_style/pull/871

Can you take a look and tell me what you think?

cpelling:
Looks great, thanks! LGTM.

PR #835, opened 2019-08-12
Optimize _looksLikeClassName() in AOT builds.

munificent:
The AOT compiler does not currently compile regexes to machine code,
which makes them significantly slower. The regex in this function is
perf critical and makes the dart_style benchmark 65% slower on AOT than
on the JIT.

This fixes some of that regression.

Before:

```
// JIT
Best   : 4.93ms ========================
Best   : 4.83ms ========================
Best   : 4.80ms ========================

// AOT
Best   : 8.33ms =========================================
Best   : 8.43ms ==========================================
Best   : 8.37ms =========================================
```

After:

```
// JIT
Best   : 4.93ms ========================
Best   : 4.83ms ========================
Best   : 4.80ms ========================

// AOT
Best   : 6.93ms ==================================
Best   : 6.87ms ==================================
Best   : 6.77ms =================================
```

cc @mit-mit @mkustermann

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -94,17 +94,21 @@ class SourceVisitor extends ThrowingAstVisitor {
     const lowerA = 97;
     const lowerZ = 122;
 
-    var firstChar = name.codeUnitAt(0);
+    var start = 0;
+    var firstChar = name.codeUnitAt(start++);
 
     // It can be private.
-    if (firstChar == underscore) firstChar = name.codeUnitAt(1);
+    if (firstChar == underscore) {
+      if (name.length == 1) return false;
+      firstChar = name.codeUnitAt(start++);
+    }
 
     // It must start with a capital letter.
     if (firstChar < capitalA || firstChar > capitalZ) return false;
 
     // And have at least one lowercase letter in it. Otherwise it could be a
     // SCREAMING_CAPS constant.
-    for (var i = 0; i < name.length; i++) {
+    for (var i = start; i < name.length; i++) {
       var char = name.codeUnitAt(i);
       if (char >= lowerA && char <= lowerZ) return true;
     }
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -23,8 +23,6 @@ import 'source_code.dart';
 import 'style_fix.dart';
 import 'whitespace.dart';
 
-final _capitalPattern = RegExp(r"^_?[A-Z].*[a-z]");
-
 /// Visits every token of the AST and passes all of the relevant bits to a
 /// [ChunkBuilder].
 class SourceVisitor extends ThrowingAstVisitor {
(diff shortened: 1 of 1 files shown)


lib/src/source_visitor.dart
```diff
@@ -76,14 +74,42 @@ class SourceVisitor extends ThrowingAstVisitor {
...
     if (name == "bool") return true;
     if (name == "double") return true;
     if (name == "int") return true;
     if (name == "num") return true;
 
-    return _capitalPattern.hasMatch(name);
+    // TODO(rnystrom): A simpler implementation is to test against the regex
+    // "_?[A-Z].*?[a-z]". However, that currently has much worse performance on
+    // AOT: https://github.com/dart-lang/sdk/issues/37785.
+    const underscore = 95;
```
srawlins: I imagine you know about this package, but [charcode](https://pub.dev/packages/charcode) makes code like this readable.

munificent: Yeah, I didn't feel like it was worth adding a dependency for five integers. :)

lib/src/source_visitor.dart
```diff
@@ -76,14 +74,42 @@ class SourceVisitor extends ThrowingAstVisitor {
     return target is SimpleIdentifier && _looksLikeClassName(target.name);
   }
 
+  /// Whether [name] appears to be a type name.
+  ///
+  /// Type names begin with a capital letter and contain at least one lowercase
```
natebosch: [out of scope] Is it sensible to consider `_A` or `A` to be a class name?

lib/src/source_visitor.dart
```diff
@@ -76,14 +74,42 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+
+    // It can be private.
+    if (firstChar == underscore) firstChar = name.codeUnitAt(1);
+
+    // It must start with a capital letter.
+    if (firstChar < capitalA || firstChar > capitalZ) return false;
+
+    // And have at least one lowercase letter in it. Otherwise it could be a
+    // SCREAMING_CAPS constant.
+    for (var i = 0; i < name.length; i++) {
```
mkustermann: dbc: The `i == 0` and possibly `i == 1` cases were already handled above and could be avoided here.

lib/src/source_visitor.dart
```diff
@@ -76,14 +74,42 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+    const underscore = 95;
+    const capitalA = 65;
+    const capitalZ = 90;
+    const lowerA = 97;
+    const lowerZ = 122;
+
+    var firstChar = name.codeUnitAt(0);
+
+    // It can be private.
+    if (firstChar == underscore) firstChar = name.codeUnitAt(1);
```
mkustermann: dbc: If `name == "_"` then this would throw index out of bounds exception.

lib/src/source_visitor.dart
munificent: Honestly, I don't think it matters much one way or the other. The only way this could come into play is a (1) static method call that (2) didn't fit on one line on (3) an identifier whose name (4) is a single capital letter.

lib/src/source_visitor.dart
munificent: Done.

lib/src/source_visitor.dart
munificent: Nice catch, thank you!

munificent:
Hey @srawlins or @natebosch, would one of you mind giving this the old `:+1` if you think it's ready?

PR #844, opened 2019-09-20
Update to analyzer 0.38.3.

munificent:
Now that it correctly parses method chains with null assertion
operators, add some more tests around that.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -11,7 +11,7 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.38.0 <0.39.0'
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
 
 dependencies:
-  analyzer: '>=0.38.0 <0.39.0'
+  analyzer: '>=0.38.3 <0.39.0'
```
natebosch: How about `^0.38.3`?

munificent: Good idea. Though I've got three other PRs that make this same change, so I'll wait to merge those first to keep things easier on me.

PR #851, opened 2019-09-24
Don't force a split before "." on parenthesized targets.

### LINKED ISSUE #704: Latest release formats calls on multiline literals strangely
This code used to be formatted as so:

```dart
void main() {
  test("synchronized files are up-to-date", () {
    ({
      'lib/src/visitor/async_evaluate.dart': 'lib/src/visitor/evaluate.dart',
      'lib/src/async_environment.dart': 'lib/src/environment.dart'
    }).forEach((sourcePath, targetPath) {
      var source = new File(sourcePath).readAsStringSync();
      var target = new File(targetPath).readAsStringSync();

      var hash = sha1.convert(utf8.encode(source));
      if (!target.contains("Checksum: $hash")) {
        fail("$targetPath is out-of-date.\n"
            "Run pub run grinder to update it.");
      }
    });
  });
}
```

but as of the latest 1.1.0 release, it's formatted as so:

```dart
void main() {
  test("synchronized files are up-to-date", () {
    ({
      'lib/src/visitor/async_evaluate.dart': 'lib/src/visitor/evaluate.dart',
      'lib/src/async_environment.dart': 'lib/src/environment.dart'
    })
        .forEach((sourcePath, targetPath) {
      var source = new File(sourcePath).readAsStringSync();
      var target = new File(targetPath).readAsStringSync();

      var hash = sha1.convert(utf8.encode(source));
      if (!target.contains("Checksum: $hash")) {
        fail("$targetPath is out-of-date.\n"
            "Run pub run grinder to update it.");
      }
    });
  });
}
```

The `.forEach()` is now on its own line, which looks worse than the original formatting to me.

munificent:
The method chain splitting logic allows "." to immediately follow some
targets, like collection literals:

    [
      element,
    ].forEach((_) => ...);

This extends that to allow parenthesized expressions:

    ([
      element,
    ]).forEach((_) => ...);

Fix #704.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Format the `late` modifier.
 * Format the `required` modifier.
 * Better formatting of empty spread collections (#831).
+* Don't force split before `.` when the target is parenthesized (#704).
 
 # 1.2.10
 
--- a/lib/src/call_chain_visitor.dart
+++ b/lib/src/call_chain_visitor.dart
@@ -284,9 +284,13 @@ class CallChainVisitor {
   bool _forcesSplit(Expression expression) {
     // TODO(rnystrom): Other cases we may want to consider handling and
     // recursing into:
-    // * ParenthesizedExpression.
     // * The right operand in an infix operator call.
     // * The body of a `=>` function.
+
+    // Unwrap parentheses.
+    while (expression is ParenthesizedExpression) {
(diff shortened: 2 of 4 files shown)


munificent:
"Risky" means "this might change the formatting of other unrelated code". Any time I touch method call splitting, it's risky. But in this case the fix was very targeted.

PR #853, opened 2019-09-25
Simplify dep version constraints

natebosch:
- Switch `analyzer` dep to caret style since it's a shorter way to
  express the same constraint.
- Switch `args` dep to caret style since there are no earlier version
  which also support the SDK used by this package.


munificent:
I just saw this after publishing 1.3.0. :( So you'll also need to bump the version to 1.3.1-dev too and it will go out in the next release.

PR #854, opened 2019-09-26
Revert "Remove npm bits"

derekxu16:
Reverts dart-lang/dart_style#787
dart-style needs to be updated for js_facade_gen


natebosch:
Have we investigated the possibility of rewriting `js_facade_gen` in Dart? I'm not excited about having continued dependencies on node for this...

munificent:
They're only dev dependencies so they don't really impact normal users of dart_style. This is basically just a utility script that can run to push dartfmt to npm.

PR #858, opened 2019-09-26
Fix inconsistencies after revert

natebosch:
- Remove `async` dep since it is unused.
- Use consistent single quotes for imports.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -17,7 +17,6 @@ dependencies:
   source_span: ^1.4.0
 
 dev_dependencies:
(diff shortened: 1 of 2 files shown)


munificent:
Oh, you'll need to bump the version in the pubspec to 1.3.2-dev too.

natebosch:
> you'll need to bump the version in the pubspec to 1.3.2-dev too.

It already is...

https://github.com/dart-lang/dart_style/blob/79c55f8ea62ee83a4982b502cb0163b8c06ea274/pubspec.yaml#L3

PR #867, opened 2019-10-22
Expand analyzer dependency to allow 0.39.0.

munificent:
There isn't a version of test that supports it yet, so users still won't
get that version. But I tested it locally by overriding the dependency
and ran into no problems, so it should be fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.3.2+1
+
+* Support `package:analyzer` `0.39.0`.
+
 # 1.3.2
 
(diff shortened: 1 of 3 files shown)


kevmoo:
> There isn't a version of test that supports it yet, so users still won't
> get that version. But I tested it locally by overriding the dependency
> and ran into no problems, so it should be fine.

https://github.com/dart-lang/test/pull/1103 😄

munificent:
Covered by ea55d48ad46b4515fd70f1617bd9f52c6d40158d.

PR #887, opened 2020-01-16
Introduce the verbose flag into dart_style 

jwren:
Introduce the verbose flag into dart_style to make it consistent with other dart CLI that have this flag (`-v` / `--verbose`)


munificent:
This seems a little confusing because `--verbose` doesn't do anything except with `--help`. What's the context behind this change?

jwren:
Thanks @munificent, landing the change.

PR #890, opened 2020-02-08
Make file writing atomic

refi64:
This ensures that the original file stays intact, even if dartfmt is
interrupted somehow. Should work on Windows and POSIX systems, as Dart
itself already passes the proper flags on Windows such that renames will
cleanly replace their target.

### DIFF
--- a/lib/src/formatter_options.dart
+++ b/lib/src/formatter_options.dart
@@ -130,12 +130,20 @@ class _OverwriteReporter extends _PrintReporter {
   @override
   void afterFile(File file, String label, SourceCode output, {bool changed}) {
     if (changed) {
+      // Write to a temporary file first, then rename it on top of the
+      // original. This ensures that, if the program is interrupted during the
+      // write, the original file's contents stay intact.
+      File tmpFile = File(file.path + '.tmp');
       try {
-        file.writeAsStringSync(output.text);
+        tmpFile.writeAsStringSync(output.text);
+        tmpFile.renameSync(file.path);
         print('Formatted $label');
       } on FileSystemException catch (err) {
         stderr.writeln('Could not overwrite $label: '
             '${err.osError.message} (error code ${err.osError.errorCode})');
+        try {
+          tmpFile.deleteSync();
+        } on FileSystemException {}
       }
     } else {
       print('Unchanged $label');


natebosch:
I think there are likely still some gaps (I don't know if `writeAsStringSync()` would necessarily do an `fsync()`, for instance) but this should be safer than where we are today.

In my opinion dropping `.tmp` files immediately next to the originals isn't optimal if we can avoid it. It would be a bit better to put them somewhere else. Maybe in `.dart_tool`? Or maybe we should create a unique temp directory (in this one) with https://api.dart.dev/stable/2.7.1/dart-io/Directory/createTempSync.html
Unfortunately I don't think we have APIs for creating uniquely named temp files.

@munificent - is there are higher level place we can hook in to create a temp directory so we don't need to make one for each file we write?

natebosch:
cc @jakemac53 - this might cause problematic behavior with our file watching in `build_runner`.

jakemac53:
We already have to handle this in build_runner because of certain editors, the same logic should work here (basically we just collect changes until we go a certain amount of time without seeing any).

refi64:
Creating it in .dart_tool should be fine. However, if created in a
temporary directory, there is a lot more opportunity for remnants to be
left behind (e.g. if an interruption occurs after the rename but before the
now empty directory is removed).

For some context, I basically have had bad luck with cases in the past
where applications would get stuck writing files, and I didn't want to end
up with a blank file because dartfmt got interrupted at the perfect time.

As for flushing, does seem like passing `flush: true` will sync data to the
disk. Sadly, macOS does seem to mess with that a bit and requires a
separate fcntl to do a "full flush"[1]. That being said, that would be
creeping into Dart SDK territory; at minimum, nothing can be done about
that on this end.

[1]: https://www.slideshare.net/mobile/nan1nan1/eat-my-data

natebosch:
Another option to consider is reusing a single file like `.dart_tool/dartfmt/swapfile`. As long as we can guarantee that we aren't trying to write more than one file at once this is safe and saves us from having to recreate directory structures under `.dart_tool/dartfmt/` to avoid name conflicts.

@munificent - do you have any thoughts on the structure of this code? Is it safe to assume we only write 1 file at a time? Is there some place we could hook in to initialize a directory like `.dart_tool/dartfmt/<temp dir name>`?

munificent:
Sorry for the long delay. I haven't had much time for dartfmt and got way behind on GitHub notifications.

I'm generally hesitant to add complexity that is (a) hard to test and (b) not clearly necessary. That kind of code tends to be brittle and bit-rot. How sure are we that this carries its weight? In many years of heavy dartfmt use by thousands of people, I can't recall ever hearing from a user that it left a file in a half-written state.

This change seems to mean that now dartfmt will overwrite a file whose write permission is disabled. At least in local testing on my Mac, the OS is happy to have rename clobber a non-writable file. That seems undesirable. The current behavior is that it will *not* overwrite a non-writable file. I think that's what users would expect and want.

There is also the question of where to put the temp file. If we don't put it next to the original file, we *will* run into nasty permission problems. We ran into this with pub where it placed some temp file (I can't remember what for) in its cache directory before moving it elsewhere. Issues we ran into:

1. Some users had this cache directory read-only. So even though they could have written to the ultimate destination, the operation failed since the intermediate location couldn't be written to.

2. You can't rename across volumes. (Obvious in retrospect.)

Putting the temp files next to the destination file is generally better but then runs the risk of leaving clutter if a failure occurs. Of course, times when failure occurs are exactly when this feature is supposed to be useful, so that seems suboptimal to me.

refi64:
Oof, this got lost in my email.

> This change seems to mean that now dartfmt will overwrite a file whose write permission is disabled. At least in local testing on my Mac, the OS is happy to have rename clobber a non-writable file. That seems undesirable. The current behavior is that it will not overwrite a non-writable file. I think that's what users would expect and want.

Whoops :sweat_smile: This can be worked around though, you copy the original file to the temporary one and then clear out the temporary copy. That way, all attributes that can be copied (which should be everything but file birth times afaik) will be preserved.

FWIW the dilemma you pointed out with using a separate cache directory is something I'm familiar with and is why I had it in the same directory.

As for whether or not it's worthwhile...IME it just isn't until it is. A while back I used to use a hard drive connected over USB, and I would constantly run into issues where writes would sometimes take a bit longer due to limited I/O bandwith. As a result, I've encountered half-written files (or, more often, files that were entirely emptied) enough times that I'm no longer really comfortable using tools that do non-atomic writes.

It's correct that the file may stay around in the directory if it gets interrupted. However, if you run into a scenario where the temporary file is left behind, that may have been originally a scenario where the original file would have been cleared out. And, if the worst-case scenario is a temporary file left behind vs losing the entire file contents, I think the temporary file would probably be the safer choice...

For the record, I also primarily use Linux, where the primary filesystem ext4...doesn't always have the best reliability track record. This may be far more rare on e.g. Macs if the default filesystem there is more stable overall, but I don't know enough about all that to really make a judgement.

PR #904, opened 2020-03-19
For split control flow elements that contain other ones.

### LINKED ISSUE #869: Should we force nested for/if elements to separate lines in collection literals
Things can feel a little dense when there are multiple conditions or loops on the same line. It might work out for readers accustomed to generators in Python, but coming from chains of iterable calls or non-element for loop it can be hard to read.

Should we break these up? If we do break them up, should the inner one be forced to separate lines, or can it be all one line?

Today:

```dart
    var failed = [
      for (var e in _restResultEvents) if (!e.isSuccess) e.message,
    ];
```

Option 1:

```dart
    var failed = [
      for (var e in _restResultEvents)
        if (!e.isSuccess)
          e.message,
    ];
```

Option 2:

```dart
    var failed = [
      for (var e in _restResultEvents)
        if (!e.isSuccess) e.message,
    ];
```

munificent:
Fix #869.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -117,6 +117,9 @@ class SourceVisitor extends ThrowingAstVisitor {
     return false;
   }
 
+  static bool _isControlFlowElement(AstNode node) =>
+      node is IfElement || node is ForElement;
+
   /// The builder for the block that is currently being visited.
   ChunkBuilder builder;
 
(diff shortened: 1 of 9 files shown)


lib/src/source_visitor.dart
```diff
@@ -3023,8 +3043,8 @@ class SourceVisitor extends ThrowingAstVisitor {
   /// and followed by the given functions.
   void visitNodes(Iterable<AstNode> nodes,
       {void Function() before,
-      void Function() between,
-      void Function() after}) {
+        void Function() between,
```
natebosch: When did this change?

munificent: Not sure, may just be from my copy/pasting some code and forgetting to format it. Reformatted the whole file.

PR #910, opened 2020-03-23
Remove "--verbose" as a general flag.

munificent:
Dartfmt doesn't have verbose *output*, it just has verbose *help*, so
only allow "-v" as an option to "--help".


munificent:
Never mind, let's not go this direction.

PR #911, opened 2020-03-27
Add a hidden Command library for integrating into dartdev

### LINKED ISSUE #889: Expose library to create dart_style ArgParser
This library will be "hidden" in that only dartdev is expected and supported to use it. But it will let dartdev create a subcommand for formatting whose options and arguments match dart_style.

However, we do want to tweak some of the behavior. Right now, dartfmt has two command line behaviors that aren't great defaults:

* If you don't pass a path, it assumes you want to read from stdin and just sits there waiting for you. This trips up users who don't know what's going on (#776). A better default is to just print an error and require an explicit option (maybe just `--`) to read from stdin.

* You have to pass `-w` to overwrite files. By default it dumps to stdout. Instead, it should default to overwriting and take an option (maybe `--print`?) to print the result to stdout.

We don't want dartfmt itself to change its command line API since that's a breaking change. (And, in particular, changing the existing behavior to starting overwriting files is a *very* breaking change.) So this library should let you configure whether you get the new or old option set.

This is probably a good opportunity to revisit some other things about the command-line API such as:

* A more coherent way to specify output format. `--machine` and `--dry-run` are sort of a hodgepodge.

* A way (by default?) to not print the list of formatted files while its running.

munificent:
I know this is a huge amount of code changes, but it was the best I could do not break the old CLI while also not compromising the new UI. I've broken it into a series of commits to hopefully make it easier to review, and added copious tests to make sure everything works as expected.

Fix #889.

### DIFF
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -16,7 +16,7 @@ import 'package:dart_style/src/style_fix.dart';
 void main(List<String> args) {
   var parser = ArgParser(allowTrailingOptions: true);
 
-  defineOptions(parser);
+  defineOptions(parser, oldCli: true);
 
   ArgResults argResults;
   try {
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -2,87 +2,18 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:convert';
 import 'dart:io';
 
 import 'package:args/args.dart';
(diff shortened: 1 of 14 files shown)


munificent:
OK, this is landed now. I can publish a new version of dart_style, but if possible I'd like to verify that the new Command at least roughly works from within dartdev.

Can you locally test dartdev against a Git or path dependency on dart_style and see if it's doing what it's supposed to do?

PR #912, opened 2020-04-03
Update to use the new null-aware subscript syntax.

munificent:


pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
```
kevmoo, line 10: also need to bump the min SDK support?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: dart
 
 dart:
-  - 2.3.0
```
kevmoo, line 4: It's good to keep this in sync w/ the lowerbound of the pubspec.yaml sdk constraint

pubspec.yaml
munificent: OK, I'll try 2.7.0 and see if that satisfies Travis and the version solver. (The reason I bumped this is that there's no analyzer package that works on 2.3.0 and meets dart_style's new higher constraint.)

.travis.yml
munificent: Good to know. Done.

kevmoo:
It keeps us honest to make sure we test the lower bound of the SDK that we
claim to support!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Always place a blank line after script tags (#782).
 * Don't add unneeded splits on if elements near comments (#888).
 * Indent blocks in initializers of multiple-variable declarations.
+* Update the null-aware subscript syntax from `?.[]` to `?[]`.
```
kevmoo, line 12: tis nice to tell folks you bumped the min SDK here, too! 😊

munificent: This doesn't seem CHANGELOG worthy to me.

PR #921, opened 2020-04-21
Generate native executables to use when running the integration tests.

munificent:
There are two test suites that spawn Dart executables to validate the
CLI behavior. Those are painfully slow when running the programs from
source. This precompiles them to native executes and then spawns those.

On my machine it takes the tests from ~4 minutes down to ~40 seconds.

(Also this should fix the failure when running the tests inside the SDK.)

### DIFF
--- a/test/utils.dart
+++ b/test/utils.dart
@@ -53,8 +53,7 @@ void compileFormatterExecutable() {
 /// subsequent tests and to clean up the executable.
 void compileCommandExecutable() {
   setUpAll(() async {
-    _commandExecutablePath =
-        await _compileSnapshot('tool/command_shell.dart');
+    _commandExecutablePath = await _compileSnapshot('tool/command_shell.dart');
   });
 
   tearDownAll(() async {
--- a/test/utils.dart
+++ b/test/utils.dart
@@ -33,42 +33,45 @@ String _commandExecutablePath;
 String _formatterExecutablePath;
 
 /// Compiles format.dart to a native executable for tests to use.
+///
+/// Calls [setupAll()] and [tearDownAll()] to coordinate this when the
+/// subsequent tests and to clean up the executable.
 void compileFormatterExecutable() {
   setUpAll(() async {
-    _formatterExecutablePath = await _compileExecutable('bin/format.dart');
+    _formatterExecutablePath = await _compileSnapshot('bin/format.dart');
   });
 
   tearDownAll(() async {
-    await _deleteExecutable(_formatterExecutablePath);
+    await _deleteSnapshot(_formatterExecutablePath);
     _formatterExecutablePath = null;
   });
 }
 
 /// Compiles command_shell.dart to a native executable for tests to use.
+///
+/// Calls [setupAll()] and [tearDownAll()] to coordinate this when the
+/// subsequent tests and to clean up the executable.
 void compileCommandExecutable() {
   setUpAll(() async {
     _commandExecutablePath =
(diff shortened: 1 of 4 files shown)


test/utils.dart
```diff
@@ -25,17 +25,89 @@ const formattedOutput = 'void main() => print("hello");';
...
-/// Runs the command line formatter, passing it [args].
-Future<TestProcess> runFormatter([List<String> args]) {
+/// If tool/command_shell.dart has been compiled to a snapshot, this is the path
+/// to it.
+String _commandExecutablePath;
+
+/// If bin/format.dart has been compiled to a snapshot, this is the path to it.
+String _formatterExecutablePath;
+
+/// Compiles format.dart to a native executable for tests to use.
```
natebosch, line 35: I would mention the use of `setUpAll` and `tearDownAll` here. The name makes me think it would happen within the call.

test/utils.dart
```diff
@@ -25,17 +25,89 @@ const formattedOutput = 'void main() => print("hello");';
...
+  var scriptName = p.basename(script);
+  var tempDir =
+      await Directory.systemTemp.createTemp(p.withoutExtension(scriptName));
+  var executable = p.join(tempDir.path, '$scriptName.aot');
+
+  var dart2Native =
+      p.join(p.dirname(Platform.resolvedExecutable), 'dart2native');
+
   // Locate the "test" directory. Use mirrors so that this works with the test
   // package, which loads this suite into an isolate.
```
natebosch, line 74: [out of scope] I don't think you should need mirrors for this. `Isolate.resolvePackageUri` would let you find `lib/` and I think it's fair to assume relative path of `test/` from `lib/`, especially since we already assume a relative path of `bin/` from `test/`...

test/utils.dart
```diff
@@ -25,17 +25,89 @@ const formattedOutput = 'void main() => print("hello");';
...
+///
+/// We do this instead of spawning the script from source each time because it's
+/// much faster when the same script needs to be run several times.
+Future<String> _compileExecutable(String script) async {
+  var scriptName = p.basename(script);
+  var tempDir =
+      await Directory.systemTemp.createTemp(p.withoutExtension(scriptName));
+  var executable = p.join(tempDir.path, '$scriptName.aot');
+
+  var dart2Native =
```
natebosch, line 70: Have you compared this against `dart --snapshot` for overall runtime?

munificent: I didn't. I thought about it, but the docs for the old kernel snapshots sound like it wouldn't offer much speed improvement, and the app-JIT snapshots require some kind of training data, which seemed like more trouble than it was worth.

But now you have piqued my curiosity. I tried all three options:

AOT:

```
00:42 +1801: All tests passed!
real  0m43.214s
user  1m17.798s
sys 0m7.271s
```

Kernel snapshot:

```
00:39 +1801: All tests passed!
real  0m40.140s
user  1m15.609s
sys 0m9.052s
```

App JIT snapshot (trained using just `--help`):

```
00:36 +1801: All tests passed!
real  0m36.581s
user  1m12.847s
sys 0m8.341s
```

So changed it to use the latter instead.

test/utils.dart
munificent: I want to get this landed so I can roll into the SDK soon, so added a TODO to look into this later.

test/utils.dart
munificent: Done.

munificent:
Thanks for the review!

PR #941, opened 2020-08-10
Add "--verbose" to "--help" and hide most options normally.

### LINKED ISSUE #938: support a more verbose option for printing help text
The current help text for `dart format` is pretty long (appended to this issue for reference). It would be nice if it was a bit more terse - perhaps not much longer that a typical user's terminal window height.

The combined dart CLI tool is using the `FormatCommand` from this package to define the flags for `dart format`. It would be great if that was parametrizable, so we could both pass in whether we were running in a `-h -v` context (`verboseHelp`), and whether some options should be documented or supported at all.

For instance, `-output` options might only be shown if the users runs `dart format -h -v`; the `--version` option (for `dart format`) might not even be supported - the user can already run `dart --version`.

```
Idiomatically formats Dart source code.

Usage: dart format [options...] <files or directories...>
-h, --help                             Print this usage information.

Common options:
-o, --output                           Where formatted output should be written.

          [json]                       Print code and selection as JSON
          [none]                       Discard.
          [show]                       Print code to terminal.
          [write] (default)            Overwrite formatted files on disc.

    --show                             Which filenames to print.

          [all]                        All visited files and directories.
          [changed] (default)          Only the names of files whose formatting is changed.
          [none]                       No file names or directories.

    --summary                          Summary shown after formatting completes.

          [line] (default)             Single line summary.
          [none]                       No summary.
          [profile]                    Tracks how long it took for format each file.

Non-whitespace fixes (off by default):
    --fix                              Apply all style fixes.
    --fix-doc-comments                 Use triple slash for documentation comments.
    --fix-function-typedefs            Use new syntax for function type typedefs.
    --fix-named-default-separator      Use "=" as the separator before named parameter default values.
    --fix-optional-const               Remove "const" keyword inside constant context.
    --fix-optional-new                 Remove "new" keyword.
    --fix-single-cascade-statements    Remove unnecessary single cascades from expression statements.

Other options:
-l, --line-length                      Wrap lines longer than this.
                                       (defaults to "80")
-i, --indent                           Spaces of leading indentation.
                                       (defaults to "0")
    --set-exit-if-changed              Return exit code 1 if there are any formatting changes.
    --follow-links                     Follow links to files and directories.
                                       If unset, links will be ignored.
    --version                          Show version information.

Options when formatting from stdin:
    --selection                        Selection to preserve formatted as "start:length".
    --stdin-name                       The path name to show when an error occurs.
                                       (defaults to "stdin")

Run "dart help" to see global options.
```

munificent:
Makes the "dart format help" and error messages much easier on the eyes.

Fix #938.

cc @devoncarew

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.3.7
+
+* Split help into verbose and non-verbose lists (#938).
+
 # 1.3.6
 
 * Change the path used in error messages when reading from stdin from "<stdin>"
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -16,7 +16,8 @@ import 'package:dart_style/src/style_fix.dart';
 void main(List<String> args) {
   var parser = ArgParser(allowTrailingOptions: true);
 
-  defineOptions(parser, oldCli: true);
+  defineOptions(parser,
+      oldCli: true, verbose: args.contains('--verbose') || args.contains('-v'));
 
   ArgResults argResults;
   try {
@@ -33,6 +34,10 @@ void main(List<String> args) {
   if (argResults['version']) {
     print(dartStyleVersion);
     return;
+  }
+
+  if (argResults['verbose'] && !argResults['help']) {
+    usageError(parser, 'Can only use --verbose with --help.');
   }
 
   List<int> selection;
--- a/lib/src/cli/format_command.dart
+++ b/lib/src/cli/format_command.dart
@@ -24,8 +24,8 @@ class FormatCommand extends Command {
   String get invocation =>
       '${runner.executableName} $name [options...] <files or directories...>';
 
-  FormatCommand() {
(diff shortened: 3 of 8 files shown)


devoncarew:
Thank!

Note that we'll need to bring this version (`1.3.7`) into the dart sdk (currently on `1.3.6`) in order to consume this new API from dartdev. cc @jwren

munificent:
> Note that we'll need to bring this version (1.3.7) into the dart sdk (currently on 1.3.6) in order to consume this new API from dartdev.

Yup! I'm going to see if I can get a couple of other bug fixes in first. If it gets close to the wire and you want me to publish a version, feel free to remind me and I'll kick one out.

devoncarew:
Gotcha, thanks. Note that we're branching on or around the 2nd week of Sept for the next stable dart release (https://github.com/dart-lang/sdk/milestone/83).

PR #947, opened 2020-08-19
Format `external` and `abstract` on variables.

### LINKED ISSUE #946: Dartfmt: Implement external and abstract variables
Cf. the [implementation plan](https://github.com/dart-lang/language/blob/master/accepted/future-releases/abstract-external-fields/implementation-plan.md), the addition of external and abstract variables to Dart may require some updates.

munificent:
Fix #946.

The dependency_override is temporary. Once a version of test that I can use is published, I'll fix the pubspec here before I merge into master. I just wanted to get this out for review since the dart_style changes are done and working.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
 
 dependencies:
-  analyzer: ">=0.39.5 <0.40.0"
+  analyzer: ^0.40.0
   args: ^1.0.0
   path: ^1.0.0
   source_span: ^1.4.0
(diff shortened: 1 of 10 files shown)


munificent:
@natebosch, can you take a look at my pubspec and Travis config changes and see if those look reasonable?

pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.7.0 <3.0.0'
```
natebosch, line 10: Any particular reason not to bump the min SDK? Is it important that `dart_style` gets version updates on older SDKs?

munificent: Not really. Went ahead and bumped it (somewhat arbitrarily) to 2.9.0.

PR #968, opened 2020-10-29
Make FormatCommand implement Command<int>.

### LINKED ISSUE #964: Consider making the exported Command a Command<int>
https://github.com/dart-lang/dart_style/blob/master/lib/src/cli/format_command.dart#L16 is a `Command<dynamic`, compatible with current dartdev/dart cli.

I suggest we make dart cli take `Command<int>`, and likewise make the command exported hare a `Command<int>` with the value being the exit code.
The `run` function already returns `Future<int>`, so the change would be trivial.

Pub needs to do the same.

See: https://dart-review.googlesource.com/c/sdk/+/168940 for a bit of context.

munificent:
Fix #964.

### DIFF
--- a/lib/src/cli/format_command.dart
+++ b/lib/src/cli/format_command.dart
@@ -13,7 +13,7 @@ import 'output.dart';
 import 'show.dart';
 import 'summary.dart';
 
-class FormatCommand extends Command {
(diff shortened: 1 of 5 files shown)


sigurdm (approved):
LGTM thanks!

PR #969, opened 2020-10-29
Don't duplicate comments on chained if elements.

### LINKED ISSUE #966: Can't format with comment above in-lined conditional in List
In a list literal created with a for-loop (with conditionals) inside, having a comment placed directly above a conditional caused the formatter to fail

```
return [
      for (final x in exampleList)
        if (conditionA)
          ItemConstructorA(...)
       // Sample multi-line comment
      // which broke the formatter
        else if (conditionB)
           ItemConstructorB(...)
     else
       ItemConstructorC(...)
];
```

Error message was `Exception from the Dart analysis server: The formatter produced unexpected output. Input was: [...]`

Moving the comment to below the conditional and above the statement like so:

```
return [
      for (final x in exampleList)
        if (conditionA)
          ItemConstructorA(...)
        else if (conditionB)
       // Sample multi-line comment
      // which no longer breaks formatter
           ItemConstructorB(...)
     else
       ItemConstructorC(...)
];
```

caused the formatter to be able to format the file once again.

munificent:
There was a bug where every "else" in a chained if element used the
first element's "else" token instead of its own. This had the
unfortunate side effect of duplicating any comment that appeared before
the first one (and discarding comments on the later ones).

Fix #966.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dart_style
 # Note: See tool/grind.dart for how to bump the version.
-version: 1.3.8+1
+version: 1.3.9-dev
 description: >-
   Opinionated, automatic Dart source code formatter.
   Provides an API and a CLI tool.
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dart_style
 # Note: See tool/grind.dart for how to bump the version.
-version: 1.3.9-dev
+version: 1.3.9
 description: >-
   Opinionated, automatic Dart source code formatter.
   Provides an API and a CLI tool.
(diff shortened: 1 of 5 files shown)


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dart_style
 # Note: See tool/grind.dart for how to bump the version.
-version: 1.3.8+1
```
natebosch, line 3: FYI for future changes this should have been `1.3.9` before. We avoid the `+1` once the major version is non-zero to avoid confusion.

natebosch (approved):

pubspec.yaml
munificent: Oops, right.

PR #970, opened 2020-11-02
Remove unused dart:async imports.

srawlins:
As of Dart 2.1, Future/Stream have been exported from dart:core.


munificent:
Would you mind bumping the pubspec version to 1.3.10-dev? Then LGTM.

munificent:
Thanks!

PR #981, opened 2020-12-21
Migrate to GitHub Actions

athomas:


athomas:
The lints are already failing on travis, but fatal-infos isn't passed there.

munificent (approved):

PR #994, opened 2021-02-17
Preliminary support for generic annotations.

stereotype441:
This change provides enough support to avoid a crash and ensure that
generic annotations are preserved by the formatter.

We don't yet insert as many line breaks as we arguably should (see the
example in test/splitting/annotations.unit); we can improve that in a
future change.

### DIFF
--- a/test/splitting/annotations.unit
+++ b/test/splitting/annotations.unit
@@ -1,0 +1,9 @@
+40 columns                              |
+>>> blet
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument) int i = 0;
+<<<
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
+    veryLongArgument,
+    otherVeryLongArgument,
+    thirdVeryLongArgument)
+int i = 0;
--- a/test/splitting/annotations.unit
+++ b/test/splitting/annotations.unit
@@ -1,9 +1,95 @@
 40 columns                              |
->>> blet
-@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument) int i = 0;
+>>> complex annotation that all fits on one line
+@a.B<C,D>(e,f,g) int i = 0;
 <<<
-@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
+@a.B<C, D>(e, f, g)
+int i = 0;
+>>> prefer to split between args even when they all fit on next line
+@BigLongClassName<First, Second, Third>() int i = 0;
+<<<
+@BigLongClassName<First, Second,
+    Third>()
+int i = 0;
+>>> split before first if needed
+@BigLongClassName<FirstTypeArgumentIsLong, Second>() int i = 0;
+<<<
+@BigLongClassName<
+    FirstTypeArgumentIsLong, Second>()
+int i = 0;
+>>> split in middle if fit in two lines
+@BigLongClassName<First, Second, Third, Fourth, Fifth, Sixth, Seventh>() int i = 0;
+<<<
+@BigLongClassName<First, Second, Third,
+    Fourth, Fifth, Sixth, Seventh>()
+int i = 0;
+>>> split one per line if they don't fit in two lines
+@BigLongClassName<First, Second, Third, Fourth, Fifth, Sixth, Seventh, Eighth>() int i = 0;
+<<<
+@BigLongClassName<
+    First,
+    Second,
+    Third,
+    Fourth,
+    Fifth,
+    Sixth,
+    Seventh,
+    Eighth>()
+int i = 0;
+>>> prefers to not split at type arguments
+@SomeBigClass<
+    TypeArgument>(valueArgument)
+int i = 0;
+<<<
+@SomeBigClass<TypeArgument>(
+    valueArgument)
+int i = 0;
+>>> split both type and value arguments
+@SomeBigClass<First, Second, Third, Fourth, Fifth, Sixth, Seventh, Eighth>(first, second, third, fourth, fifth, sixth, seventh, eighth) int i = 0;
+<<<
+@SomeBigClass<
+        First,
+        Second,
+        Third,
+        Fourth,
+        Fifth,
+        Sixth,
+        Seventh,
+        Eighth>(
+    first,
+    second,
+    third,
+    fourth,
+    fifth,
+    sixth,
+    seventh,
+    eighth)
+int i = 0;
+>>> prefer to split at arguments rather than prefix
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument,OtherVeryLongTypeArgument>(veryLongArgument,otherVeryLongArgument,thirdVeryLongArgument) int i = 0;
+<<<
+@veryLongPrefix.VeryLongClassName<
+        VeryLongTypeArgument,
+        OtherVeryLongTypeArgument>(
     veryLongArgument,
     otherVeryLongArgument,
     thirdVeryLongArgument)
+int i = 0;
+>>> nested type arguments, no splitting required
+@A<B<C,D>,E<F,G>>() int i = 0;
+<<<
+@A<B<C, D>, E<F, G>>()
+int i = 0;
+>>> nested type arguments, split at outer level when possible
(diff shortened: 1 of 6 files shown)


test/splitting/annotations.unit
```diff
@@ -0,0 +1,9 @@
+40 columns                              |
+>>> blet
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument) int i = 0;
+<<<
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
```
stereotype441, line 5: @munificent It seems like this behavior could definitely be improved upon (it's way over 40 columns, and AFACT unnecessarily so).  But I'm not sure how to proceed.  I spent a while trying to modify `visitAnnotation` to make this format better (I was adding calls to `builder.startSpan/endSpan` and `builder.nestExpression/unnest`, trying to mimic what is done in `visitInstanceCreationExpression`).  But I didn't have any luck.  If you want to give me guidance on what else to try I'd be happy to pursue this further.

test/splitting/annotations.unit
```diff
@@ -0,0 +1,9 @@
+40 columns                              |
+>>> blet
```
munificent, line 2: "blet"? :)

test/splitting/annotations.unit
munificent: Yeah, this is definitely not ideal. I'll give you some pointers for how to fix this and how to debug issues in general in case you end up doing more hacking on dartfmt. When I want to test its behavior, I use example/format.dart. That formats a given snippet of code and also enables a bunch of diagnostic logging. You can also run that script in a debugger if you want to step through the formatter.

In this case, I changed this line:

```dart
  formatStmt('a is int????;');
```

to:

```dart
  formatUnit('''
@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument) int i = 0;
  ''', 40);
```

Now when I run `dart example/format.dart`, it prints:

```
Built:
0:  @veryLongPrefix              ╖      $1  5             nest 0:4
1:  .VeryLongClassName<          ╜╖     $4  TypeArg8      nest 0:4
2:  VeryLongTypeArgument,         ║     $4  TypeArg8      nest 0:4
3:  OtherVeryLongTypeArgument>(   ╜╖    $1  Pos12         nest 0:4
4:  veryLongArgument,              ║2   $1  Pos12         nest 0:4
5:  otherVeryLongArgument,         ║║   $1  Pos12         nest 0:4
6:  thirdVeryLongArgument)         ╜╜       16!                       divide
7:  int                              ╖  $1  18            nest 0:4
8:  i =                              ╜  $1  21            nest 0:4
9:  0;                                      22!
```

This section shows you all of the chunks it creates (roughly its notion of tokens) with the splitting rules attached to the end of each chunk, the spans that group them together (the little bars), and the expression nesting stack in play for each chunk. It's a little (lot) cryptic, but there's a lot of information here.

Then:

```
Writing:
0:  @veryLongPrefix              ╖     $1  5             nest 0:4
1:  .VeryLongClassName<          ╜╖    $4  TypeArg8      nest 0:4
2:  VeryLongTypeArgument,         ║    $4  TypeArg8      nest 0:4
3:  OtherVeryLongTypeArgument>(   ╜╖   $1  Pos12         nest 0:4
4:  veryLongArgument,              ║2  $1  Pos12         nest 0:4
5:  otherVeryLongArgument,         ║║  $1  Pos12         nest 0:4
6:  thirdVeryLongArgument)         ╜╜      16!                       divide
```

This is basically a subset of the above. It prints one of these for each independently splittable region. In this case, it knows the annotation can be handled separately from the variable declaration.

Then it prints the line splitters attempts:

```
? ? ? 1   $0 (106 over) (best)
@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument)

1 0 1 1   $4 (10059 over)
@veryLongPrefix
| | .VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
| | veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument)

snip many here...

1 2 2 1   $11 (20053 over)
@veryLongPrefix
| | .VeryLongClassName<
| | VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument,
| | thirdVeryLongArgument)

0 0 4 1   $4 (43 over) (winner)
@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
| | veryLongArgument,
| | otherVeryLongArgument,
| | thirdVeryLongArgument)
```

This shows each attempted solution it came up with and the score it assigned. The leading numbers show which chunks were split, the `$` is the score, and the `(... over)` is how many characters didn't fit within the line length.

Some of those overflow numbers are really large. That's [this hack](https://github.com/dart-lang/dart_style/blob/master/lib/src/line_splitting/solve_state.dart#L382-L404). That gross hack should be rarely hit. Ideally, we'd craft the right set of formatting rules and constraints to completely eliminate it, but I was never fully able to. Argument list formatting is pretty complex.

In this case, you're hitting it because the new type argument list interacts poorly with the following normal argument list. It keeps trying to format things like:

```
@Foo<Bar,
    Baz(argument,
    another)
```

Then it ends up with two lines with the same indentation for different reasons and it punishes itself with a really high score.

Fortunately, normal method calls have a similar problem, so we can just borrow the solution from there. In `visitMethodInvocation()`, we have:

```dart
      builder.nestExpression();
      visit(node.typeArguments);
      visitArgumentList(node.argumentList, nestExpression: false);
      builder.unnest();
```

That outer `nestExpression()` and the `nestExpression: false` bits wrap both the type arguments and value arguments in a single shared level of expression nesting. This way, the first +4 indentation applies to both and then the type argument list will indent another +4 if it splits. Crucially, this avoids the hack where the formatter thinks the two lines have the same indentation for unrelated reasons.

Applied to annotations, it looks like:

```dart
  @override
  void visitAnnotation(Annotation node) {
    token(node.atSign);
    visit(node.name);

    builder.nestExpression();
    visit(node.typeArguments);
    token(node.period);
    visit(node.constructorName);

    // Metadata annotations are always const contexts.
    _constNesting++;

    visitArgumentList(node.arguments, nestExpression: false);

    _constNesting--;

    builder.unnest();
  }
```

Try that and you'll get the much nicer:

```dart
@veryLongPrefix.VeryLongClassName<
        VeryLongTypeArgument,
        OtherVeryLongTypeArgument>(
    veryLongArgument,
    otherVeryLongArgument,
    thirdVeryLongArgument)
int i = 0;
```

test/splitting/annotations.unit
```diff
@@ -0,0 +1,9 @@
...
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(veryLongArgument, otherVeryLongArgument, thirdVeryLongArgument) int i = 0;
+<<<
+@veryLongPrefix.VeryLongClassName<VeryLongTypeArgument, OtherVeryLongTypeArgument>(
+    veryLongArgument,
+    otherVeryLongArgument,
+    thirdVeryLongArgument)
+int i = 0;
```
munificent, line 9: It would be good to have some examples of nested generics too, and how splitting and indentation is handled in them. You can copy the snippets from splitting/type_arguments.stmt and tweak them to be annotations.

munificent (commented):
Thanks for doing this! I dug into the problem you ran into and gave you a whole walk through. If you end up hacking on the formatter more (which would be great), that should be useful. A couple of other things most PRs should do:

* Add an entry to the CHANGELOG for the change.
* If the pubspec is not already at a `-dev` version, bump the number and add `-dev` to it. That way we know there are unreleased changes on master.

test/splitting/annotations.unit
stereotype441: I think for now I'll just take your suggested change verbatim, but thanks so much for the detailed explanation nonetheless.  It will definitely be useful next time I hack on the formatter :)

test/splitting/annotations.unit
stereotype441: Whoops.  An old college friend of mine used to use that word in place of "foo" all the time and I kind of borrowed it from him.  In this case it means "Hey Paul, don't forget to replace this with a meaningful comment"... and clearly I did forget.

test/splitting/annotations.unit
stereotype441: Will do.

munificent (approved):
Looks great, thanks!

PR #996, opened 2021-02-25
Migrate to null safety!

munificent:
There should be no real changes to the public API but bumping to 2.0.0
just in case.

### DIFF
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -172,7 +172,10 @@ class ArgumentListVisitor {
       this._allArguments,
       this._arguments,
       this._functions,
-      this._argumentsAfterFunctions);
+      this._argumentsAfterFunctions) {
+    assert(_functions == null || _argumentsAfterFunctions != null,
+        'If _functions is passed, _argumentsAfterFunctions must be too.');
+  }
 
   /// Builds chunks for the argument list.
   void visit() {
@@ -187,20 +190,21 @@ class ArgumentListVisitor {
 
     _visitor.builder.endSpan();
 
-    if (_functions != null) {
+    var functions = _functions;
+    if (functions != null) {
       // TODO(rnystrom): It might look better to treat the parameter list of the
       // first function as if it were an argument in the preceding argument list
       // instead of just having this little solo split here. That would try to
       // keep the parameter list with other arguments when possible, and, I
       // think, generally look nicer.
-      if (_functions!.first == _allArguments.first) {
+      if (functions.first == _allArguments.first) {
         _visitor.soloZeroSplit();
       } else {
         _visitor.soloSplit();
       }
 
-      for (var argument in _functions!) {
-        if (argument != _functions!.first) _visitor.space();
+      for (var argument in functions) {
+        if (argument != functions.first) _visitor.space();
 
         _visitor.visit(argument);
 
@@ -437,11 +441,12 @@ class ArgumentSublist {
   void _visitArgument(
       SourceVisitor visitor, ArgumentRule rule, Expression argument) {
     // If we're about to write a block argument, handle it specially.
-    if (_blocks.containsKey(argument)) {
+    var argumentBlock = _blocks[argument];
+    if (argumentBlock != null) {
       rule.disableSplitOnInnerRules();
 
       // Tell it to use the rule we've already created.
-      visitor.beforeBlock(_blocks[argument]!, blockRule!, previousSplit);
+      visitor.beforeBlock(argumentBlock, blockRule!, previousSplit);
     } else if (_allArguments.length > 1) {
       // Edge case: Only bump the nesting if there are multiple arguments. This
       // lets us avoid spurious indentation in cases like:
@@ -465,7 +470,7 @@ class ArgumentSublist {
       visitor.visit(argument);
     }
 
-    if (_blocks.containsKey(argument)) {
+    if (argumentBlock != null) {
       rule.enableSplitOnInnerRules();
     } else if (_allArguments.length > 1) {
       visitor.builder.endBlockArgumentNesting();
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -1,9 +1,6 @@
 // Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
-
-library dart_style.src.argument_list_visitor;
-
 import 'dart:math' as math;
 
 import 'package:analyzer/dart/ast/ast.dart';
@@ -36,12 +33,12 @@ class ArgumentListVisitor {
   /// The contiguous list of block function arguments, if any.
   ///
   /// Otherwise, this is `null`.
-  final List<Expression> _functions;
+  final List<Expression>? _functions;
 
   /// If there are block function arguments, this is the arguments after them.
   ///
   /// Otherwise, this is `null`.
-  final ArgumentSublist _argumentsAfterFunctions;
+  final ArgumentSublist? _argumentsAfterFunctions;
 
   /// Returns `true` if there is only a single positional argument.
   bool get _isSingle =>
@@ -129,16 +126,17 @@ class ArgumentListVisitor {
       }
 
       for (var i = 0; i < functionsStart; i++) {
-        if (arguments[i] is! NamedExpression) continue;
+        var argument = arguments[i];
+        if (argument is! NamedExpression) continue;
 
-        if (isArrow(arguments[i])) {
+        if (isArrow(argument)) {
           functionsStart = null;
           break;
         }
       }
 
       for (var i = functionsEnd; i < arguments.length; i++) {
-        if (isArrow(arguments[i])) {
+        if (isArrow(arguments[i] as NamedExpression)) {
           functionsStart = null;
           break;
         }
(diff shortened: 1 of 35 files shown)


lib/src/argument_list_visitor.dart
```diff
@@ -213,7 +211,7 @@ class ArgumentListVisitor {
       }
 
       _visitor.builder.startSpan();
-      _argumentsAfterFunctions.visit(_visitor);
+      _argumentsAfterFunctions!.visit(_visitor);
```
stereotype441, line 214: Consider adding an assertion to the constructor to justify why this is safe, e.g.:

```dart
  ArgumentListVisitor._(
      this._visitor,
      this._leftParenthesis,
      this._rightParenthesis,
      this._allArguments,
      this._arguments,
      this._functions,
      this._argumentsAfterFunctions)
  : assert(_functions == null || _argumentsAfterFunctions != null,
        'If _functions is specified, _argumentsAfterFunctions should be too');
```

lib/src/argument_list_visitor.dart
```diff
@@ -195,14 +193,14 @@ class ArgumentListVisitor {
       // instead of just having this little solo split here. That would try to
       // keep the parameter list with other arguments when possible, and, I
       // think, generally look nicer.
-      if (_functions.first == _allArguments.first) {
+      if (_functions!.first == _allArguments.first) {
```
stereotype441, line 196: Consider adding a local variable so that these `!`s aren't necessary, e.g.:

```dart
var _functions = this._functions;
if (_functions != null) {
  // ...
  if (_functions.first == _allArguments.first) {
  ...
}
```

lib/src/argument_list_visitor.dart
```diff
@@ -447,7 +441,7 @@ class ArgumentSublist {
       rule.disableSplitOnInnerRules();
 
       // Tell it to use the rule we've already created.
-      visitor.beforeBlock(_blocks[argument], blockRule, previousSplit);
+      visitor.beforeBlock(_blocks[argument]!, blockRule!, previousSplit);
```
stereotype441, line 444: To avoid the first `!` on this line, consider doing:
```dart
var argumentBlock = _blocks[argument];
if (argumentBlock != null) {
  rule.disableSplitOnInnerRules();

  // Tell it to use the rule we've already created.
  visitor.beforeBlock(argumentBlock, blockRule!, previousSplit);
  ...
}
```

lib/src/call_chain_visitor.dart
```diff
@@ -250,7 +247,7 @@ class CallChainVisitor {
       _visitor.zeroSplit();
       _disableRule();
 
-      for (var blockCall in _blockCalls) {
+      for (var blockCall in _blockCalls!) {
```
stereotype441, line 250: Consider making a local variable to avoid this `!`

lib/src/chunk.dart
```diff
@@ -260,11 +254,11 @@ class Chunk extends Selection {
     if (_rule == null) {
       parts.add('(no split)');
     } else {
-      parts.add(rule.toString());
-      if (rule.isHardened) parts.add('(hard)');
+      parts.add(rule!.toString());
```
stereotype441, line 257: Consider making a local variable to avoid these `!`s

lib/src/debug.dart
```diff
@@ -127,30 +126,29 @@ void dumpChunks(int start, List<Chunk> chunks) {
...
       writeIf(constrainedRules.isNotEmpty,
           () => "-> ${constrainedRules.join(" ")}");
     }
 
     writeIf(chunk.indent != null && chunk.indent != 0,
         () => 'indent ${chunk.indent}');
 
-    writeIf(chunk.nesting != null && chunk.nesting.indent != 0,
-        () => 'nest ${chunk.nesting}');
+    writeIf(chunk.nesting?.indent != 0, () => 'nest ${chunk.nesting}');
```
stereotype441, line 149: If `chunck.nesting` is `null`, then this will write `nest null`.  I don't think you want that behavior.  I think you want:

```dart
writeIf(chunk.nesting != null && chunk.nesting!.indent != 0, ...);
```

or maybe this is better? (I can't decide if it's too clever):

```dart
writeIf((chunk.nesting?.indent ?? 0) != 0, ...);
```

lib/src/io.dart
```diff
@@ -15,7 +15,7 @@ import 'exceptions.dart';
 import 'source_code.dart';
 
 /// Reads input from stdin until it's closed, and the formats it.
```
stereotype441, line 17: Nit: `and the formats it` should be `and then formats it`

lib/src/line_splitting/line_splitter.dart
```diff
@@ -122,7 +119,7 @@ class LineSplitter {
   final _queue = SolveStateQueue();
 
   /// The lowest cost solution found so far.
-  SolveState _bestSolution;
+  SolveState? _bestSolution;
```
stereotype441, line 122: Does this need to be a field?  It looks like it's only used in `apply`, and assuming that `apply` is only called once for each instance of `LineSplitter` (which I *think* is the case), we could make this a local variable and the benefit of flow analysis to promote it to non-nullable sometimes.

lib/src/line_splitting/solve_state.dart
```diff
@@ -86,7 +83,7 @@ class SolveState {
 
   /// The constraints the bound rules in this state have on the remaining
   /// unbound rules.
-  Map<Rule, int> _constraints;
+  Map<Rule, int>? _constraints;
```
stereotype441, line 86: I would recommend using a late final pattern here, e.g.:
```dart
late final Map<Rule, int> _constraints = _computeConstraints();
```

And then change _ensureConstraints to something like this:

```dart
Map<Rule, int> _computeConstraints() {
  _unboundRules = <Rule>{};
  _boundRules = <Rule>{};
  
  for (var rule in _splitter.rules) {
    ...
  }

  var constraints = <Rule, int>{};

  for (var bound in _boundRules) {
    ...
  }

  return constraints;
}
```

Then your use sites no longer have to call `_ensureConstraints`.  They can just refer to `_constraints` directly.

A similar comment applies to `_unboundConstraints` and `boundRulesInUnboundLines`.

lib/src/line_splitting/solve_state.dart
```diff
@@ -197,10 +194,10 @@ class SolveState {
         for (var value = 1; value < rule.numValues; value++) {
           var boundRules = unsplitRules.clone();
 
-          List<Rule> mustSplitRules;
+          List<Rule>? mustSplitRules;
```
stereotype441, line 197: Personally I would do:
```dart
var mustSplitRules = <Rule>[];
var valid = boundRules.tryBind(_splitter.rules, rule, value, mustSplitRules.add);
```
And then change the `if (mustSplitRules != null)` block below to `if (mustSplitRules.isNotEmpty)`.

It's an extra allocation but I suspect the readability improvement will outweigh the performance cost.

lib/src/line_splitting/solve_state.dart
```diff
@@ -511,28 +508,29 @@ class SolveState {
...
+  ///
+  /// [_ensureConstraints()] should be called first.
+  Map<Rule, Set<int>> _ensureUnboundConstraints() {
+    var unboundConstraints = _unboundConstraints;
+    if (unboundConstraints != null) return unboundConstraints;
 
+    unboundConstraints = <Rule, Set<int>>{};
     for (var unbound in _unboundRules) {
-      Set<int> disallowedValues;
+      Set<int>? disallowedValues;
```
stereotype441, line 533: I'm tempted to suggest this:
```dart
late final Set<int> disallowedValues = unboundConstraints[unbound] = <int>{};
```

And then you can remove this code below:
```dart
if (disallowedValues == null) {
  disallowedValues = <int>{};
  unboundConstraints[unbound] = disallowedValues;
}
```
Basically you would be allowing lazy initialization to take care of storing `disallowedValues` in `unboundConstraints[unbound]`, only if there's actually at least one disallowed value.

But I dunno.  Maybe it's too clever to be comprehensible.  Your call.

stereotype441 (approved):
Nice work!  It was quite satisfying to see all the places where flow analysis made `as` checks unnecessary.

With one exception, my comments are nits and suggestions; take 'em or leave 'em as you see fit.

The one exception is the use of `writeIf(chunk.nesting?.indent != 0, ...)` in `lib/src/debug.dart`.  I'm pretty sure that is a genuine change in semantics that you didn't intend.  Assuming that is addressed, LGTM.

lib/src/argument_list_visitor.dart
munificent: Done.

lib/src/debug.dart
munificent: Good catch. I had tried to make `nesting` non-nullable and thus deleted the null check here. But then I ended up backing out that change. But I just tried again and was able to make it non-nullable so the null check here is indeed no longer needed.

lib/src/line_splitting/line_splitter.dart
munificent: Ah, good catch!

lib/src/line_splitting/solve_state.dart
munificent: Love it.

lib/src/line_splitting/solve_state.dart
munificent: I dig it. Added a comment to make it a little clearer.

PR #1002, opened 2021-02-27
Fix a couple of crashes I found when testing the migration on a corpus.

munificent:
There are basically three issues here:

- The debug library, which isn't part of the public API, could crash
  when logging a chunk that wasn't fully initialized. Turned
  NestingLevel from late back to nullable to fix this. This doesn't
  affect users.

- The null safety migration introduced a crash when formatting blocks
  inside string interpolation (why would someone do that?!). To fix
  that, instead of returning null chunks, changed ChunkBuilder to
  return dummy chunks. They end up getting discarded anyway.

- Cascaded index operators always crash. I need to investigate more, but
  I think this bug may exist in the previous version of dart_style,
  which is in the SDK. I'll look into that next.

### DIFF
--- a/lib/src/call_chain_visitor.dart
+++ b/lib/src/call_chain_visitor.dart
@@ -558,7 +558,7 @@ Expression _unwrapTarget(Expression node, List<_Selector> calls) {
   }
 
   // Postfix expressions.
-  if (node is IndexExpression) {
+  if (node is IndexExpression && node.target != null) {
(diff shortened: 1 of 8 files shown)


munificent:
> I think this bug may exist in the previous version of dart_style, which is in the SDK. I'll look into that next.

Phew. Good news: The stable SDK looks like it's fine. The null just flows through without causing any problems.

stereotype441 (approved):
LGTM

Issue #772, opened 2019-02-07
Is it / can it be possible to specify a line is not formatted and override and manually format just one line and let the formatter handle the rest?


twistedinferno:
I use the formatter all the time and it is really very good.  

The following is how I would like the code to be broken up
```
    test("1", () =>
      expect("a really long striiiing", "another really long striiiing"));
```

but the following is how it is formatted
```
    test(
        "1",
        () =>
            expect("a really long striiiing", "another really long striiiing"));
```

Is there a way to tell the code to stay together and not break, override the formatting just for one line?  Maybe by adding a comment??
```
    test("1", () => //
      expect("a really long striiiing", "another really long striiiing"));
```

srawlins:
I've noticed a theme in recent dartfmt complaints that it does not position arrow-functions as you might expect. I've sometimes changed code to instead use a code block, which works well in your case specifically because the closure passed to `test` is not expected to return anything:

```dart
void main() {
  test("1", () {
    expect("a really long striiiing", "aaaanother really long striiiing");
  });
}
```

I think the general theme I've seen is: When an arrow function is placed inside another statement (i.e. as a function argument), including its parameters in parens, **needs to be line-wrapped**, then things might get weird, and one might consider using a function block instead.

twistedinferno:
That is a good idea, the code block saves me a line but I would still prefer to save two!  I was guessing there might be another situation or other whereby I just want to be explicit and say "don't add a line break here"...kind of the reverse of how adding a comment or an extra comma inside a list works.
```
  [1, 2];
  [
    1,
    2,
  ];
  [
    1, //
    2
  ];
```

twistedinferno:
This is more readable
```
    test("1", () =>
      expect("a really long striiiing", "another really long striiiing"));
```

than this
```
    test(
        "1",
        () =>
            expect("a really long striiiing", "another really long striiiing"));
```

munificent:
The question is answered [in the FAQ](https://github.com/dart-lang/dart_style/wiki/FAQ#why-cant-i-tell-the-formatter-to-ignore-a-region-of-code). (It comes up a lot.)

For your specific case here, like @srawlins suggests, this code will look more idiomatic if you use `{}` instead of `=>` for the function.

PR #1012, opened 2021-03-29
Don't indent nested ternaries

### LINKED ISSUE #722: Ternary preference
My preference for chained ternaries

```dart
child: 
    project.locked ? Icon(Icons.lock)
  : project.fav ? Icon(Icons.star)
  : project.taps == null ? Icon(Icons.notifications)
  : Text(
      suffixNumber(project.taps),
      textAlign: TextAlign.center,
      style: TextStyle(
        fontSize: 18.0, 
        fontWeight: FontWeight.w600,
      ),
    ),
```

`dartfmt` preference
```dart
child: project.locked
    ? Icon(Icons.lock)
    : project.fav
        ? Icon(Icons.star)
        : project.taps == null
            ? Icon(Icons.notifications)
            : Text(
                suffixNumber(project.taps),
                textAlign: TextAlign.center,
                style: TextStyle(
                  fontSize: 18.0,
                  fontWeight: FontWeight.w600,
                ),
              ),
```

Is there a reason why the second format has been selected? Is there any room for discussion to adopt the aforementioned style?

RedHatter:
Fixes #722

Don't indent ternary expressions that are in the else leg of another ternary. This change makes nested ternaries used as a series of `else if` statements much more readable.

This
```
String type = status == 'OK'
      ? 'notices'
      : status == 'NO'
      ? 'warnings'
      : status == 'BAD'
      ? 'errors'
      : '';
```
instead of 
```
String type = status == 'OK'
      ? 'notices'
      : status == 'NO'
        ? 'warnings'
        : status == 'BAD'
          ? 'errors'
          : '';
```

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -821,7 +821,10 @@ class SourceVisitor extends ThrowingAstVisitor {
 
   @override
   void visitConditionalExpression(ConditionalExpression node) {
-    builder.nestExpression();
+    // Don't nest if this expression is the else leg of another ternary
+    final shouldNest = !(node.parent is ConditionalExpression &&
+        (node.parent as ConditionalExpression).elseExpression == node);
+    if (shouldNest) builder.nestExpression();
 
     // Start lazily so we don't force the operator to split if a line comment
     // appears before the first operand. If we split after one clause in a
@@ -856,7 +859,7 @@ class SourceVisitor extends ThrowingAstVisitor {
     builder.endRule();
     builder.endSpan();
     builder.endBlockArgumentNesting();
-    builder.unnest();
+    if (shouldNest) builder.unnest();
   }
 
   @override
--- a/test/regression/0400/0407.unit
+++ b/test/regression/0400/0407.unit
@@ -12,22 +12,22 @@ void main() {
       ..tags = (new Account_Tags()
         ..accountHotlist.add(new Hotlist()..hotlistId = new Int64(10))));
 }
->>> (indent 4)
+>>> (indent 6)
 main() {
 receiver
       ..formattedTotal = _total == 0
           ? ""
           : _chartType == "PieChart"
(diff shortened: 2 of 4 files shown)


RedHatter:
@googlebot I signed it!

RedHatter:
@googlebot I signed it!

munificent:
This is great, thank you!

jamesderlin:
Although the indentation creep was annoying, I think it was clearer.  After looking at some code that was formatted with this:

```dart
  var x= foo
      ? bar
      : baz
      ? qux
      : garply
      ? waldo
      : fred;
```
it's not immediately obvious to me what the leaves are. (This isn't true for an `if`-`else if` chain where it's much more obvious which expressions are conditions.)

Or maybe I just need to get accustomed to this new style.

munificent:
> This isn't true for an `if`-`else if` chain

It is basically like an `if` - `else if` chain. Read it like:

```
  var x = /* if */    foo
      /* then */    ? bar
      /* else if */ : baz
      /* then */    ? qux
      /* else if */ : garply
      /* then */    ? waldo
      /* else */    : fred;
```

So the `?` lines are "then", and the `:` lines are "else if" or "else".

munificent:
OK, I'm seeing enough unsatisfied users that I'm going to roll this change back to the previous behavior. When the initial request was filed to flatten the indentation, there was almost no pushback, but I guess that's because people happy with the current behavior didn't see the issue.

jamesderlin:
> > This isn't true for an `if`-`else if` chain
> 
> It is basically like an `if` - `else if` chain. Read it like:

The different parts of an  `if`-`else if` chain are visibly distinct:
* Conditions are enclosed in parentheses.
* The "then" or "else" bodies are indented.

FWIW, I'd be totally fine if Dart had, say, `if`-expressions that could be used outside of collection literals. ;)

lukepighetti:
As the author of #772 I am fully in support of reverting this change

PR #1033, opened 2021-06-11
Simpler cascades

munificent:
This PR proposes three changes to how cascades are formatted. In principle, we could do them separately, but the first two made it easier to implement the third, so I lumped them together.

I tested these changes on 6,413,157 lines of pub packages. The new formatting rules lead to **1 changed line for every ~1100 lines of code** (or ~0.09% of lines). Despite that small output change, performance is significantly improved. On my MacBook Pro formatting the corpus went from 260.84s to 111.83s. (In other words, this change **makes dart format more than twice as fast overall.**)

Unless I hear compelling argument otherwise, **I plan to merge these changes into master next week. If you have feedback, please provide it on this PR. In particular, I’ll look at 👍 and 👎 to gauge overall sentiment.**

For the curious, here are the three changes in detail:

### [Disallow multiple cascade sections on a single line](https://github.com/dart-lang/dart_style/commit/f0fc7f7566d4a29d7e17ce7d075a4f9c780a6190)

(Submitted by @ManoyloK. Thanks!)

When we first implemented formatting for cascades, the feature was pretty new and we weren't sure how users would use it. Since the whole point of the feature is brevity, it seemed sensible to support packing multiple cascades into a single line. Otherwise, why not just use multiple statements? At the same time, the low precedence of cascades can make single-line cascades extremely confusing to read. Few Dart users will interpret this correctly:

```dart
var x = condition ? whenTrue : whenFalse..cascade();
```

Whereas this is (somewhat) more clearer:

```dart
var x = condition ? whenTrue : whenFalse
  ..cascade();
```

(The `..cascade()` is applied to the result of the entire `?:` expression, not just the `whenFalse` clause).

To balance that, the formatter only allowed multiple cascades on a single line *if they were all the same method name*. That allowed idiomatic cases like:

```dart
var buffer = StringBuffer()..write('text')..write('more text');
```

In practice, it's not clear that this rule was ever loved and we got feedback that it was actively disliked by some. In general, users say they want less code packed into single lines. So we removed this special case. The rules now are that:

1.  A cascade will only stay on the same line as its receiver if the receiver is fairly high precedence expression (like a method call). That avoids the confusing case above while still allowing code like:

    ```dart
    var sortedItems = iterable.toList()..sort();
    ```

2.  If there are multiple cascade sections, they always split onto multiple lines, including the first.

In the process of reviewing this, I discovered a bug in the old rules:

```dart
   var library =
       Library((b) => b..body.addAll([buildMixin()])..directives.addAll([]));
```

Here, those two `addAll()` calls are *not cascading on the same receiver*, so should definitely not be on the same line. This change also fixes that bug. Here are a couple of other examples of the change:

```dart
// before:
       final val = StringBuffer(prefix)..write(pluralPrefix)..write(' ');
// after:
       final val = StringBuffer(prefix)
         ..write(pluralPrefix)
         ..write(' ');

// before:
  Set<Id> idSet = {}..addAll(idValuePerId.keys)..addAll(actualDataPerId.keys);
// after:
  Set<Id> idSet = {}
    ..addAll(idValuePerId.keys)
    ..addAll(actualDataPerId.keys);

// before:
    return BuilderOptions({}..addAll(config)..addAll(other.config),
        isRoot: other.isRoot);
// after:
    return BuilderOptions(
        {}
          ..addAll(config)
          ..addAll(other.config),
        isRoot: other.isRoot);
```

Note that, in most cases, the affected code can more idiomatically be written today using collection literals and spreads:

```dart
  Set<Id> idSet = {...idValuePerId.keys, ...actualDataPerId.keys};

    return BuilderOptions({...config, ...other.config}),
      isRoot: other.isRoot)
```

### [Use normal indentation for cascade receivers](https://github.com/dart-lang/dart_style/commit/d0a102b721a1b07a27eae01fc408f61d52bad827)

The formatter generally indents subexpressions farther than the expression that contains them, as in:

```dart
var milliseconds = (position.inMilliseconds +
        ((DateTime.now().millisecondsSinceEpoch -
                updateTime.millisecondsSinceEpoch) *
            (speed ?? 1.0)))
    .toInt();
```

Here, the `.toInt()` method is called on the result of the entire preceding expression, so the three lines above it are indented farther.

Because cascade syntax is weird and we weren't sure how users would take to it, we added a special rule that inverted that natural indentation when method chains and cascades were mixed:

```dart
var lock = File(lockFilePath(workingDirectory))
    .openSync(mode: FileMode.write)
      ..lockSync();
```

Here, `..lockSync()` is indented farther, even though that cascade is called on the result of the entire preceding expression. I removed that special rule and reverted to normal indentation, which gives you:

```dart
var lock = File(lockFilePath(workingDirectory))
    .openSync(mode: FileMode.write)
  ..lockSync();
```

In practice, since cascades are indented +2 while other expressions are +4, the result is that the cascade loses four spaces of indentation. Here are a couple of before/after examples:

```dart
// before:
  factory BkTradeList._fromMap(Map<String, dynamic> json) => BkTradeList(
        error: json['error'],
        trades: (json['result'] as List<dynamic>)
            .map((e) => BkTradeInfo._fromList(e))
            .toList()
              ..removeWhere((e) => e == null),
      );
// after:
  factory BkTradeList._fromMap(Map<String, dynamic> json) => BkTradeList(
        error: json['error'],
        trades: (json['result'] as List<dynamic>)
            .map((e) => BkTradeInfo._fromList(e))
            .toList()
          ..removeWhere((e) => e == null),
      );

// before:
    var lock = File(lockFilePath(workingDirectory))
        .openSync(mode: FileMode.write)
          ..lockSync();
// after:
    var lock = File(lockFilePath(workingDirectory))
        .openSync(mode: FileMode.write)
      ..lockSync();

// before:
    var stdoutLines = process.stdout
        .transform(Utf8Decoder())
        .transform(LineSplitter())
        .asBroadcastStream()
          ..listen((line) {
            stdoutSink?.add(line);
            printOnFailure(line);
          });
// after:
    var stdoutLines = process.stdout
        .transform(Utf8Decoder())
        .transform(LineSplitter())
        .asBroadcastStream()
      ..listen((line) {
        stdoutSink?.add(line);
        printOnFailure(line);
      });

// before:
    final terminators = (mode.contains
            .map((c) =>
                c.beginKeywords != null ? '\\.?(?:${c.begin})\\.?' : c.begin)
            .toList()
              ..addAll([mode.terminatorEnd, mode.illegal]))
        .where((x) => x != null && x.isNotEmpty)
        .toList();
// after:
    final terminators = (mode.contains
            .map((c) =>
                c.beginKeywords != null ? '\\.?(?:${c.begin})\\.?' : c.begin)
            .toList()
          ..addAll([mode.terminatorEnd, mode.illegal]))
        .where((x) => x != null && x.isNotEmpty)
        .toList();
```

### [Optimize formatting large cascade expressions](https://github.com/dart-lang/dart_style/commit/47248ba01bcc20ff2239ddc10e04a64ba055a597)

The last change is the most valuable, but doesn't change the output at all. The formatter runs [fairly complex constraint solving][solve] when deciding how to best line split an expression. This solving is exponential in the worst case. In order to avoid locking up entirely, the formatter will give up and yield long lines if it takes too long to format a single expression.

[solve]: http://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/

Nested collection literals and function bodies avoid that by being line split separately. That lets each element in a huge collection literal be formatted independently. Unfortunately, cascades didn't get the same optimization, in part because of the special rule allowing multiple sections on one line. The last change is to get that optimization working for cascades.

Before this change, the ten slowest files to format in the corpus were:

```
0:00:22.432430: streaming.pb.dart
0:00:14.911033: permissions.pb.dart
0:00:13.586810: bridge.pb.dart
0:00:11.256575: descriptor.pb.dart
0:00:10.782197: messages.pb.dart
0:00:07.168568: channels.pb.dart
0:00:06.210043: guilds.pb.dart
0:00:05.341587: types.pb.dart
0:00:05.302011: messages.pb.dart
0:00:04.976277: bledata.pb.dart
```

Those are seconds, so >22 seconds for a single file (!). Note that those are all generated protobuf libraries, which use cascades heavily. After the optimization, the ten slowest files are:

```
0:00:01.893950: parser.dart
0:00:01.572537: trackball_renderer.dart
0:00:01.232390: slider_theme.dart
0:00:01.221646: range_selector_theme.dart
0:00:01.186164: range_slider_theme.dart
0:00:00.936228: drawer3.dart
0:00:00.854479: user_interaction/tooltip_painter.dart
0:00:00.746739: codes_generated.dart
0:00:00.721166: iversion_prompt_dialog.dart
0:00:00.686001: drawer2.dart
```

The cascade-heavy libraries no longer show up in the list at all. (These times are still pretty slow. Eventually, I'll try to apply the same optimization to large argument lists, which should fix these.)

If you'd like to see the full diff of how these changes affect a corpus of pub packages, it's [here][diff].

[diff]: https://gist.github.com/munificent/ae39fda42b2d9f831c6cf3ebc2ad1455


munificent:
Hey, @ManoyloK. I am working on incorporating your change along with some other improvements to cascades in this branch. Would you mind commenting here to let the bot know you are OK with me merging your change into this branch? Thanks!

ManoyloK:
@googlebot I consent.

lib/src/source_visitor.dart
```diff
@@ -621,8 +609,90 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+  ///
+  ///     var x = someLeadingExpression
+  ///       ..firstCascade()
+  ///       ..secondCascade()
+  ///       ..thirdCascade()
+  ///       ..fourthCascade();
+  ///
+  /// To handle that, we don't put the last section in the block and instead
+  /// format it with the surrounding expression. So, from the formatter's
+  /// view, a cascade like:
```
natebosch, line 637: [nit] We have the example cascade above. I would phrase this as

```
  /// format it with the surrounding expression. So, from the formatter's
  /// view, a cascade is formatted like:
  ///
  ///     var x = someLeadingExpression
  ///       [ begin block ]
  ///       ..firstCascade()
  ///       ..secondCascade()
  ///       ..thirdCascade()
  ///       [ end block ]
  ///       ..fourthCascade(); 
 ```

File: tool/change_metrics.dart [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+        totalFiles++;
+        totalLines += lines.length;
+      } catch (error) {
+        print('Could not read ${entry.path}:\n$error');
+      }
+    }
+  }
+
+  print('Getting diff stats...');
+  var result = Process.runSync('git', ['diff', '--shortstat'],
```
natebosch, line 40: You might consider adding

```dart
environment: {
  'GIT_CONFIG_NOGLOBAL': 'true',
  'GIT_CONFIG_NOSYSTEM': 'true',
}
```

This avoids breaking if a user or system level config changes the behavior of the ` git diff` command.

tool/change_metrics.dart
```diff
@@ -0,0 +1,71 @@
...
+      workingDirectory: directory);
+  if (result.exitCode != 0) {
+    print('Git failure:\n${result.stdout}\n${result.stderr}');
+    exit(1);
+  }
+
+  var stdout = result.stdout as String;
+  var insertions = _parseGitStdout(stdout, r'(\d+) insertions');
+  var deletions = _parseGitStdout(stdout, r'(\d+) deletions');
```
natebosch, line 49: If a formatting change only adds or removes blank lines one of these parses will fail.

This might be rare enough to ignore since the interesting changes are likely to move code and not just add or remove blank lines?

tool/change_metrics.dart
```diff
@@ -0,0 +1,71 @@
...
+  var changes = insertions + deletions;
```
natebosch, line 50: Changes that are only on a single line are going to be double counted here - for example changing indentation in 1 line will show up as a delta of 2. This might also be fine to ignore as an approximation? I would expect that checking only inserted lines might be a better approximation for most changes though...

tool/change_metrics.dart
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:io';
+
+/// Calculates the amount of formatting changes in a given directory of code.
```
natebosch, line 6: Can this go out in a separate PR?

lib/src/source_visitor.dart
```diff
@@ -621,8 +609,90 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+    // Rule to split before the first `..`.
+    builder.startLazyRule(Rule.hard());
+    visit(node.target);
+
+    builder.nestExpression(indent: Indent.cascade, now: true);
+    builder.startBlockArgumentNesting();
+
+    zeroSplit();
+
+    // If there are comments before the first section, keep them outside of the
```
natebosch, line 667: Why is this necessary?

natebosch (approved):

tool/change_metrics.dart
munificent: I, uh, in fact, didn't intend to commit this yet.

I'll take it out of the PR. :)

lib/src/chunk.dart
```diff
@@ -290,10 +290,13 @@ class ChunkBlock {
   /// may need extra expression-level indentation.
   final Chunk? argument;
 
+  /// Whether the first chunk should have a level of indentation before it.
```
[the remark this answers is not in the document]
munificent: > I'm not sure how to interpret "a level of indentation" - I would consider "no indentation" to be "a level" of indentation.

The docs in dart_style style tend to treat "level" as a non-mass noun, not a numeric attribute. So 6 spaces would be "3 levels of indentation" and zero spaces means "no levels of indentation". I was never able to come up with a better noun. :-/

lib/src/chunk_builder.dart
```diff
@@ -589,14 +589,12 @@ class ChunkBuilder {
   /// Starts a new block as a child of the current chunk.
   ///
   /// Nested blocks are handled using their own independent [LineWriter].
-  ChunkBuilder startBlock(Chunk? argumentChunk) {
+  ChunkBuilder startBlock(Chunk? argumentChunk, {bool indent = true}) {
```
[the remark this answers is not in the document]
munificent: Done.

lib/src/chunk.dart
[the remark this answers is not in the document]
natebosch: > How many levels the other thing has.

What is "the other thing"?

lib/src/source_visitor.dart
munificent: Done. (Copy/paste error.)

tool/change_metrics.dart
munificent: It's good enough for broadcast, but yeah. Removed this file from the PR.

tool/change_metrics.dart
munificent: I think it's OK if some lines are double counted. The goal is to give a rough measure of impact and over-estimating that is OK. It's probably generally better to be conservative with formatting changes.

lib/src/source_visitor.dart
munificent: The way nested blocks work is a little inelegant. It's part of the formatter that evolved over time and is more bolted on than I would like. Say you have code like:

```dart
var list = [ // hi
  one,
  two,
  three
];
```

Then the `// hi` gets claimed by the inner block and formatted like:

```dart
var list = [
  [ begin block ]
  // hi
  one,
  two,
  three
  [ end block ]
];
```

That generally looks right for collection literals and lambdas (which is what the nested chunk stuff was added for), but no for cascades. If you have:

```dart
receiver // hi
  ..cascade()
  ..another();
```

We want it to start formatted like that and not end up as:

```dart
receiver
  // hi
  ..cascade()
  ..another();
```

Keeping the comment outside of the block preserves that. Added a comment to explain.

(The formatter's codebase is not super heavy on tech debt, but there is definitely a simpler formatter hiding inside it trying to get out that I would love to surface if I had the ability to simplify some of the formatting rules and the time to do a massive refactor.)

tool/change_metrics.dart
munificent: Done.

Hixie:
It would be interesting to see the diff that results from comparing the result of running the current formatter on flutter/flutter vs running this formatter on that same code. (e.g. open a PR on flutter/flutter where you have one commit that formats everything today, then a second commit that formats things using this.)

munificent:
Setting up a PR might be a little more trouble than it's worth, but I can generate a diff locally pretty easily. Here are all of the changes: https://gist.github.com/munificent/7f0b5e0f287341f559a2f04cfd79c06e

Most of the changes are just removing a little indentation on cascades whose receiver is a split method chain, like all the `..sort()` changes. [This one](https://gist.github.com/munificent/7f0b5e0f287341f559a2f04cfd79c06e#file-flutter-diff-L53-L61) is an example where the old formatter was timing out and giving up on the large cascade where the new formatter can handle it. [This](https://gist.github.com/munificent/7f0b5e0f287341f559a2f04cfd79c06e#file-flutter-diff-L314-L317) is another example of the bug where it puts two cascades on the same line even though they are on different receivers.

I'm curious to hear what you think. Let me know if you think it's worth putting together a real PR on the Flutter repo to see it as a nicer diff.

Hixie:
If you have a checkout it should be trivial to get a PR, just push back to the origin or upstream (depending on if it's the kind of checkout you get from installing the tarball or the kind of checkout you get from following our contributor guide), and it'll give you a link to create the PR.

Hixie:
Anyway, overall the changes seem neutral or positive, I don't see anything that got worse (other than things that would have been worse before but for the timeout thing). So this gets my upvote. :-)

munificent:
> Anyway, overall the changes seem neutral or positive, I don't see anything that got worse (other than things that would have been worse before but for the timeout thing).

Yeah, that's my impression too. I don't think idiomatic Flutter code leans very heavily on cascades, so it's not affected much.

munificent:
Well this looks like a pretty clean win. :) Thanks, everyone!

PR #1036, opened 2021-06-18
Return correct exit code from FormatCommand when formatting stdin.

### LINKED ISSUE #1035: dart format should exit non-zero if the source code has errors
`echo 'not dart' | dartfmt` and `echo 'not dart' | dart format` both produce equivalent error messages, but `dartfmt` exits with code `65` while `dart format` exits with `0`.

We use the exit code in `dart-vim-plugin` to know whether the output should be treated as the formatted file, or as errors. Unfortunately in vim it isn't trivial to distinguish between `stderr` and `stdout` so using the output stream to determine success would take some additional hacking and running the command twice.

munificent:
It reads from stdin asynchronously. Before this fix, the FormatCommand
did not wait for that to complete before returning the exit code, so it
would return before stdin had been formatted and any parse error
detected.

Fix #1035.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 2.0.2-dev
 
 * Don't unnecessarily split argument lists with `/* */` comments (#837).
+* Return correct exit code from `FormatCommand` when formatting stdin (#1035).
 
 # 2.0.1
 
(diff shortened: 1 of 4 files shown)


natebosch (approved):

mit-mit:
```
Formatted 47 files (1 changed) in 0.80 seconds.
Error: Process completed with exit code 1.
```

Oh the irony when a `dart_style` PR fails the formatting check; you should use for formatter Bob! 😂

PR #1041, opened 2021-07-01
Split empty catch blocks with finally clauses.

### LINKED ISSUE #1029: Formatting of empty catch blocks can appear misleading
`dart format` really likes to collapse empty `{}` blocks onto a single line.  However, in the case of a `try`-`catch`-`finally` chain,  I think that it would be better not to do that if there is an non-empty block of statements later in the chain.  For example:

```dart
void main() {
  try {
    doSomething();
  } on Exception catch (e) {
  } finally {
    cleanupSomething();
  }
}
```
`dart format` (version 2.13.1 stable) formats it as:
```dart
void main() {
  try {
    doSomething();
  } on Exception catch (e) {} finally {
    cleanupSomething();
  }
}
```
which I think is rather misleading since it appears that `cleanupSomething()` is called only in response to a thrown `Exception`.

munificent:
Fix #1029

### DIFF
--- a/lib/src/nesting_level.dart
+++ b/lib/src/nesting_level.dart
@@ -24,8 +24,7 @@ import 'fast_hash.dart';
 class NestingLevel extends FastHash {
   /// The nesting level surrounding this one, or `null` if this is represents
   /// top level code in a block.
-  NestingLevel? get parent => _parent;
-  final NestingLevel? _parent;
+  final NestingLevel? parent;
 
   /// The number of characters that this nesting level is indented relative to
   /// the containing level.
@@ -40,13 +39,13 @@ class NestingLevel extends FastHash {
   int get totalUsedIndent => _totalUsedIndent!;
   int? _totalUsedIndent;
 
-  bool get isNested => _parent != null;
+  bool get isNested => parent != null;
 
   NestingLevel()
-      : _parent = null,
+      : parent = null,
         indent = 0;
 
-  NestingLevel._(this._parent, this.indent);
+  NestingLevel._(this.parent, this.indent);
 
   /// Creates a new deeper level of nesting indented [spaces] more characters
   /// that the outer level.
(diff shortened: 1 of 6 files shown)


lib/src/nesting_level.dart
```diff
@@ -25,7 +25,7 @@ class NestingLevel extends FastHash {
   /// The nesting level surrounding this one, or `null` if this is represents
   /// top level code in a block.
   NestingLevel? get parent => _parent;
-  NestingLevel? _parent;
+  final NestingLevel? _parent;
```
natebosch, line 28: Now that it is `final` you can make the field public and remove the getter.

lib/src/source_visitor.dart
```diff
@@ -2034,7 +2022,6 @@ class SourceVisitor extends ThrowingAstVisitor {
     token(node.rightParenthesis);
     builder.unnest();
 
-    @override
```
natebosch, line 2037: How is this related to the rest of the CL?

Did the analyzer classes remove this method?

natebosch (approved):

lib/src/nesting_level.dart
munificent: Ah, good catch. Done.

lib/src/source_visitor.dart
munificent: It's not related, I just noticed this in passing. This is a _local_ function. That metadata annotation never did anything useful. I must have accidentally copied it from a method at some point.

jamesderlin:
Is this going to work only if there's a `finally` block?  Will something like:

```dart
void main() {
  try {
    doSomething();
  } on FooException {
  } on BarException {
    doSomething();
  }
}
```
be left alone, or will it be transformed to:

```dart
void main() {
  try {
    doSomething();
  } on FooException {} on BarException {
    doSomething();
  }
}
```
?

munificent:
Ah, good question. I did not cover that case, but it would be good to do so. I'll take a stab at it.

munificent:
Fix out: https://github.com/dart-lang/dart_style/pull/1042

Issue #1051, opened 2021-09-09
Error: 'FunctionReference' isn't a type


pb-iatros:
The upgrade from 2.0.3 to 2.1.0 throws an error when running build_runner

`$ flutter packages pub run build_runner build --delete-conflicting-outputs`

```
Failed to precompile build_runner:build_runner:
../../snap/flutter/common/flutter/.pub-cache/hosted/pub.dartlang.org/dart_style-2.1.0/lib/src/source_visitor.dart:1809:31: Error: Type 'FunctionReference' not found.
  void visitFunctionReference(FunctionReference node) {
                              ^^^^^^^^^^^^^^^^^
../../snap/flutter/common/flutter/.pub-cache/hosted/pub.dartlang.org/dart_style-2.1.0/lib/src/source_visitor.dart:1809:31: Error: 'FunctionReference' isn't a type.
  void visitFunctionReference(FunctionReference node) {
                              ^^^^^^^^^^^^^^^^^
```

srawlins:
I think the analyzer dep needs to be bumped to `^1.6.0`:

https://pub.dev/packages/analyzer/changelog

If you run `pub upgrade` (or equivalent flutter command) and retry, does it still crash?

pb-iatros:
We have a couple of circular dependencies that block the analyzer update, so we can't test that in the short term.
Meanwhile, we hope that someone is able to test your suggestion

srawlins:
That's interesting. dart_style 2.1.0 requires analyzer `^1.3.0`. If you have some dependency that blocks upgrading to analyzer 1.6.0, the only situation I can think of is a dependency which has specified an explicit upper bound on analyzer's version which excludes 1.6.0. (A constraint with caret, like `^1.x.0`, should allow `1.6.0` for any `x` <= 6.) Do you know which package is allowing analyzer 1.3.0 but blocking analyzer 1.6.0?

dened:
We did nothing(didn't upgrade any package), but have the same problems on CI.

dened:
Ok, I broke my local code, did 'flutter pub upgrade', and have transitive dev_dependencies: dart_style 2.1.0 now.

dened:
Adding override in pubspec.yaml  works
```
dependency_overrides:
  dart_style: 2.0.3
```

munificent:
[...]

PR #1058, opened 2021-09-21
Revert support for constructor tear-offs and republish as 2.1.1.

munificent:
This enables pub to avoid selecting 2.1.0, which has a bad constraint
on analyzer (#1051).


munificent:
This is a clean patch onto master which explicitly reverts all of the changes in 2.1.0 which rely a version of the analyzer that dart_style's constraint isn't tight enough to require. After this lands, I'll publish it and then send out another PR that brings the constructor tear-off back and fixes the analyzer constraint.

natebosch (approved):

PR #1062, opened 2021-09-27
Temp apply import shorthand

munificent:
DO NOT MERGE.

This is just so we can see the diff.

### DIFF
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -1,16 +1,16 @@
 // Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
-import 'dart:io';
+import dart:io;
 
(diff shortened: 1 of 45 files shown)


bin/format.dart
```diff
@@ -1,16 +1,16 @@
...
-import 'package:dart_style/src/cli/options.dart';
-import 'package:dart_style/src/cli/output.dart';
-import 'package:dart_style/src/cli/show.dart';
-import 'package:dart_style/src/cli/summary.dart';
-import 'package:dart_style/src/io.dart';
-import 'package:dart_style/src/style_fix.dart';
+import dart:io;
+
+import args;
+import /src/cli/formatter_options;
```
kevmoo, line 7: I worry about confusion here w/ a similar file in the bin directory ... hrm

munificent: Yeah, we could do `:` instead of `/` here (which is what the proposal has).

Issue #1071, opened 2021-10-29
Extra space after comment inside generic


timsneath:
I've commented out the type of `Pointer` for various reasons that aren't relevant to this issue, but I leave it in for the benefit of readers.

The extra space in the second example is a little odd, in that it makes it look like a less-than sign rather than the end of the generic.

```dart
  external Pointer/*<SAFEARRAY>*/ parray;
  external Pointer<Pointer/*<SAFEARRAY>*/ > pparray;
```

munificent:
Yeah, I agree that isn't ideal. Should hopefully be an easy fix.

HelioStrike:
Hey! Can I give this issue a shot? I haven't contributed to this repo before so I'm not sure what the rules of the land are.

munificent:
I spent some time digging into this and, unfortunately, it's harder than it seems. Back before Dart 2.0, Dart had no support for generic methods. We started adding support for them in the "strong mode" type checker, but there was no syntax to actually write them. To work around that, users could write them using special comments and the parser used by strong mode would look for those.

To make those comments format less weird, the formatter would recognize them and not put spaces before and after them. The contents inside the comments in the issue Tim filed just *happens* to have the same structure used to recognize those magic generic method comments, but that's not always the case. The formatter also shouldn't put a space in code like:

```dart
List<int /* comment */> list;
```

But in this case, the `comment` doesn't look like a generic method comment, so your change won't recognize it.

Ideally, the formatter would never put a space between a comment in a generic parameter/argument list before the `>`. The problem is that at the point that it makes that decision, it doesn't know if the `>` is for generics or is actually an infix greater-than operator. We definitely *do* want spaces before `>` when used as an operator.

The right solution would be to plumb that context all the way through from SourceVisitor to ChunkBuilder. But that's a lot more work than your patch and probably not worth the effort given the rarity and harmlessness of the bug. The simplest thing to do is to just accept that, yes, you'll get a slightly funny looking space here.

Also, now that Dart 2.0 is far in the past, I can probably tear out the old code for treating generic method comments specially.

munificent:
The [forthcoming](https://github.com/dart-lang/dart_style/issues/1403) [tall style](https://github.com/dart-lang/dart_style/issues/1253) doesn't have any special support for the ancient pre-Dart 2.0 generic method comments, so you get:

```dart
  external Pointer /*<SAFEARRAY>*/ parray;
[...]

PR #1087, opened 2021-12-30
Remove extra space after comment inside generic #1071

HelioStrike:
Solves #1071.

### DIFF
--- a/lib/src/chunk_builder.dart
+++ b/lib/src/chunk_builder.dart
@@ -819,8 +819,10 @@ class ChunkBuilder {
     // Not at the beginning of a line.
     if (!_chunks.last.canAddText) return false;
 
-    // Magic generic method comments like "Foo/*<T>*/" don't get spaces.
-    if (_isGenericMethodComment(comments.last) && token == '(') {
+    // Magic generic method comments like "Foo/*<T>*/" and
+    // comments inside a generic like "Foo<X/*<Y>*/>" don't get spaces.
+    if (_isGenericMethodComment(comments.last) &&
+        (token == '(' || token == '>')) {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


munificent:
Thank you for giving this one a try and I'm sorry for the very long delay. I've been busy with language stuff lately and haven't had much bandwidth for dart_style.

I spent some time digging into this and, unforunately, I don't think the fix you have here generalizes well enough.

It builds on what is basically an old hack in the formatter. Back before Dart 2.0, Dart had no support for generic methods. We started adding support for them in the "strong mode" type checker, but there was no syntax to actually write them. To workaround that, users could write them using special comments and the parser used by strong mode would look for those.

To make those comments format less weird, the formatter would recognize them and not put spaces before and after them. The contents inside the comments in the issue Tim filed just *happens* to have the same structure used to recognize those magic generic method comments, but that's not always the case. The formatter also shouldn't put a space in code like:

```dart
List<int /* comment */> list;
```

But in this case, the `comment` doesn't look like a generic method comment, so your change won't recognize it.

Ideally, the formatter would never put a space between a comment in a generic parameter/argument list before the `>`. The problem is that at the point that it makes that decision, it doesn't know if the `>` is for generics or is actually an infix greater-than operator. We definitely *do* want spaces before `>` when used as an operator.

The right solution would be to plumb that context all the way through from SourceVisitor to ChunkBuilder. But that's a lot more work than your patch and probably not worth the effort given the rarity and harmlessness of the bug. The simplest thing to do is to just accept that, yes, you'll get a slightly funny looking space here.

I'm going to close this PR, but I really do appreciate you taking a look at this. I love external contributions, but the formatter can be more challenging to hack on than it seems. Even simple changes tend to have far-reaching consequences. :-/

PR #1093, opened 2022-02-12
Stop using 'astFactory', use specialized internal API.

scheglov:
Clients should not build AST manually like this. There is one right way to do this - parse.

I was able to move almost all clients off the `AstFactory`, and `dart_style` is the last one. Unfortunately I was not able to find any better solution than to provide such specialized internal API for it.


scheglov:
ping

scheglov:
ping

munificent:
Sorry for the delay and thanks for the clean up!

PR #1094, opened 2022-02-17
Formatting support for named arguments anywhere.

### LINKED ISSUE #1072: dart_style implementation for named arguments anywhere
The "[named arguments anywhere](https://github.com/dart-lang/sdk/issues/47451)" language change allows named arguments to come before positional ones. Today, the argument list formatter assumes all named arguments will appear at the end of the argument list. Fixing this—and even figuring out *how* these mixed arguments lists should be formatted—might be pretty hard.

munificent:
I opted to add support for this while being minimally invasive to the
existing style and formatting. All existing code which does not have any
positional arguments after named ones retains its previous formatting.

For new argument lists that use positional arguments after named ones,
I try to mostly follow the existing style rules even though they can be
fairly complex. In particular, it will be pretty aggressive about
applying block-style formatting to function literals inside argument
lists even with named args anywhere. I think that's important to support
the main use case I know of for the feature which is trailing positional
closures like:

```dart
function(named: 1, another: 2, () {
  block...
});
```

In argument lists using named args anywhere that don't have block
functions, it treats all arguments like named ones. That provides nice
simple formatting like:

```dart
function(
    argument1,
    named: argument2,
    argument3,
    another: argument4);
```

I think that does a good job of highlighting which arguments are named,
which is what we want.

Fix #1072.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.2.2
+
+* Format named arguments anywhere (#1072).
+
 # 2.2.1
 
 * Require `package:analyzer` version `2.6.0`.
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -294,11 +294,15 @@ class ArgumentSublist {
   /// The full argument list from the AST.
   final List<Expression> _allArguments;
 
-  /// The leading positional arguments, in order.
+  /// If all positional arguments occur before all named arguments, then this
+  /// contains the positional arguments, in order. Otherwise (there are no
+  /// positional arguments or they are interleaved with named ones), this is
+  /// empty.
   final List<Expression> _positional;
 
-  /// The named arguments, in order, and any positional arguments that follow
-  /// at least one named argument.
+  /// The named arguments, in order. If there are any named arguments that occur
+  /// before positional arguments, then all arguments are treated as named and
+  /// end up in this list.
   final List<Expression> _named;
 
   /// Maps each block argument, excluding functions, to the first token for that
--- a/lib/src/argument_list_visitor.dart
+++ b/lib/src/argument_list_visitor.dart
@@ -63,87 +63,9 @@ class ArgumentListVisitor {
       Token leftParenthesis,
       Token rightParenthesis,
       List<Expression> arguments) {
-    // Look for a single contiguous range of block function arguments.
-    int? functionsStart;
-    int? functionsEnd;
+    var functionRange = _contiguousFunctions(arguments);
 
-    for (var i = 0; i < arguments.length; i++) {
-      var argument = arguments[i];
-      if (_isBlockFunction(argument)) {
-        functionsStart ??= i;
-
-        // The functions must be one contiguous section.
-        if (functionsEnd != null && functionsEnd != i) {
-          functionsStart = null;
-          functionsEnd = null;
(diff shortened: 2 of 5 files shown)


lib/src/argument_list_visitor.dart
```diff
@@ -224,6 +146,83 @@ class ArgumentListVisitor {
     if (_isSingle) _visitor.builder.endSpan();
   }
 
+  /// Look for a single contiguous range of block function [arguments] that
+  /// should receive special formatting.
+  ///
+  /// Returns a list of (start, end] indexes if found, otherwise returns `null`.
+  static List<int>? _contiguousFunctions(List<Expression> arguments) {
+    int? functionsStart;
```
munificent, line 154: This code is essentially unchanged from line 66 on. The main difference is that the old code used to use `arguments[0] is NamedExpression` as a clever check to see if all arguments are named. That doesn't work with named args anywhere so now it iterates the list and checks them all.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.2.2
```
natebosch, line 1: How do you normally choose when to bump the feature version? Support for a new language feature could justify it?

lib/src/argument_list_visitor.dart
```diff
@@ -295,10 +294,11 @@ class ArgumentSublist {
   /// The full argument list from the AST.
   final List<Expression> _allArguments;
 
-  /// The positional arguments, in order.
+  /// The leading positional arguments, in order.
   final List<Expression> _positional;
 
-  /// The named arguments, in order.
+  /// The named arguments, in order, and any positional arguments that follow
+  /// at least one named argument.
```
natebosch, line 301: These comments don't match my understanding of `_splitArgumentLists`. IIUC This list can contain positional arguments that don't follow any named argument, as long as _some other_ positional argument follows some named argument.

natebosch (approved):

CHANGELOG.md
munificent: I'm usually pretty conservative. This isn't a breaking change and doesn't really add any new "features" (in terms of the CLI arguments or library API). Also, it doesn't change the formatting of any existing code. That sounds like a point release to me, even though it's significant in that it adds support for new syntax. But since no code using that syntax exists in the wild, from the user's perspective, there's no visible change in dart_style at all.

lib/src/argument_list_visitor.dart
munificent: Good catch. Reworded.

lib/src/argument_list_visitor.dart
```diff
@@ -294,11 +294,15 @@ class ArgumentSublist {
...
-  /// The leading positional arguments, in order.
+  /// If all positional arguments occur before all named arguments, then this
+  /// contains the positional arguments, in order. Otherwise (there are no
+  /// positional arguments or they are interleaved with named ones), this is
+  /// empty.
   final List<Expression> _positional;
 
-  /// The named arguments, in order, and any positional arguments that follow
-  /// at least one named argument.
+  /// The named arguments, in order. If there are any named arguments that occur
```
natebosch, line 303: [nit] https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

PR #1097, opened 2022-03-03
Format "super." parameters.

### LINKED ISSUE #1091: dart_style implementation for super-initializer parameters
See https://github.com/dart-lang/sdk/issues/48055 for the main tracking issue.

munificent:
Fix #1091.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -802,7 +802,7 @@ class SourceVisitor extends ThrowingAstVisitor {
       token(node.equals);
       space();
 
-      visit(node.superclass2);
+      visit(node.superclass);
 
       builder.startRule(CombinatorRule());
       visit(node.withClause);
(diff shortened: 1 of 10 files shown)


lib/src/source_visitor.dart
```diff
@@ -2528,6 +2528,21 @@ class SourceVisitor extends ThrowingAstVisitor {
     token(node.superKeyword);
   }
 
+  @override
+  void visitSuperFormalParameter(SuperFormalParameter node) {
```
munificent, line 2532: This is virtually identical to the code for `visitFieldFormalParameter()` (`this.` parameters), which is to be expected. Unfortunately, I couldn't easily hoist it out to something reusable because FieldFormalParameter and SuperFormalParameter don't define their fields on any shared base class.

natebosch (approved):

PR #1111, opened 2022-05-26
Move a bunch of functions on AST nodes over to extensions.

munificent:
These were mostly arbitrarily defined in SourceVisitor and that class
is huge enough as it is, so this moves them out.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -1,0 +1,228 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+
+extension AstNodeExtensions on AstNode {
+  /// The comma token immediately following this if there is one, or `null`.
+  Token? get commaAfter {
+    var next = endToken.next!;
+    if (next.type == TokenType.COMMA) return next;
+
+    // TODO(sdk#38990): endToken doesn't include the "?" on a nullable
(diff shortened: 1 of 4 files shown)


munificent:
Thank you for reviewing that giant refactor! This one should be much easier. :)

natebosch:
> Thank you for reviewing that giant refactor! Your reward is more reviews ;)

lib/src/ast_extensions.dart
```diff
@@ -0,0 +1,228 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+
+extension AstNodeExtensions on AstNode {
+  /// The comma token immediately following this if there is one, or `null`.
+  Token? get commaAfter {
+    var next = endToken.next!;
+    if (next.type == TokenType.COMMA) return next;
+
+    // TODO(sdk#38990): endToken doesn't include the "?" on a nullable
```
natebosch, line 13: This issue is closed

https://github.com/dart-lang/sdk/issues/38990

Can this TODO be addressed?

lib/src/ast_extensions.dart
```diff
@@ -0,0 +1,228 @@
...
+    // function-typed formal, so check for that case and handle it.
+    if (next.type == TokenType.QUESTION && next.next!.type == TokenType.COMMA) {
+      return next.next;
+    }
+
+    return null;
+  }
+
+  // TODO: Define this on Iterable<AstNode> too?
```
natebosch, line 22: This looks TODONE on line 63...

natebosch (approved):

lib/src/ast_extensions.dart
munificent: Oops, yes!

PR #1112, opened 2022-05-27
Remove preserveWhitespace() and related code.

munificent:
There are some places where a newline or blank line in the original
source code affects the output formatting. For example, users can put a
blank line between statements if they want.

The formatter used to track that lazily like it did other split
information. It used a Whitespace enum that could be in a fully pinned
down or still partially unknown state. Then when the subsequent comment
or text was written, it would count the newlines and figure out how to
split.

Since there is already a lot of laziness and pending state in
ChunkBuilder, that gets pretty complex.

This removes all that. Instead, right at the point in SourceVisitor
where we write whitespace that depends on the input newlines, we look
ahead and eagerly decide how to handle it.


lib/src/chunk.dart
```diff
@@ -250,58 +250,6 @@ class ChunkBlock {
   ChunkBlock(this.argument);
 }
 
-/// Constants for the cost heuristics used to determine which set of splits is
```
munificent, line 253: I renamed "whitespace.dart" to "constants.dart" and moved this over to that file mainly because it seemed silly to have a file with just one tiny set of constants. Also, having this in "chunk.dart" was always kind of arbitrary.

lib/src/source_visitor.dart
```diff
@@ -3539,33 +3563,57 @@ class SourceVisitor extends ThrowingAstVisitor {
...
   void oneOrTwoNewlines() {
-    builder.writeWhitespace(Whitespace.oneOrTwoNewlines);
+    if (_linesBeforeNextToken > 1) {
+      twoNewlines();
+    } else {
+      newline();
+    }
+  }
+
+  /// Whether there is a blank line between the last written token and the next
```
natebosch, line 3607: [nit] Comment reads as if it were a bool. Maybe "The number of blank lines"

munificent: Oops, good catch. Done.

PR #1116, opened 2022-06-09
Refactor to use enhanced enums for Output and Show.

munificent:
Also some other minor clean-up while I was at it:

- Remove unneeded library tags.
- Fix some raw typed collections.
- Use list element where it makes sense.
- Remove unneeded .whereType<Rule>() calls. (They were to filter out
  null, but the rule field is non-nullable now.)

### DIFF
--- a/lib/src/line_splitting/line_splitter.dart
+++ b/lib/src/line_splitting/line_splitter.dart
@@ -119,11 +119,8 @@ class LineSplitter {
   /// page width.
   LineSplitter(this.writer, this.chunks, this.blockIndentation)
       : // Collect the set of rules that we need to select values for.
-        rules = chunks
-            .map((chunk) => chunk.rule)
-            .whereType<Rule>()
-            .toSet()
-            .toList(growable: false) {
+        rules =
+            chunks.map((chunk) => chunk.rule).toSet().toList(growable: false) {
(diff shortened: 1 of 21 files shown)


lib/src/line_splitting/line_splitter.dart
```diff
@@ -119,11 +119,8 @@ class LineSplitter {
   /// page width.
   LineSplitter(this.writer, this.chunks, this.blockIndentation)
       : // Collect the set of rules that we need to select values for.
-        rules = chunks
-            .map((chunk) => chunk.rule)
-            .whereType<Rule>()
-            .toSet()
-            .toList(growable: false) {
+        rules =
+            chunks.map((chunk) => chunk.rule).toSet().toList(growable: false) {
```
natebosch, line 123:
```suggestion
            {for (var chunk in chunks) chunk.rule}.toList(growable: false) {
```

natebosch (approved):

lib/src/line_splitting/line_splitter.dart
munificent: It's only three characters shorter and I think the current code fits a little nicer into a method chain inside a constructor initializer, so I'll leave it as-is.

PR #1122, opened 2022-07-12
Use a more general mechanism for tracking constraints between rules.

munificent:
Rules were initially decoupled from each other. Then I added _implies
to handle subexpressions forcing outer expressions to split. Then I
added some object-oriented "constrain()" and "addConstrainedRules()"
API so that Rules could dynamically add other ad-hoc constraints.

At this point, it's better to just have a single general mechanism for
expressing constraints between rules that can subsume both of those.
This does that. I think the resulting code is simpler. It is also easier
to extend for other kinds of constraints in the future.


lib/src/rule/rule.dart
```diff
@@ -22,6 +22,12 @@ class Rule extends FastHash {
   /// It disallows [unsplit] but allows any other value.
   static const mustSplit = -1;
 
+  /// Rule constraint that means the rule must split to its fully split value.
+  /// This is used instead of the actual full split value because at the point
```
natebosch, line 26: [nit] blank line following header

natebosch (approved):

lib/src/rule/rule.dart
munificent: Done.

PR #1132, opened 2022-08-29
Add flutter tool cli flag

Jasguerrero:
When `dart format --fix` is called inside the flutter tool it gets stuck because it waits on the stdin a work around is to add a flag to the command that tells it is invoked from the flutter tool. Let me know if this is reasonable. 
Flutter issue: https://github.com/flutter/flutter/issues/109553

### DIFF
--- a/lib/src/cli/format_command.dart
+++ b/lib/src/cli/format_command.dart
@@ -122,9 +122,9 @@ class FormatCommand extends Command<int> {
     var followLinks = argResults['follow-links'];
     var setExitIfChanged = argResults['set-exit-if-changed'] as bool;
 
-    // If stdin isn't connected to a pipe, then the user is not passing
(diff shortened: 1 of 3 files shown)


Jasguerrero:
Additionally I can change the name of the new flag to something more generic like `no-stdin`

christopherfujino:
cc @munificent can you take a look?

munificent:
Sorry for the delay. I've been swamped by design work for records and pattern matching.

I definitely don't think dart_style should have any CLI flags that mention Flutter by name. This flag doesn't have anything to do with Flutter. It just has to do with being spawned as a process, thinking it will receive input from stdin, and then not reporting an error because the intention is not to send it any input that way.

I'm not an expert on processes and pipes, but is it possible for Flutter to spawn the dart_style process without opening stdin to it? It feels a little weird to add a command line flag to communicate something that should be communicable directly using the primitives the OS gives us.

PR #1133, opened 2022-09-04
Better grammar

asashour:
I guess it is better to say `which was formatted to`, instead of `which formatted to`.


munificent:
I agree what you have is better. But it's only a small improvement, and I'm worried that changing the wording of this output might break user scripts that are looking for this message, so I'm just going to leave it as is.

PR #1140, opened 2022-09-26
Bump to analyzer 5.1.0 and handle deprecations

srawlins:
Since one of the deprecations is moving `.name2` _back_ to `.name`, I think it's necessary to bump the lower bound. In analyzer 4.4.0, `.name` may have meant a different thing...

In any case, analyzer 5.1.0 is the first release to allow unnamed libraries, so I think it is necessary for the follow up PR which adds such support in dart_style.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -598,7 +598,7 @@ class SourceVisitor extends ThrowingAstVisitor {
     modifier(node.abstractKeyword);
     token(node.classKeyword);
     space();
-    token(node.name2);
+    token(node.name);
     visit(node.typeParameters);
     visit(node.extendsClause);
     _visitClauses(node.withClause, node.implementsClause);
@@ -617,7 +617,7 @@ class SourceVisitor extends ThrowingAstVisitor {
       modifier(node.abstractKeyword);
       token(node.typedefKeyword);
       space();
-      token(node.name2);
+      token(node.name);
       visit(node.typeParameters);
       space();
       token(node.equals);
@@ -764,7 +764,7 @@ class SourceVisitor extends ThrowingAstVisitor {
     modifier(node.factoryKeyword);
     visit(node.returnType);
     token(node.period);
-    token(node.name2);
+    token(node.name);
 
     // Make the rule for the ":" span both the preceding parameter list and
     // the entire initialization list. This ensures that we split before the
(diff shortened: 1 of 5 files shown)


srawlins:
I think test and analyzer 5.0.0 have some pub version solve issue, so that effectively, analyzer 5.0.0 paired with package:test are only supported on Dart 2.18.0. Should I change the GitHub actions to handle this? (As it is, I guess we're now testing 2.17.0 and 2.19.0-dev.)

srawlins:
I've merged in a bump to the GitHub actions, combining #1142 with this. Good talk.

srawlins:
This is a preliminary step to #1141 

PTAL @munificent @natebosch

lib/src/source_visitor.dart
```diff
@@ -2022,7 +2022,7 @@ class SourceVisitor extends ThrowingAstVisitor {
     _simpleStatement(node, () {
       token(node.libraryKeyword);
       space();
-      visit(node.name);
+      visit(node.name2);
```
natebosch, line 2025: what's happening with this one?

srawlins: In analyzer 5.1.0, `.name` is deprecated in favor of `.name2`.

natebosch: Is this on a different type than the rest of the invocations?

srawlins: Oops, my bad. In all of the _other_ changes, `.name2` is deprecated in favor of `.name`.

For this change, `LibraryDirective.name`, a non-nullable String is being replaced with `LibraryDirective.name2`, a nullable String (for the case of an omitted named).

pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
   sdk: ">=2.17.0 <3.0.0"
 
 dependencies:
-  analyzer: '>=4.4.0 <6.0.0'
```
natebosch, line 12: I can't leave a comment on the right line - but we need to bump the version of `dart_style`

natebosch (commented):
I looked up the relevant APIs... It feels weird that on some nodes we're flipping one way, and on other nodes the opposite way... 🤷

pubspec.yaml
srawlins: Done

srawlins:
Yeah that is definitely a weird situation in this release. Accidental on my and Konstantin's part.

munificent (approved):

Issue #1115, opened 2022-06-02
remove the configurable line_length feature


NANASHI0X74:
There's been much discussion about `dart format`'s --line-length parameter and the anguish and excess work that the lack of configurability of this parameter on a per-package basis via configuration files is causing in this feature request: #918

As it's apparent that this feature is not something the dart/dart_style team is interested in adding, may I propose going in the other direction and completely removing the ability to configure the line length?

I would personally not be completely happy with that, but the current situation is just a mess (which I do hope you recognize as well) and it needs to be addressed one way or another.

There's this proverb in go: [Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=8m43s) Which I think is pretty astute and instructive in this case. I think there's huge value for developers in having a consistent code format/style enforced and I think that's more valuable than giving developers the freedom to choose their own formatting rules- at least to an extent. I personally don't believe it's a good thing to enforce exacly the same style across the entire ecosystem of a language, I'd say that there are some parameters that it's better to just leave up to project maintainers to choose. But not having those parameters to configure is definitely better than to have them, but then not having any sensible way to enforce them apart from "out of band" methods like using commit hooks or sharing the configuration of the developers' editors (which doesn't work unless everyone uses the same editor, which opens a whole new fun can of worms 🙄)

lukepighetti:
I personally think we should either get rid of the ability to configure line length altogether (as you suggest) or get the green light to implement a pubspec.yaml line length setting. What's not going to work for me is living this in the middle where the only thing it provides is pain.

munificent:
Given how much anger already gets sent my way around line length, I can't imagine taking away control would do anything except make the people who care about this issue even angrier than they already are. And the people who aren't unhappy about this don't use the command line argument anyway, so having it there isn't harmful.

(Also, pragmatically, the formatter needs to support custom line lengths internally because its tests rely on that. So removing the configurability completely wouldn't let me simplify the code any.)

NANASHI0X74:
So I guess that means you will add the project-level line length configuration option then, right?

JakeSays:
Given that personal expression is an anti-goal of the dart project I do not see why this option is even available.

I was thinking of switching from vs code to intellij for a dart project of mine. I open the code, start typing, press enter, and *BAM* my existing code is re-formatted because I violated the style rules of the almighty dart gods.

bsutton:
@munificent 
[...]

PR #1147, opened 2022-10-11
Hide the --line-length option

natebosch:
Towards #1115

This is option is not intended to be user facing. It is not ergonomic, and it causes confusion in the current state. The option is not removed because it is used for tests, and we don't want to break scripts which are passing the option today.

### DIFF
--- a/lib/src/cli/options.dart
+++ b/lib/src/cli/options.dart
@@ -85,7 +85,10 @@ void defineOptions(ArgParser parser,
   if (verbose) parser.addSeparator('Other options:');
 
   parser.addOption('line-length',
-      abbr: 'l', help: 'Wrap lines longer than this.', defaultsTo: '80');
+      abbr: 'l',
+      help: 'Wrap lines longer than this.',
+      defaultsTo: '80',
+      hide: true);
   parser.addOption('indent',
       abbr: 'i',
       help: 'Add this many spaces of leading indentation.',


natebosch:
I don't see anywhere in the SDK this option is redefined, so I assume that `dart format` is picking up args from here.

munificent:
I don't think the option causes any real confusion. It does exactly what it says it does. I don't think hiding the option will satisfy any of the users that are unsatisfied with not supporting specifying line length in some configuration file somewhere.

This PR doesn't feel to me like it solves any problems.

natebosch:
> I don't think the option causes any real confusion.

See #1115 

> It does exactly what it says it does.

That's not an argument against hiding it. Most hidden options in any CLI do what they say they do. They are hidden because they are not intended for general usage.

munificent (approved):
OK, SGTM.

DanTup:
FWIW, this value is exposed in VS Code as a setting (which calls the formatting API via the analysis server). I don't have any numbers on how many users use it, but I do sometimes use it myself in my own projects (I code on an ultrawide screen and find 80 characters far too narrow to force wrapping even with two windows side-by-side and the sidebar visible).

Someone asked on Discord about how to use this setting from the CLI which they were using in VS Code (I guess that may be useful to know if you want to set up CI to verify formatting). I was confused to find this setting not listed in `dart format --help -v` until I found this PR.

If the plan is to remove this functionality, we should also deprecate and remove it from the editors/server too. I'm not sure I understand the reasoning for doing so though. The referenced PR seems to suggest if it won't be supported at a project level it should be removed entirely, and this doesn't feel sound to me. In VS Code you can set this at a workspace folder level which I feel like _almost_ covers project-level (or what most people would want - a setting that applies to a whole workspace, or a particular repo on disk).

natebosch:
> If the plan is to remove this functionality, we should also deprecate and remove it from the editors/server too

No plans to remove it entirely - it's useful within this repo for testing, and it would be more trouble than it's worth to remove existing usage across the ecosystem.

PR #1158, opened 2023-01-10
Format records and record type annotations.

munificent:
In the process of working on this, I discovered that multi-line function types with trailing commas are pretty busted looking. (I guess no one puts trailing commas there.) I fixed those too, since they are structurally similar to record types in many ways.

(If you don't have time to review this, @natebosch, feel free to bounce it to someone else.)

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -100,8 +100,8 @@ extension ExpressionExtensions on Expression {
   ///     ]..addAll(numbers);
   bool get isCollectionLike {
     var expression = this;
-    if (expression is ListLiteral) return false;
-    if (expression is SetOrMapLiteral) return false;
+    if (expression is ListLiteral) return true;
+    if (expression is SetOrMapLiteral) return true;
 
     // If the target is a call with a trailing comma in the argument list,
     // treat it like a collection literal.
@@ -119,7 +119,7 @@ extension ExpressionExtensions on Expression {
     //       element
     //     ])..cascade();
 
-    return arguments == null || !arguments.arguments.hasCommaAfter;
+    return arguments != null && arguments.arguments.hasCommaAfter;
   }
 
   /// Whether this is an argument in an argument list with a trailing comma.
--- a/lib/src/line_writer.dart
+++ b/lib/src/line_writer.dart
@@ -180,10 +180,11 @@ class LineWriter {
   void _writeChunksUnsplit(BlockChunk block) {
     for (var chunk in block.children) {
       if (chunk.spaceWhenUnsplit) _buffer.write(' ');
-      _writeChunk(chunk);
 
       // Recurse into the block.
       if (chunk is BlockChunk) _writeChunksUnsplit(chunk);
+
+      _writeChunk(chunk);
     }
   }
 
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -3127,9 +3127,7 @@ class SourceVisitor extends ThrowingAstVisitor {
     // Unlike other collections, records don't force outer ones to split.
     if (node is! RecordLiteral) {
       // Force all of the surrounding collections to split.
-      for (var i = 0; i < _collectionSplits.length; i++) {
-        _collectionSplits[i] = true;
-      }
+      _collectionSplits.fillRange(0, _collectionSplits.length, true);
 
       // Add this collection to the stack.
       _collectionSplits.add(false);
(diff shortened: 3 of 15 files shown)


lib/src/ast_extensions.dart
```diff
@@ -100,8 +100,8 @@ extension ExpressionExtensions on Expression {
   ///     ]..addAll(numbers);
   bool get isCollectionLike {
```
munificent, line 101: This method returned the opposite of what its name says. Negated everything in the body and then negated the one callsite so that the name is now accurate.

lib/src/line_writer.dart
```diff
@@ -180,10 +180,11 @@ class LineWriter {
   void _writeChunksUnsplit(BlockChunk block) {
     for (var chunk in block.children) {
       if (chunk.spaceWhenUnsplit) _buffer.write(' ');
-      _writeChunk(chunk);
 
       // Recurse into the block.
       if (chunk is BlockChunk) _writeChunksUnsplit(chunk);
+
+      _writeChunk(chunk);
```
munificent, line 187: This was an unrelated bug. But I think it never surfaced because you couldn't get into a state where you had an inner block that split inside a block that did not.

lib/src/source_visitor.dart
```diff
@@ -2295,6 +2300,110 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+  void visitRecordLiteral(RecordLiteral node) {
+    modifier(node.constKeyword);
+    _visitCollectionLiteral(
+        node, node.leftParenthesis, node.fields, node.rightParenthesis);
+  }
+
+  @override
+  void visitRecordTypeAnnotation(RecordTypeAnnotation node) {
+    var namedFields = node.namedFields;
+
```
munificent, line 2313: This is largely copied from _visitTrailingCommaParameterList(), but there are enough differences that it's hard to share code between them.

lib/src/source_visitor.dart
```diff
@@ -2777,18 +2885,30 @@ class SourceVisitor extends ThrowingAstVisitor {
...
-      // Write the trailing comma.
+      // Write the comma separator.
       if (node != nodes.last) {
-        token(node.endToken.next);
+        var comma = node.endToken.next;
+
+        // TODO(rnystrom): There is a bug in analyzer where the end token of a
+        // nullable record type is the ")" and not the "?". This works around
+        // that. Remove that's fixed.
+        if (comma?.lexeme == '?') comma = comma?.next;
```
munificent, line 2902: I have a fix out for this already: https://dart-review.googlesource.com/c/sdk/+/278696

But the workaround it harmless and this way I can land before waiting for the analyzer folks to kick out a release.

munificent:
The basic idea with the style is that record expressions and type annotations are mostly formatted like collection literals (lists and maps). That means they either put all the fields on one line, or the whole thing splits and each field gets its own line with +2 indentation, with the closing delimiter (`)` in this case) on the next line.

The differences are:

* Collection literals have a special rule where a nested collection forces surrounding ones to split. If you do `var list = [[]];`, it will split the outer one even though it's otherwise not necessary. I didn't do this rule for records because I think nested records might be useful for packing in data and are reasonable to fit on one line.

* Collections always split if there's a trailing comma. Records do too *except* if there is only a single positional element. There, the trailing comma is required for disambiguation. So the formatter doesn't interpret that as a request to split the record. Otherwise, there would be no way to have a single-line single-positional-element record.

* Collections inside argument lists get some special handling where they may produce "block-like" formatting, as in:

    ```dart
    blockyList([
      1,
      2,
      3,
    ]);
    ```

    Records don't have that special handling. Partially it's because this special formatting doesn't always do a great job (it gets complex when there are multiple arguments with different shapes). More importantly, records are more like fixed-sized values than unbounded collections, so I don't think having a formatting style that scales to large records is useful. Since there is no reflective API for records—no way to iterate over their fields—the size of records is practically limited.

I think that's about it.

lib/src/source_visitor.dart
```diff
@@ -3004,16 +3124,19 @@ class SourceVisitor extends ThrowingAstVisitor {
...
-    // Force all of the surrounding collections to split.
-    for (var i = 0; i < _collectionSplits.length; i++) {
-      _collectionSplits[i] = true;
-    }
+    // Unlike other collections, records don't force outer ones to split.
+    if (node is! RecordLiteral) {
+      // Force all of the surrounding collections to split.
+      for (var i = 0; i < _collectionSplits.length; i++) {
+        _collectionSplits[i] = true;
+      }
```
natebosch, line 3132: [optional]
```suggestion
      _collectionSplits.fillRange(0, _collectionSplits.length, true);
```

natebosch (approved):

munificent:
Thanks for the review!

PR #1171, opened 2023-02-09
Format object patterns, const constructors, and constant expressions.

munificent:
The latter two were already implemented as part of keeping the existing switch tests passing, but this adds tests for their splitting and other behavior.

For object patterns, I had to decide whether to split them like collections (like other patterns) or like argument lists (which they mirror). I went collection-like because they will appear mixed with other patterns. Also, if we migrate to a consistent flutter-like style at some point, it will be one less thing to change.

Let me know what you think.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -2290,6 +2290,38 @@ class SourceVisitor extends ThrowingAstVisitor {
   }
 
   @override
+  void visitObjectPattern(ObjectPattern node) {
+    // Even though object patterns syntactically resemble constructor or
+    // function calls, we format them like collections (or like argument lists
+    // with trailing commas). In other words, like this:
+    //
+    //     case Foo(
+    //         first: 1,
+    //         second: 2,
+    //         third: 3
+    //       ):
+    //       body;
+    //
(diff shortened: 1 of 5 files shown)


lib/src/source_visitor.dart
```diff
@@ -2289,6 +2289,38 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+    //     case Foo(
+    //           first: 1,
+    //           second: 2,
+    //           third: 3):
+    //       body;
+    //
+    // This is less consistent with the corresponding expression form, but is
+    // more consistent with all of the other delimited patterns -- list, map,
+    // and record -- which have collection-like formatting.
+    // TODO(rnystrom): If we move to consistently using collection-like
```
natebosch, line 2316: [optional] This is a place where linking the TODO to an issue instead of author will greatly increase the likelihood of it getting noticed along with other changes for that issue.

munificent: Yeah, that's a good idea. Though I'm hesitant to file an issue for this because of the volume of discussion it will generate that I don't have time to moderate right now.

PR #1174, opened 2023-02-12
Format switch expressions, guards, and patterns in switches.

munificent:
This also changes how regular cases in switch statements are formatted. Previously, a case's body was always moved to the next line:

    switch (obj) {
      case 1:
        body;
      case 2:
        body;
    }

I honestly never really loved that style, but it worked OK. With switch expressions, it obviously makes sense to keep the expression inline if it fits:

    switch (obj) {
      1 => body,
      2 => body
    }

I think people will generally prefer that style. I don't want them to use a switch expression where all they need is a switch statement just because it fits tighter, so I also allow switch statement case bodies to go next to the case if they fit now:

    switch (obj) {
      case 1: body;
      case 2: body;
    }

I think this will be nice, especially since switch statements will likely become more common.

Guards appear in three place: switch statement cases, switch expression cases, and if-case statements. There is little code reuse for how they are formatted in those three contexts. This is unfortunate, but deliberate. The indentation rules for guards are slightly different in all of those places because the surrounding context is different. I tried to make each context look as best as I could, which leads to unique formatting in each.


natebosch:
> I think this will be nice

Have we checked impact on existing code? How many lines do we expect to churn?

natebosch (approved):

munificent:
> Have we checked impact on existing code?

I haven't yet, but I'll run it on a big corpus and look at the diffs. If it seems big enough, I can go through the official change process. But I strongly suspect that we will want to make this change. Almost all of the pattern matching examples using `switch` have each case body on the same line as the case.

natebosch:
> I strongly suspect that we will want to make this change.

Yeah, I'll be curious to see the diff if it goes against my expectations, but I'd think this should be a positive change overall and worth making.

PR #1176, opened 2023-02-14
Better style for inline case bodies.

munificent:
In the previous PR, any case body that fit on one line was allowed to even if other cases in the same switch didn't. I tested it on a corpus and I found that led to confusing switches where it wasn't always clear where the case body starts.

I think you really want it all or nothing: either every single case fits on the same line in which case you can make the whole switch compact, or every case should be on its own line, even the ones that would fit.

Unfortunately, it's a little tricky to have formatter rules that span code containing hard splits, so getting that working took some doing. It also regressed performance pretty badly. But I figured out some optimizations in ChunkBuilder and it's basically back to the same performance it had before.

Also, this incidentally fixes a bug where parameter metadata in trailing comma parameter lists was also supposed to have that same all-or-nothing splitting logic but didn't.

I've tried this on a corpus and I'm pretty happy with the results. Right now, relatively few switches benefit because the mandatory breaks mean a lot of switches have at least two statements (which always causes the case to split). But as those breaks are removed, I think we'll see more compact switches. Even today, this code does improve some switches where every case is just a short return statement.

### DIFF
--- a/lib/src/chunk_builder.dart
+++ b/lib/src/chunk_builder.dart
@@ -949,9 +949,10 @@ class ChunkBuilder {
     return true;
   }
 
-  /// For each non-hardened rule, find the range of chunks that share it. We
-  /// must ensure that those chunks are all split together so that they don't
-  /// try to pick different values for the same rule.
+  /// For each non-hardened rule, find the range of chunks that share it.
+  ///
+  /// We must ensure that those chunks are all split together so that they
+  /// don't try to pick different values for the same rule.
   ///
   /// Returns a list of `[low, high]` ranges.
   ///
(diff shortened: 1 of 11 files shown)


lib/src/chunk_builder.dart
```diff
@@ -891,38 +891,116 @@ class ChunkBuilder {
...
     // and its children from the preceding beginning of the block.
-    if (_chunks[i] is BlockChunk) return false;
+    if (chunk is BlockChunk) return false;
 
     return true;
   }
 
-  /// Pre-processes the chunks after they are done being written by the visitor
-  /// but before they are run through the line splitter.
+  /// For each non-hardened rule, find the range of chunks that share it. We
```
natebosch, line 952: [nit] blank line following header

natebosch (approved):

PR #1177, opened 2023-02-14
Format switch cases that aren't valid patterns.

### LINKED ISSUE #1164: How to handle breaking changes to switch cases?
The upcoming patterns support reinterprets all existing switch case constant expressions as patterns. Most constant expressions seen in switches are also valid patterns, so that change is mostly harmless. However, there are [some constant expressions that are not valid patterns][breaking]. This is perfectly fine Dart 2.19 code:

[breaking]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#breaking-existing-switches

```dart
switch (obj) {
  case 1 + 2: print('three'); break;
}
```

But the `1 + 2` is a syntax error in Dart 3.0. This is mostly not a problem for users. Cases like this are rare and users can fix their code when they upgrade their package or library's language version to 3.0.

However, this *is* a real problem for the formatter. The formatter is not currently aware of the language version of any file that it processes. When it parses a file, it just unconditionally tells analyzer to treat the file as being [at the latest language version with all experiments enabled](https://github.com/dart-lang/dart_style/blob/master/lib/src/dart_formatter.dart#L91). This is a little hacky, but has worked fine up until now since language changes were always syntactic additions. The latest version of the language was (syntactically) a superset of all previous ones.

With patterns, that's no longer the case. I'm filing this issue to draw attention to that and start talking about a solution. I can think of two so far:

1.  Make dart_style aware of the language version of each file it parses. This is consistent with other tools which generally know the language version. It future-proofs us from running into this problem again in future versions.

    Users have long asked for the ability to configure formatter options in their pubspec. One reason we haven't done that is that dart_style doesn't even know what a pubspec *is*. Making dart_style read the pubspec to ascertain the language version would also put the infrastructure in place to enable this.

    On the other hand, this adds complexity to how the formatter works. It will have to search surrounding directories looking for pubspecs. It will potentially have to look for language version markers in every file before parsing.

    We'll have to figure out how to handle the ways the formatter can be run where that context isn't available. Right now, you can format strings through the library API. You can also pass code to it directly through stdin. You can format files that aren't part of a pub package. I worry about brittleness.

2.  Extend the parser to parse switch cases that are now invalid. Instead of only parsing cases as patterns when a library is at the latest version, parse to a cover grammar that is the union of the pattern grammar plus the old on-longer-supported expression forms. The latter would report syntax errors, but still be parsed correctly and yield a complete AST. The formatter would then ignore those particular errors and work with the AST.

    This makes the formatter's (and my) life easier. But it adds complexity to the parser and AST. I think most of the AST complexity is already done. `SwitchStatement` contains a list of `SwitchMember`s and there are subclasses of `SwitchMember` for `SwitchCase` (old version with an `Expression` for the case) and `SwitchPatternCase` (new one).

    I don't know how much work it would be in the parser to be able to parse a switch statement containing a mixture of pattern and constant expression cases. It may be that the overall user experience and error recovery is better if we do this. Right now, if I enable patterns and parse:

    ```dart
    switch (c) {
      case 1 + 2: oops;
    }
    ```

    I get:

    ```
    line 2, column 12 of .: Expected to find ':'.
      ╷
    2 │     case 1 + 2: oops;
      │            ^
      ╵
    line 2, column 12 of .: Expected an identifier.
      ╷
    2 │     case 1 + 2: oops;
      │            ^
      ╵
    line 2, column 14 of .: Expected to find ';'.
      ╷
    2 │     case 1 + 2: oops;
      │              ^
      ╵
    line 2, column 15 of .: Expected an identifier.
      ╷
    2 │     case 1 + 2: oops;
      │               ^
      ╵
    line 2, column 15 of .: Unexpected text ':'.
      ╷
    2 │     case 1 + 2: oops;
      │               ^
      ╵
    ```

    It would be better (especially for users migrating their code to Dart 3.0) if the error was more like:

    ```
    line 2, column 10 of .: This constant is expression is not a valid pattern in Dart 3.0.
      ╷
    2 │     case 1 + 2: oops;
      │          ^^^^^
      ╵
    ```

Thoughts @leafpetersen @mit-mit @bwilkerson @jensjoha?

munificent:
Fix #1164.

The solution is kind of hacky, but users will probably never run into it and it avoids complicated the user experience of the formatter.

To get this working, I had to update to analyzer 5.5.0 because 5.4.0 had an assert failure when it tried to parse an invalid switch case. But 5.5.0 also has a bug which is causing a couple of formatter tests to fail: https://github.com/dart-lang/sdk/issues/51415.

I'll probably wait until there's a fix for that out before this gets merged to master.

Analyzer 5.5.0 also changes some of the AST types. Refactored how binary expressions and patterns are formatted to avoid copy/paste from that change.

### DIFF
--- a/lib/src/dart_formatter.dart
+++ b/lib/src/dart_formatter.dart
@@ -84,10 +84,11 @@ class DartFormatter {
   /// Returns a new [SourceCode] containing the formatted code and the resulting
   /// selection, if any.
   SourceCode formatSource(SourceCode source) {
-    // If we're parsing a single statement, wrap the source in a fake function.
     var inputOffset = 0;
     var text = source.text;
     var unitSourceCode = source;
+
+    // If we're parsing a single statement, wrap the source in a fake function.
     if (!source.isCompilationUnit) {
       var prefix = 'void foo() { ';
       inputOffset = prefix.length;
--- a/lib/src/dart_formatter.dart
+++ b/lib/src/dart_formatter.dart
@@ -4,6 +4,7 @@
 import 'dart:math' as math;
 
 import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/analysis/utilities.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
@@ -83,22 +84,7 @@ class DartFormatter {
   /// Returns a new [SourceCode] containing the formatted code and the resulting
   /// selection, if any.
   SourceCode formatSource(SourceCode source) {
-    // Enable all features that are enabled by default in the current analyzer
-    // version.
-    // TODO(paulberry): consider plumbing in experiment enable flags from the
-    // command line.
-    var featureSet = FeatureSet.fromEnableFlags2(
-      sdkLanguageVersion: Version(2, 19, 0),
-      flags: [
-        // TODO(rnystrom): This breaks existing switch cases containing constant
-        // expressions that aren't valid patterns. See:
-        // https://github.com/dart-lang/dart_style/issues/1164
-        'patterns',
-        'records',
-        'unnamed-libraries',
-      ],
-    );
-
+    // If we're parsing a single statement, wrap the source in a fake function.
     var inputOffset = 0;
     var text = source.text;
     var unitSourceCode = source;
@@ -118,12 +104,19 @@ class DartFormatter {
     }
 
     // Parse it.
-    var parseResult = parseString(
-      content: text,
-      featureSet: featureSet,
-      path: source.uri,
-      throwIfDiagnostics: false,
-    );
+    var parseResult = _parse(text, source.uri, patterns: true);
+
+    // If we couldn't parse it with patterns enabled, it may be because of
+    // one of the breaking syntax changes to switch cases. Try parsing it
+    // again without patterns.
+    if (parseResult.errors.isNotEmpty) {
+      var withoutPatternsResult = _parse(text, source.uri, patterns: false);
+
+      // If we succeeded this time, use this parse instead.
+      if (withoutPatternsResult.errors.isEmpty) {
+        parseResult = withoutPatternsResult;
+      }
+    }
 
     // Infer the line ending if not given one. Do it here since now we know
     // where the lines start.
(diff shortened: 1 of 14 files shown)


lib/src/dart_formatter.dart
```diff
@@ -83,22 +84,7 @@ class DartFormatter {
...
-        // TODO(rnystrom): This breaks existing switch cases containing constant
-        // expressions that aren't valid patterns. See:
-        // https://github.com/dart-lang/dart_style/issues/1164
-        'patterns',
-        'records',
-        'unnamed-libraries',
-      ],
-    );
-
+    // If we're parsing a single statement, wrap the source in a fake function.
```
natebosch, line 87: [nit] consider moving this comment to immediately above or below the `if` statement it relates to.

lib/src/source_visitor.dart
```diff
@@ -3205,6 +3160,65 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+  /// Visits an infix operator-like AST node: a binary operator expression, or
+  /// binary pattern.
+  ///
+  /// Applies the same formatting to all. In particular, it flattens out a
+  /// series of the same operator precedence so that if one splits, they all do.
+  ///
+  /// [T] is the type of node being visited and [destructureNode] is a callback
+  /// that takes one of those and yields the operands and operator. We need
+  /// this since there's no interface shared by the various binary operator
+  /// AST nodes.
```
natebosch, line 3172: I'm trying to understand the implications of the last sentence. Can a `LogicalAndPattern` be an operand in a `LogicalOrPattern` or vice versa?

Why do we _only_ care to treat `LogicalAndPattern` specially  when it is recursively under another `LogicalAndPattern` as the entrypoint?

Can you expand this doc comment to so that it is clear why it is useful to recurse through `traverse` specifically for the T -> T nesting case, while non-T go through `visit`?

natebosch: Am I correct that both `LogicalAndPattern` and `LogicalOrPattern` were previously modeled as the single type `BinaryPattern`? If those _can_ nest I don't think this will behave the same unless each call knows how to handle both of those types.

munificent: > Am I correct that both `LogicalAndPattern` and `LogicalOrPattern` were previously modeled as the single type `BinaryPattern`?

That's correct. But they have different precedences so they behave the same with this change as they did before. A LogicalOrPattern should only traverse and flatten out the other LogicalOrPatterns in its operands. Any LogicAndPatterns (which are higher precedence) should do their own internal flattening. For example:

```dart
if (object
    case first && second ||
        third &&
            fourth &&
            fifth &&
            sixth ||
        seventh) {
  ;
}
```

All of the `||` form a single flattened chain and we split at all of them even though `seventh` could fit on the previous line. Then, within those, each series of `&&` forms its own independent chain. The `first && second` one doesn't split, but the `third ... sixth` one does. I'll see if I can word the doc better.

lib/src/dart_formatter.dart
munificent: Done.

lib/src/source_visitor.dart
natebosch: > Now that they're different AST types, that check isn't needed. (It is still needed for `BinaryExpression`, though.)

Ah ok, that does clear up my confusion. I noticed that the precedence check wasn't applied, it makes sense that it's because we know that the precedence and type are 1:1 for that case.

natebosch (approved):

PR #1182, opened 2023-02-24
Don't crash on switches with empty trailing cases.

### LINKED ISSUE #1181: Crash in patterns branch when switch statement's last case is empty
This will crash the latest code on the "patterns" branch:

```dart
switch (e) {
  case E.e1:
    break;
  case E.e2:
}
```

It has to do with the new compact switch formatting and the unusual behavior where there is a rule for the cases that goes across the hard splits between each case. To handle that correctly, the ChunkBuilder doesn't divide at a hard split if it sees that a rule is used both before and after it. But in this case (heh), since the last case is empty, the rule for splitting the case bodies doesn't actually appear after the hard split before the last `case`.

So it divides the two cases. But the body splitting rule does show up in a *constraint* of one of the other rules. Because of that, it ends up getting an index assigned when it splits the first case and then the LineSplitter gets very confused when it gets to the second case because the assumption is that every rule has a `null` index before line splitting begins.

The core problem is that we shouldn't divide the chunks between the two cases. I'll need to come up with a better way to determine that. I tried traverse the rule constraints and preventing a divide if any rule is used on both sides of a hard split as a chunk's rule or in a constraint, but it's much too slow.

munificent:
In the process, I came up with a cleaner and faster solution for handling rules that span hard splits.

Fix #1181.

### DIFF
--- a/lib/src/chunk.dart
+++ b/lib/src/chunk.dart
@@ -114,10 +114,8 @@ class Chunk extends Selection {
 
   /// Whether this chunk marks the end of a range of chunks that can be line
   /// split independently of the following chunks.
-  ///
-  /// You must call markDivide() before accessing this.
   bool get canDivide => _canDivide;
-  late final bool _canDivide;
+  late bool _canDivide = true;
 
   /// The number of characters in this chunk when unsplit.
   int get length => (_spaceWhenUnsplit ? 1 : 0) + _text.length;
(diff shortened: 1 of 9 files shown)


lib/src/chunk.dart
```diff
@@ -114,10 +114,8 @@ class Chunk extends Selection {
 
   /// Whether this chunk marks the end of a range of chunks that can be line
   /// split independently of the following chunks.
-  ///
-  /// You must call markDivide() before accessing this.
   bool get canDivide => _canDivide;
-  late final bool _canDivide;
+  late bool _canDivide = true;
```
kevmoo, line 118: no longer needs to be `late` right?

munificent: Oops, you're right. Thanks!

kevmoo: There ought'a be a lint for that!

natebosch (approved):

Issue #1190, opened 2023-03-02
Large switches with single-statement bodies get incorrect formatting


munificent:
The change to allow more compact switches can lead to regressed formatting when a switch has a large number of very long single-statement bodies, as in:

Before (identifiers sanitized):

```dart
switch (aaaaaaAaaa) {
  case aaaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
        aaaaaaAaaaaaaaa.aaaaaaAaaaaaa(aaaaaaa?.aaaaaa);
  case aaaaaaaaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
        aaaaaaAaaaaaaaa.aaaaaaAaaaaaa(aaaaaaa?.aaaaaaaaaaa);
  case aaaaaaaaaaAaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
        aaaaaaAaaaaaaaa.aaaaaaAaaaa(aaaaaaa?.aaaaaaaaaaAaaaa, aaaaaaaaAaaa,
            aaaaaaaAaaaaaaaAaaaaa: false);
  case aaaAaaaaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
        aaaaaaAaaaaaaaa.aaaaaaAaaaaaa(aaaaaaa?.aaaAaaaaaaa);
  case aaaaaaaaaaAaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
        aaaaaaAaaaaaaaa.aaaaaaAaaaaaaaaaAaaaaAaaaaa(
            aaaaaaa?.aaaAaaaaaaaaaAaaaaAaaaaa() == true
                ? aaaaaaa!.aaaaaaaaaaAaaaaAaaaaa
                : null);
  case aaaAaaaaaaaaaAaaaaAaaaaaaaAaaaaaAaaa:
    return (Aaaaaaaa? aaaaaaa, Aaaaaaaa? aaaaaaaa) =>
[...]

PR #1191, opened 2023-03-03
Remove support for compact switch statements.

munificent:
I'm still interested in supporting this, but it causes a lot of churn. That means it really needs to go through the full dart style change process. Given how many other things are in the air with Dart 3.0, it seems like now is probably not the ideal time for that.

So, for now, I think we should keep the old formatting where case bodies always go on their own line.

cc @natebosch @iinozemtsev

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,21 @@
 # 2.3.0-dev
 
+## New language features
+
 * Format patterns and related features.
+* Format record expressions and record type annotations.
+* Format class modifiers `base`, `final`, `interface`, `mixin`, and `sealed`.
+* Format unnamed libraries.
+
+## Bug fixes and style changes
+
 * Handle `sync*` and `async*` functions with `=>` bodies.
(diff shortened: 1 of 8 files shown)


munificent:
Also, this fixes #1190.

kevmoo (approved):
Less churn is good. Thanks!

dnfield:
Is there anywhere it's possible to discuss (and vote against) this change? :)

I'm glad to see it got backed out, I was pretty disturbed to see it in my code base, especially since I have `always_put_control_body_on_new_line` enabled.

munificent:
> Is there anywhere it's possible to discuss (and vote against) this change? :)

The formatter isn't a pure democracy, but it is heavily feedback driven. When I look at this change again, I'll go through the actual change process: https://github.com/dart-lang/dart_style/wiki/Change-process and you can register your (dis)-approval through that. (I should have gone through it when I first landed the change, but I underestimated the impact.)

PR #1195, opened 2023-03-16
Hide "--fix" and related options in help output.

### LINKED ISSUE #1153: Remove `--fix`
The `--fix` support in dart_style is older than the `dart fix` command. That command can now apply all of the fixes that `dart format --fix` can and many many more. It's redundant and confusing to have two ways to apply automated code clean-ups. And the implementations of the fixes in dart_style is kind of hacky and brittle. I believe there are cases where the resulting fixed code doesn't end up with the same formatting you'd get if you applied the fix separately.

Dart 3 gives us a chance to clean up our CLI user experience, so it's a great time to remove `--fix` from `dart format` and delete all of that crufty code. We can leave the options in there as hidden. If you pass any, instead of applying the fix, it will print a diagnostic telling you how to run `dart fix`.

munificent:
This doesn't completely fix #1153 because the command is still there and supported, but it's a step towards deprecating it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 2.3.1-dev
+
+* Hide `--fix` and related options in `--help`. The options are still there and
+  supported, but are no longer shown by default. Eventually, we would like all
+  users to move to using `dart fix` instead of `dart format --fix`.
+
 # 2.3.0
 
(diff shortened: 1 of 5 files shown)


mit-mit (approved):
Thanks!

PR #1196, opened 2023-03-20
Remove the mixin modifiers no longer supported in the latest proposal.

munificent:


kallentu (approved):
Ooh, that was fast. Thank you!

PR #1199, opened 2023-03-25
Don't insert space into empty switches.

### LINKED ISSUE #1198: Extra space inserted into empty switch statements and switch expressions
Attempting to format an empty switch expression produces extra space between the curly braces:
```dart
void main() {
  var _ = switch (0) {  };
}
```
Similarly, attempting to format an empty switch statement produces an extra blank line between the curly braces:
```dart
void main() {
  switch (0) {
    
  }
}
```

Given that empty switch expressions and switch statements are now marginally useful as placeholder code (see https://github.com/dart-lang/language/issues/2939), it would be nice for them to be formatted without extra space, similar to how empty statement blocks are formatted.  In other words:
```dart
void main() {
  var _ = switch (0) {};
}
```
and:
```dart
void main() {
  switch (0) {}
}
```

stereotype441:
Fixes #1198.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -2723,6 +2723,15 @@ class SourceVisitor extends ThrowingAstVisitor {
 
   @override
   void visitSwitchExpression(SwitchExpression node) {
+    if (node.cases.isEmpty) {
+      // Don't allow splitting an empty switch expression.
+      _visitSwitchValue(node.switchKeyword, node.leftParenthesis,
+          node.expression, node.rightParenthesis);
+      token(node.leftBracket);
+      token(node.rightBracket);
+      return;
+    }
+
     // Start the rule for splitting between the cases before the value. That
     // way, if the value expression splits, the cases do too. Avoids:
     //
@@ -2739,10 +2748,7 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     visitCommaSeparatedNodes(node.cases, between: split);
 
-    // A switch with no cases isn't syntactically valid, but handle it
-    // gracefully instead of crashing.
-    var hasTrailingComma =
-        node.cases.isNotEmpty && node.cases.last.commaAfter != null;
+    var hasTrailingComma = node.cases.last.commaAfter != null;
     _endBody(node.rightBracket, forceSplit: hasTrailingComma);
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/source_visitor.dart
```diff
@@ -2723,6 +2723,15 @@ class SourceVisitor extends ThrowingAstVisitor {
 
   @override
   void visitSwitchExpression(SwitchExpression node) {
+    if (node.cases.isEmpty) {
+      // Don't allow splitting an empty switch expression.
+      _visitSwitchValue(node.switchKeyword, node.leftParenthesis,
+          node.expression, node.rightParenthesis);
+      token(node.leftBracket);
+      token(node.rightBracket);
```
munificent, line 2731: You'll also want to test how this behaves when the switch is empty but contains comments.

To test/comments/switch.stmt, I would add:

```
>>> line comment
switch (e) {
  // comment
}
<<<
switch (e) {
  // comment
}
>>> line comment on opening line
switch (e) { // comment
}
<<<
switch (e) {
  // comment
}
>>> indented block comment
switch (e) {
  /* comment */
}
<<<
switch (e) {
  /* comment */
}
>>> block comment with trailing newline
switch (e) {/* comment */
}
<<<
switch (e) {
  /* comment */
}
>>> block comment with leading newline
switch (e) {
  /* comment */}
<<<
switch (e) {
  /* comment */
}
>>> inline block comment
switch (e) {  /* comment */  }
<<<
switch (e) {/* comment */}
>>> line comment
e = switch (n) {
  // comment
};
<<<
e = switch (n) {
  // comment
};
>>> line comment on opening line
e = switch (n) { // comment
};
<<<
e = switch (n) {
  // comment
};
>>> indented block comment
e = switch (n) {
  /* comment */
};
<<<
e = switch (n) {
  /* comment */
};
>>> block comment with trailing newline
e = switch (n) {/* comment */
};
<<<
e = switch (n) {
  /* comment */
};
>>> block comment with leading newline
e = switch (n) {
  /* comment */}
<<<
e = switch (n) {
  /* comment */
};
>>> inline block comment
e = switch (n) {  /* comment */  }
<<<
e = switch (n) {/* comment */};
```

You may end up needing to tweak the implementation to get it formatting all of these correctly. (Comments are one of the things that makes formatting hard.)

For comparison, `_visitBody()` (which is used for things like class bodies), has:

```dart
    // Don't allow splitting in an empty body.
    if (nodes.isEmptyBody(rightBracket)) {
      token(leftBracket);
      token(rightBracket);
      return;
    }
```

And `isEmptyBody()` is:

```dart
  bool isEmptyBody(Token rightBracket) =>
      isEmpty && rightBracket.precedingComments == null;
```

That deliberately treats a body containing nothing but a comment as still "non-empty". You may want to do something similar for switches, but then you'll have to make sure to still gracefully handle there being no cases.

stereotype441: Good point, thanks!  I'll revise this PR.

stereotype441: Ok, added the tests, and fixed the implementation so that they pass.  PTAL.

munificent (approved):
Thanks!

Issue #1202, opened 2023-04-04
Indent block-like function expression bodies when split at `=>`


munificent:
A function expression may have a `=>` body followed by a block-like expression, like a list, map, switch expression, or even another function expression. Right now, that inner expression's block indentation is not increased by virtue of being nested inside a function expression. That leads to output like:

```dart
SomeLongFunctionName(
  argument: () => [
    listElement,
    listElement,
    listElement,
  ],
);
```

I think this formatting works pretty well. Most users seem to be OK with it too since lots of code is formatted this way and I haven't heard many complaints. However, the formatting looks weird if we end up needing to split after the `=>` because the parameter list is long or before the block-like expression has something long before its opening bracket, as in:

```dart
SomeLongFunctionName(
  argument: (longParameterName______) =>
      <LongTypeArgument>[
    listElement,
    listElement,
    listElement,
  ],
);
```

This looks pretty strange because now the opening delimiter is on its own line (as opposed to be on the same line as the `=>`) but that line doesn't align with the corresponding closing delimiter. It looks particularly strange with a switch expression (#1197):

```dart
SomeLongFunctionName(
  argument: (longParameterName______) =>
      switch (e) {
    1 => listElement,
    2 => listElement,
    3 => listElement,
  },
);
```

I've tried changing it to always indent the block body, which gives:

```dart
SomeLongFunctionName(
  argument: (longParameterName______) =>
      switch (e) {
        1 => listElement,
        2 => listElement,
        3 => listElement,
      },
);
```

This example now looks better, but the first one is worst:

```dart
SomeLongFunctionName(
  argument: () => [
        listElement,
        listElement,
        listElement,
      ],
);
```

From testing on a corpus of open source code, there are a *lot* of callsites that would be affected by this, and almost all of them ended up looking worse like the last example here.

The correct fix is to have the block indentation *change* based on whether we split after the `=>`. If it splits, then indent the block, otherwise don't.

Unfortunately, the architecture of the formatter doesn't currently support that. Block indentation is calculated when the AST is transformed into chunks before line splitting.

I won't be able to fix that easily, so I'm filing this to track hopefully being able to do that better fix when I have more time.

munificent:
I ran these examples through the [forthcoming](https://github.com/dart-lang/dart_style/issues/1403) [tall style](https://github.com/dart-lang/dart_style/issues/1253) and they look like:

```dart
SomeLongFunctionName(
  argument:
      () => [
        listElement,
        listElement,
[...]

Issue #1197, opened 2023-03-23
Incorrect indentation of trailing comma argument list in switch expression


munificent:
A function call with a trailing argument list in a switch expression currently formats like this:

```dart
main() {
  {
    return TextFieldTapRegion(
      onLongPressMoveUpdate: (longPressMoveUpdateDetails) {
        (switch (Theme.of(this.context).platform) {
          TargetPlatform.iOS || TargetPlatform.macOS =>
            _renderEditable.selectPositionAt(
            from: longPressMoveUpdateDetails.globalPosition,
            cause: SelectionChangedCause.longPress,
          ),
          TargetPlatform.android ||
                TargetPlatform.fuchsia ||
                TargetPlatform.linux ||
                TargetPlatform.windows =>
            _renderEditable.selectWordsInRange(
            from: longPressMoveUpdateDetails.globalPosition -
                longPressMoveUpdateDetails.offsetFromOrigin,
            to: longPressMoveUpdateDetails.globalPosition,
            cause: SelectionChangedCause.longPress,
          )
        });
      },
    );
  }
}
```

That's definitely wrong. It should be:

```dart
main() {
  {
    return TextFieldTapRegion(
      onLongPressMoveUpdate: (longPressMoveUpdateDetails) {
        (switch (Theme.of(this.context).platform) {
          TargetPlatform.iOS || TargetPlatform.macOS =>
            _renderEditable.selectPositionAt(
              from: longPressMoveUpdateDetails.globalPosition,
              cause: SelectionChangedCause.longPress,
            ),
          TargetPlatform.android ||
                TargetPlatform.fuchsia ||
                TargetPlatform.linux ||
                TargetPlatform.windows =>
            _renderEditable.selectWordsInRange(
              from: longPressMoveUpdateDetails.globalPosition -
                  longPressMoveUpdateDetails.offsetFromOrigin,
              to: longPressMoveUpdateDetails.globalPosition,
              cause: SelectionChangedCause.longPress,
            )
        });
      },
    );
  }
}
```

We may also want to consider special-casing `||` as the outer-most pattern in a switch expression case and not indent the subsequent arms, since they're more like parallel cases:

```dart
main() {
  {
    return TextFieldTapRegion(
      onLongPressMoveUpdate: (longPressMoveUpdateDetails) {
        (switch (Theme.of(this.context).platform) {
          TargetPlatform.iOS || TargetPlatform.macOS =>
            _renderEditable.selectPositionAt(
[...]

PR #1203, opened 2023-04-04
Don't indent outermost "||" pattern operands in switch expression cases.

munificent:
This makes them all line up as if they were empty cases in a switch statement, which is essentially how they behave. Instead of:

```dart
e = switch (obj) {
  constant1 ||
     constant2 ||
     constant3 =>
    body
};
```

Gives:

```dart
e = switch (obj) {
  constant1 ||
  constant2 ||
  constant3 =>
    body
};
```

This addresses some of the bad formatting in #1197. The rest is #1202.

cc @mit-mit

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -2795,6 +2795,7 @@ class SourceVisitor extends ThrowingAstVisitor {
 
     // Write the "||" operands up to the last one.
     for (var i = 0; i < orBranches.length - 1; i++) {
+      // Note that orBranches will always have one more element than orTokens.
       visit(orBranches[i]);
       space();
       token(orTokens[i]);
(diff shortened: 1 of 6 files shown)


lib/src/source_visitor.dart
```diff
@@ -2755,25 +2755,71 @@ class SourceVisitor extends ThrowingAstVisitor {
...
     // Wrap the rule for splitting after "=>" around the pattern so that a
     // split in the pattern forces the expression to move to the next line too.
     builder.startLazyRule();
 
-    // Wrap the expression's nesting around the pattern too so that a split in
+    // Write the "||" operands up to the last one.
+    for (var i = 0; i < orBranches.length - 1; i++) {
+      visit(orBranches[i]);
+      space();
+      token(orTokens[i]);
```
kevmoo, line 2800: I guess we trust that `orTokens` is no shorter than `orBranches`?

kevmoo (approved):

lib/src/source_visitor.dart
munificent: Yes, by construction `orBranches` will always have length exactly one longer than `orTokens`. Added a comment.

PR #1207, opened 2023-04-18
Don't split cascades on record literals.

munificent:
This is consistent with how other collection literals preceding single-section cascades are formatted.


lib/src/ast_extensions.dart
```diff
@@ -88,40 +88,6 @@ extension AstIterableExtensions on Iterable<AstNode> {
 }
 
 extension ExpressionExtensions on Expression {
```
munificent, line 90: Since this was only used in one place, I went ahead and inlined it there so that it was clearer what it applied to.

natebosch (approved):

PR #1228, opened 2023-05-25
add firehose workflow

jakemac53:
This adds the worflow I talked about before - I think its worth trying out and you can always remove it if you want. You will also need to enable automated publish in the pub admin page https://pub.dev/packages/dart_style/admin.

See https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose#integrating-this-tool-into-a-repo

Note that if you ever need to publish from a dev/beta SDK then we would want to require that sdk, see https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose#publishing-from-a-specific-version-of-the-sdk. Happy to add that config if you think it is necessary.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,14 @@
+# A CI configuration to auto-publish pub packages.
+
+name: Publish
+
+on:
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  push:
+    tags: [ 'v[0-9]+.[0-9]+.[0-9]+' ]
+
+jobs:
+  publish:
+    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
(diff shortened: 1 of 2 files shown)


jakemac53:
Note that firehose does enforce a certain format for the changelog so I had to update that also

jakemac53:
See https://github.com/dart-lang/logging/pull/140#issuecomment-1553286658 as an example of how the workflow works - when you bump the version to one that looks like you want to publish it will add this comment, once you merge you can just click the linked tag to create the release and it will auto-publish.

munificent (approved):
Thanks for adding this! I enabled publishing the package on the admin page, so I'll try this out the next time I'm ready to push a release.

munificent:
> you have the tag pattern set to just `{{version}}` where here I have it set to `v{{version}}` which is more the convention we use elsewhere - it looks like a mix has been used in this package.

Yes, I have always consistently used just the version number, but some other people who have published versions of the package didn't follow the existing convention.

> Whatever you use doesn't matter, but the admin page and config here do need to agree :). Should I remove the preceding `v` from this config?

Oh, good point. Switch to `v{{version}}` on the admin page since that's what we use for other packages.

PR #1229, opened 2023-05-25
Don't drop "?" on nullable empty record types.

### LINKED ISSUE #1224: dartfmt removes "?" in "()?" typed field
While adding records support to built_value I thought I'd check the base case, and it doesn't quite work as dartfmt eats the `?` in the generated output.

```
> cat foo.dart
class A {
  ()? x;
}
```

```
dartfmt foo.dart
Hit a bug in the formatter when formatting foo.dart.
The formatter produced unexpected output. Input was:
class A {
  ()? x;
}

Which formatted to:
class A {
  () x;
}
```

munificent:
Fix #1224.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Don't split before `.` following a record literal (#1213).
 * Don't force split on a line comment before a switch expression case (#1215).
 * Require `package:analyzer` `^5.12.0`.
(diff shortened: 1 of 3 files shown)


kevmoo (approved):
oops!

davidmorgan:
Thanks :)

PR #1230, opened 2023-05-27
Add a script to automatically update formatting test expectations.

munificent:
It's not fully featured. It doesn't handle tests with selection markers or Unicode escapes. But it handles everything else.

This should make it easier for me to update the tests as I work on the Flutter style experimental branch.


lib/src/testing/test_file.dart
```diff
@@ -0,0 +1,188 @@
...
+final _fixPattern = RegExp(r'\(fix ([a-x-]+)\)');
+final _unicodeUnescapePattern = RegExp(r'×([0-9a-fA-F]{2,4})');
+final _unicodeEscapePattern = RegExp('[\x0a\x0c\x0d]');
+
+/// Get the absolute local file path to the package's "test" directory.
+Future<String> findTestDirectory() async {
+  var libraryUri = await Isolate.resolvePackageUri(
+      Uri.parse('package:dart_style/src/testing/test_file.dart'));
+  return p.normalize(
+      (p.join(p.dirname(libraryUri!.toFilePath()), '../../../test')));
```
natebosch, line 18: [nit] looks like one extra set of parens

```suggestion
      p.join(p.dirname(libraryUri!.toFilePath()), '../../../test'));
```

lib/src/testing/test_file.dart
```diff
@@ -0,0 +1,188 @@
...
+  /// `test/` directory.
+  static Future<TestFile> read(String path) async {
+    var testDir = await findTestDirectory();
+    var file = File(p.join(testDir, path));
+    return TestFile._load(file, p.relative(file.path, from: testDir));
+  }
+
+  /// Reads the test file from [file].
+  factory TestFile._load(File file, String relativePath) {
+    // Explicitly create a File, in case the entry is a Link.
```
natebosch, line 49: I don't understand this comment. It looks like this no longer applies after being modified from it's original location.

natebosch (approved):

lib/src/testing/test_file.dart
munificent: Oops, yes. That comment apparently hasn't been true for a very long time. Removed.

PR #1233, opened 2023-06-13
Fix linter issues on dev in update_tests.dart

athomas:

PR #1234, opened 2023-06-13
Change branches to main test-package.yml

athomas:


athomas:
Noticed they didn't trigger on https://github.com/dart-lang/dart_style/pull/1233

PR #1236, opened 2023-06-21
Remove trailing whitespace on first line of multiline string.

lrhn:
Proof-of-concept removal of non-participating trailing space and backslashes from the first line of multiline strings.

Slightly dangerous, becuase it had to tweak the "changes only whitespace" checking to allow removing backslashes (only precedeing a whitespace). That could allow removing the backslash in a raw string literal like `r"\ "` without being noticed. 
(I know a current experiment also allows removing commas, so maybe it's easier to integrate into that experiment, if it has already generalized the "check whitespace only changes" test.)


test/whitespace/strings.stmt
```diff
@@ -32,4 +32,36 @@ one
...
+z''';
+<<<
+var x = r'''
+z''';
+>>> trailing space multiline doublequotes
+var x = """ \   \ \
+z""";
+<<<
+var x = """
+z""";
```
munificent, line 67: If you want to test the various kinds of whitespace characters, the tests support a custom little metasyntax for Unicode literals. Take a look at test/whitespace/unicode.txt for some examples.

lib/src/string_compare.dart
```diff
@@ -20,13 +20,32 @@ bool _isWhitespace(int c) {
...
 ///
-/// Returns `true` if current contains a non-whitespace character.
-/// Returns `false` if no characters are left.
+/// Searches [str] for non-whitespace characters, starting at [i]
+/// and ending at [len].
+///
+/// Skips whitespace characters *and* single backslashes before whitespace
+/// characters. The latter is allowed to account for the first line
+/// of multiline strings (but does risk allowing the removal of a backslash
+/// from a raw string.)
```
munificent, line 31: Discarding whitespace on the first line seems reasonable to me. We discard trailing whitespace in other places where it isn't semantic.

But treating an escaped space as a whitespace character and removing that feels like a step a little too far. We don't do escape normalization anywhere else (there is an argument that we should), so I think it's probably best to leave those alone.

In practice, I can't imagine any reasonable human writing an escaped space on the first line of a multi-line string anyway, so it's probably fine to not support cleaning that up.

munificent (approved):
Thanks for doing this! I like cleaning up the leading whitespace, but not normalizing escapes. Assuming you're OK with removing the code for cleaning up the escapes, LGTM after that.

munificent:
Oh, one more thing. Before merging this, please add an entry describing the change to the CHANGELOG. Thanks!

test/whitespace/strings.stmt
lrhn: Will do.

lrhn: And done.

lib/src/string_compare.dart
lrhn: Ended up normalizing everything around the backslashes, but retaining them (for absolutely no effect, they are effectively whitespace, that is ignored by everything that comes after).

lrhn:
Merged to head, updated changelog to be at next version. Tests still succeed, so landing.

PR #1246, opened 2023-08-09
Remove test for `inline class`

oprypin:
This test is obsolete and is now problematic because it seems to fail under some circumstances.

### DIFF
--- a/test/whitespace/classes.unit
+++ b/test/whitespace/classes.unit
@@ -244,7 +244,3 @@ abstract mixin class C12 = Object
     with Mixin;
 abstract base mixin class C13 = Object
     with Mixin;
->>> inline classes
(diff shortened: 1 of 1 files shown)


munificent:
Yeah, this test is failing now because the syntax for inline classes (now extension types) has changed. I think just removing the test for now is fine.

Would you mind also bumping the version number in the pubspec to:

```
version: 2.3.3-dev
```

Thanks!

munificent (approved):
One change then looks good, thanks!

munificent:
Actually, let's just close this in favor of https://github.com/dart-lang/dart_style/pull/1247, which also removes the code that uses the `inlineKeyword` part of the analyzer AST API.

PR #1256, opened 2023-08-28
Stop passing experiment flags to parser for features that have shipped.

### LINKED ISSUE #1245: Current experiment logic prevents us from being able to expire or remove experiment flags sanely
I was recently trying to retire the dart 3.0 experiment flags (https://dart-review.googlesource.com/c/sdk/+/316460) but this ended up breaking most of our internal codebase.

The current way we deal with language features is to just enable the experiment as soon as formatting for it is supported, and we always set the `sdkLanguageVersion` to 2.19.0.  This kind of works, until we go to retire or delete an experiment flag.

As an example my first fix for the current situation was to change the logic to parse with the current language version and _no_ experiments, and then after that try to parse as 2.19.0. That works  except that we have old versions of the formatter out there, and they will no longer work for any Dart 3.0 code.

There is no analyzer api for parsing with the current version plus some experiments that I can find (cc @scheglov I could be missing something) - if we had that we could add a third fallback that enables inline classes with the current language version (that feature has no experimental release version yet, but maybe we should require it to before it gets formatter support?).

We can somewhat alleviate the situation with breaking changes in analyzer any time an experiment is either expired _or_ deleted, but that is very arduous.

Longer term, simply forcing users to explicitly enable features is more maintainable I think. I understand that places a burden on things that use the formatter as a library, but not many things actually do that.

munificent:
Fix #1245.

cc @scheglov

### DIFF
--- a/lib/src/dart_formatter.dart
+++ b/lib/src/dart_formatter.dart
@@ -38,6 +38,11 @@ class DartFormatter {
 
   final Set<StyleFix> fixes;
 
+  /// Flags to enable experimental language features.
+  ///
+  /// See dart.dev/go/experiments for details.
+  final List<String> experimentFlags;
+
   /// Creates a new formatter for Dart code.
   ///
   /// If [lineEnding] is given, that will be used for any newlines in the
@@ -49,10 +54,15 @@ class DartFormatter {
   ///
   /// While formatting, also applies any of the given [fixes].
   DartFormatter(
-      {this.lineEnding, int? pageWidth, int? indent, Iterable<StyleFix>? fixes})
+      {this.lineEnding,
+      int? pageWidth,
+      int? indent,
+      Iterable<StyleFix>? fixes,
+      List<String>? experimentFlags})
       : pageWidth = pageWidth ?? 80,
         indent = indent ?? 0,
-        fixes = {...?fixes};
+        fixes = {...?fixes},
+        experimentFlags = [...?experimentFlags];
 
   /// Formats the given [source] string containing an entire Dart compilation
   /// unit.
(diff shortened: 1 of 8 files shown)


jakemac53 (approved):

jakemac53 (approved):

jakemac53 (commented):
I think we also need an API for actually enabling experiments also?

munificent:
> I think we also need an API for actually enabling experiments?

Good point! Done.

lib/src/dart_formatter.dart
```diff
@@ -202,17 +202,15 @@ class DartFormatter {
...
-        'inline-class',
-        'class-modifiers',
-        if (patterns) 'patterns',
-        'records',
-        'sealed-class',
-        'unnamed-libraries',
-      ],
-    );
+    FeatureSet featureSet;
+    if (patterns) {
```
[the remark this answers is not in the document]
munificent: The patterns part really is special here because the whole point of this function is a hack to workaround patterns breaking syntactic compatibility in a couple of corner cases.

lib/src/dart_formatter.dart
```diff
@@ -200,17 +210,9 @@ class DartFormatter {
...
-    FeatureSet featureSet;
-    if (patterns) {
-      featureSet = FeatureSet.latestLanguageVersion();
-    } else {
-      featureSet = FeatureSet.fromEnableFlags2(
-        sdkLanguageVersion: Version(2, 19, 0),
-        flags: [],
-      );
-    }
+    var version = patterns ? Version(3, 0, 0) : Version(2, 19, 0);
```
jakemac53, line 213:
```suggestion
    var version = patterns ? FeatureSet.latestLanguageVersion() : Version(2, 19, 0);
```

jakemac53 (approved):

lib/src/dart_formatter.dart
munificent: That doesn't work since `FeatureSet.latestLanguageVersion()` is a `FeatureSet` and not a `Version`. I can't just use that `FeatureSet` directly either, because then it doesn't accept a list of experiment flags.

I'll just stick with hardcoding the version. This should work fine because whenever new syntactic features are added, I have to update the formatter anyway and will update the version then.

jakemac53: I don't know that this will interact well with language experiments still? Maybe it is fine, I don't know for sure, but it definitely doesn't feel right to hardcode it?

PR #1263, opened 2023-09-12
Start building the new Piece internal representation and new style.

munificent:
This gets the foundation in place for the new IR, the new visitor that produces it, the new line splitter that consumes it, the new style, and new tests for it. It also adds support for formatting library, import, and export directives to make sure everything is wired up together and working.

Existing formatting behavior is unchanged. You have to opt in to the new stuff by passing "tall-style" as an experiment flag.

This PR doesn't support comments, but it does leave some unused code in a few places that will be used to handle comments in future PRs. It has many UnimplementedError throws. Those will get filled in as more of the language is implemented in the new style. There are also two new kinds of TODO comments:

```
// TODO(tall): ...
```

These comments describe that work that needs to be done before the new style is fully working.

```
// TODO(perf): ...
```

These describe potential areas for optimization. Once more of the language is supported with the new IR and I can run some larger benchmark samples through it, I can start exploring where the actual performance problems are. For now, I'm just leaving breadcrumbs.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -1,0 +1,198 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../constants.dart';
+import '../piece/piece.dart';
+import 'solution.dart';
+
+/// The interface that pieces use when to produce output formatting themselves
+/// during state solving.
+class CodeWriter {
+  final int _pageWidth;
+
+  /// The state values for the pieces being written.
+  final PieceStateSet _pieceStates;
+
+  /// Buffer for the code being written.
+  final StringBuffer _buffer = StringBuffer();
+
+  /// The cost of the currently chosen line splits.
+  int _cost = 0;
+
+  /// The number of characters of code that have overflowed the page width so
+  /// far.
+  int _overflow = 0;
+
+  /// How long the line currently being written is.
+  int _column = 0;
+
+  /// Whether this solution has encountered a newline where none is allowed.
+  /// This means the solution is invalid.
+  bool _containsInvalidNewline = false;
+
+  /// For each piece being formatted from a call to [format], this tracks the
+  /// indentation of any new lines it begins.
+  ///
+  /// This is used to increase the cumulative nesting as we recurse into pieces
+  /// and then unwind that as child pieces are completed.
+  final List<_PieceOptions> _pieceOptions = [_PieceOptions(0, 0, true)];
+
+  /// The options for the current innermost piece being formatted.
+  _PieceOptions get _options => _pieceOptions.last;
+
+  CodeWriter(this._pageWidth, this._pieceStates);
+
+  /// Returns the finished code produced by formatting the tree of pieces and
+  /// the final score.
+  (String, Score) finish() {
+    _finishLine();
+    return (
+      _buffer.toString(),
+      Score(isValid: !_containsInvalidNewline, overflow: _overflow, cost: _cost)
+    );
+  }
+
+  /// Notes that a newline has been written.
+  ///
+  /// If this occurs in a place where newlines are prohibited, then invalidates
+  /// the solution.
+  ///
+  /// This is called externally by [TextPiece] to let the writer know some of
+  /// the raw text contains a newline, which can happen in multi-line block
+  /// comments and multi-line string literals.
+  void handleNewline() {
+    if (!_options.allowNewlines) _containsInvalidNewline = true;
+  }
+
+  void write(String text) {
+    _buffer.write(text);
+    _column += text.length;
+  }
+
+  /// Sets the number of spaces of indentation for code written by the current
+  /// piece to [indent].
+  ///
+  /// Replaces any previous indentation set by this piece.
+  void setIndent(int indent) {
+    // Include indentation from surrounding pieces.
+    _options.indent = _pieceOptions[_pieceOptions.length - 2].indent + indent;
+  }
+
+  /// Increase the expression nesting of the current piece if [condition] is
+  /// `true`.
+  void nestIf(bool condition) {
+    if (!condition) return;
+
+    _options.nesting += Indent.expression;
+  }
+
+  /// Sets the number of spaces of expression nesting for code written by the
+  /// current piece to [nesting].
+  ///
+  /// Replaces any previous nesting set by this piece.
+  void setNesting(int nesting) {
+    // Include nesting from surrounding pieces.
+    _options.nesting =
+        _pieceOptions[_pieceOptions.length - 2].nesting + nesting;
+  }
+
+  /// Inserts a newline if [condition] is true.
+  ///
+  /// If [space] is `true` and [condition] is `false`, writes a space.
+  ///
+  /// If [indent] is given, sets the amount of block-level indentation for this
+  /// and all subsequent newlines to [indent].
+  void splitIf(bool condition, {bool space = true, int? indent}) {
+    if (indent != null) setIndent(indent);
+
+    if (condition) {
+      newline();
+    } else if (space) {
+      this.space();
+    }
+  }
+
+  /// Writes a single space to the output.
+  void space() {
+    write(' ');
+  }
+
+  /// Inserts a line split in the output. If [blank] is true, writes an extra
+  /// newline to produce a blank line.
+  void newline({bool blank = false}) {
+    handleNewline();
+    _finishLine();
+    _buffer.writeln();
+    if (blank) _buffer.writeln();
+
+    _column = _options.combinedIndentation;
+    _buffer.write(' ' * _column);
+  }
+
+  /// Sets whether newlines are allowed to occur from this point on for the
+  /// current piece of any of its children.
+  void setAllowNewlines(bool allowed) {
+    _options.allowNewlines = allowed;
+  }
+
+  /// Format [piece] and insert the result into the code.
+  void format(Piece piece) {
+    // Don't bother recursing into the piece tree if we know the solution will
+    // be discarded.
+    if (_containsInvalidNewline) return;
+
+    // TODO(tall): Sometimes, we'll want to reset the expression nesting for
+    // an inner piece, for when a block-like construct appears inside an
+    // expression. If it turns out that we don't actually need to handle indent
+    // and nesting separately here, then merge them into a single field.
+    _pieceOptions.add(_PieceOptions(
+        _options.indent, _options.nesting, _options.allowNewlines));
+
+    var state = _pieceStates.pieceState(piece);
+
+    // TODO(tall): Support pieces with different split costs, and possibly
+    // different costs for each state value.
+    if (state != 0) _cost++;
+
+    // TODO(perf): Memoize this. Might want to create a nested PieceWriter
+    // instead of passing in `this` so we can better control what state needs
+    // to be used as the key in the memoization table.
+    piece.format(this, state);
+
+    _pieceOptions.removeLast();
+  }
+
+  /// Format [piece] if not null.
+  void formatOptional(Piece? piece) {
+    if (piece != null) format(piece);
+  }
+
+  void _finishLine() {
+    // If the completed line is too long, track the overflow.
+    if (_column >= _pageWidth) {
+      _overflow += _column - _pageWidth;
+    }
+  }
+}
+
+/// Tracks the mutable state local to a single piece currently being formatted.
+class _PieceOptions {
+  /// The number of spaces of leading indentation coming from block-like
+  /// structure or explicit extra indentation (aligning constructor
+  /// initializers, `show` clauses, etc.).
+  int indent;
+
+  /// The number of spaces of indentation from wrapped expressions.
+  int nesting;
+
+  /// The total number of spaces of indentation.
+  int get combinedIndentation => indent + nesting;
+
+  /// Whether newlines are allowed to occur.
+  ///
+  /// If a newline is written while this is `false`, the entire solution is
+  /// considered invalid and gets discarded.
+  bool allowNewlines;
+
+  _PieceOptions(this.indent, this.nesting, this.allowNewlines);
+}
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -5,8 +5,21 @@ import '../constants.dart';
 import '../piece/piece.dart';
 import 'solution.dart';
 
-/// The interface that pieces use when to produce output formatting themselves
-/// during state solving.
+/// The interface used by [Piece]s to output formatted code.
+///
+/// The back-end lowers the tree of pieces to the final formatted code by
+/// allowing each piece to produce the output for the code it represents.
+/// This way, each piece has full flexibility for how to apply its own
+/// formatting logic.
+///
+/// To build the resulting output code, when a piece is formatted, it is passed
+/// an instance of this class. It has methods that the piece can call to add
+/// output text to the resulting code, recursively format child pieces, insert
+/// whitespace, etc.
+///
+/// This class also accumulates the score (the relative desireability of a set
+/// of formatting choices) that the resulting code has by tracking things like
+/// how many characters of code overflow the page width.
 class CodeWriter {
   final int _pageWidth;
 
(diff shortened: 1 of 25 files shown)


lib/src/back_end/solver.dart
```diff
@@ -0,0 +1,89 @@
...
+///     solutions with a minimum number of line splits first.
+///
+/// -   Solutions are explored in priority order. We explore solutions with the
+///     fewest overflow characters and the lowest cost (in that order) first.
+///     This was, as soon as we find a solution with no overflow characters, we
+///     know it will be the best solution and can stop.
+// TODO(perf): At some point, we probably also want to do memoization of
+// previously formatted Piece subtrees.
+class Solver {
+  final DartFormatter _formatter;
```
natebosch, line 29: It looks like we only need the `pageWidth` field from `_formatter` - can we store that directly instead of the DartFormatter?

lib/src/dart_formatter.dart
```diff
@@ -16,11 +16,21 @@ import 'package:analyzer/src/string_source.dart';
...
 import 'style_fix.dart';
 
+/// The in-progress "tall" formatting style is enabled by passing an experiment
+/// flag with this name.
+///
+/// Note that this isn't a real Dart SDK experiment: Only the formatter supports
+/// it. We use the [experimentFlags] API to pass this in so that support for it
+/// can be removed in a later version without it being a breaking change to the
+/// dart_style library API.
+const _tallStyleExperimentFlag = 'tall-style';
```
natebosch, line 32: Should this be public instead of repeated in other files?

lib/src/front_end/piece_factory.dart
```diff
@@ -0,0 +1,150 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+
+import '../piece/import.dart';
+import '../piece/piece.dart';
+import '../piece/postfix.dart';
+import '../piece/sequence.dart';
+import 'piece_writer.dart';
+
+/// Many AST nodes are structurally similar and receive similar formatting.
```
natebosch, line 13: Add a header sentence.

lib/src/front_end/piece_writer.dart
```diff
@@ -0,0 +1,157 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../back_end/solver.dart';
+import '../dart_formatter.dart';
+import '../debug.dart' as debug;
+import '../piece/piece.dart';
+import '../source_code.dart';
+
+/// Incrementally builds [Piece]s while visiting AST nodes.
```
natebosch, line 11: Why would a "Writer" be the one that "Builds" pieces? I would expect the "Factory" to "Build" them...

lib/src/front_end/piece_writer.dart
```diff
@@ -0,0 +1,157 @@
...
+///
+/// ```
+/// a + b;
+/// ```
+///
+/// The AST structure is like:
+///
+/// ```
+/// ExpressionStatement
+///   BinaryExpression
```
natebosch, line 24: Where is the `+`?

lib/src/front_end/piece_writer.dart
```diff
@@ -0,0 +1,157 @@
...
+/// But the resulting piece tree looks like:
+///
+/// ```
+/// Infix
+///   TextPiece("a +")
+///   TextPiece("b;")
+/// ```
+///
+/// Note how the infix operator is attached to the preceding piece (which
+/// happens to just be text but could be more complex. Notice all that there
```
natebosch, line 38: s/all/also ?

lib/src/piece/import.dart
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../back_end/code_writer.dart';
+import '../constants.dart';
+import 'piece.dart';
+
+/// An import or export directive.
```
natebosch, line 9: How about naming it "DirectivePiece"?

lib/src/piece/import.dart
```diff
@@ -0,0 +1,193 @@
...
+///
+/// Contains pieces for the keyword and URI, the optional `as` clause for
+/// imports, the configurations (`if` clauses), and combinators (`show` and
+/// `hide`).
+class ImportPiece extends Piece {
+  /// The main directive and its URI.
```
natebosch, line 15: [nit] How about

/// The directive keyword and it's URI.

I think of the entire thing as "the directive", and I don't know what "main directive" means

lib/src/piece/piece.dart
```diff
@@ -0,0 +1,81 @@
...
+class TextPiece extends Piece {
+  /// The lines of text in this piece.
+  ///
+  /// Most [TextPieces] will contain only a single line, but a piece with
+  /// preceding comments that are on their own line will have multiple. These
+  /// are stored as separate lines instead of a single multi-line string so that
+  /// each line can be indented appropriately during formatting.
+  final List<String> _lines = [];
+
+  /// True if this text piece contains or ends with a mandatory newline. This
```
natebosch, line 43: [nit] blank line

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../constants.dart';
+import '../piece/piece.dart';
+import 'solution.dart';
+
+/// The interface that pieces use when to produce output formatting themselves
```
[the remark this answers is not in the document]
munificent: Good idea, done!

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+  /// the solution.
+  ///
+  /// This is called externally by [TextPiece] to let the writer know some of
+  /// the raw text contains a newline, which can happen in multi-line block
+  /// comments and multi-line string literals.
+  void handleNewline() {
+    if (!_options.allowNewlines) _containsInvalidNewline = true;
+  }
+
+  void write(String text) {
```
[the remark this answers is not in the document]
munificent: It is actually possible for `text` to contain newlines. When it does, `TextPiece` is responsible for also calling `handleNewline()`. Added a comment explaining this.

(I could explicitly split any lexemes that contain newlines into separate lines and simplify this, but I try hard to minimize the amount of character-level string manipulation I do for performance reasons.)

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+    _buffer.write(text);
+    _column += text.length;
+  }
+
+  /// Sets the number of spaces of indentation for code written by the current
+  /// piece to [indent].
+  ///
+  /// Replaces any previous indentation set by this piece.
+  void setIndent(int indent) {
+    // Include indentation from surrounding pieces.
```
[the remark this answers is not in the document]
munificent: Good point, done.

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+    _finishLine();
+    _buffer.writeln();
+    if (blank) _buffer.writeln();
+
+    _column = _options.combinedIndentation;
+    _buffer.write(' ' * _column);
+  }
+
+  /// Sets whether newlines are allowed to occur from this point on for the
+  /// current piece of any of its children.
```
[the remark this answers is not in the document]
munificent: Oops, yes.

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+  void setAllowNewlines(bool allowed) {
+    _options.allowNewlines = allowed;
+  }
+
+  /// Format [piece] and insert the result into the code.
```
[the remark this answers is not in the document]
munificent: Done.

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+
+    // TODO(perf): Memoize this. Might want to create a nested PieceWriter
+    // instead of passing in `this` so we can better control what state needs
+    // to be used as the key in the memoization table.
+    piece.format(this, state);
+
+    _pieceOptions.removeLast();
+  }
+
+  /// Format [piece] if not null.
```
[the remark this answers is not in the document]
munificent: We can, yeah. It's just a convenience method.

As more and more of the language is supported, most of the of the formatter code will be:

* "Front-end" code in AstNodeVisitor and PieceFactory that traverses the AST and builds Pieces. That code mostly calls into the PieceWriter API. There will be a lot of this, so convenience methods in PieceWriter can be worth it.
* "Back-end" code in the `format()` method in Piece classes that traverses child pieces, inserts whitespace between them and writes code. That code mostly calls into CodeWriter here. There will be a lot of it, so convenience methods in CodeWriter can be worth it.

Think of PieceWriter is sort of the internal DSL for the AST->Piece translation rules, and CodeWriter as the internal DSL for the Piece->Text translation rules.

The surface area of both of those sets of rules is quite large (because Dart is a big language), so a lot of what I've been doing with the new IR is trying to figure out a nice API for the two DSLs so that it's easy and maintainable to pump out new Piece factory and format logic.

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+
+  void _finishLine() {
+    // If the completed line is too long, track the overflow.
+    if (_column >= _pageWidth) {
+      _overflow += _column - _pageWidth;
+    }
+  }
+}
+
+/// Tracks the mutable state local to a single piece currently being formatted.
```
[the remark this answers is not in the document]
munificent: Removed "currently". Also rewrite the comment for `_pieceOptions` to better explain that it's a stack.

lib/src/back_end/code_writer.dart
```diff
@@ -0,0 +1,198 @@
...
+class _PieceOptions {
+  /// The number of spaces of leading indentation coming from block-like
```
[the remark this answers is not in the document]
munificent: Done.

lib/src/back_end/solution.dart
```diff
@@ -0,0 +1,167 @@
...
+
+    // We are looking for *lower* costs and overflows, so an invalid score is
+    // considered higher or after all others.
+    if (!isValid) return 1;
+    if (!other.isValid) return -1;
+
+    // Overflow is always penalized more than line splitting cost.
+    if (overflow != other.overflow) return overflow.compareTo(other.overflow);
+
+    return cost.compareTo(other.cost);
```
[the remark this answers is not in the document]
munificent: It's not fleshed out yet in the new IR because I don't have enough of the language working where interesting cost comes into play. With the old Chunk IR and line splitter, each Rule can have a different cost value when it's forced to split and the costs are tweaked to prefer some solutions over others. Also, you can wrap a series of chunks in a Span and whenever a split occurs between those chunks, the Span's cost is added in.

I don't know how much of that we'll need for the new IR (I'm hoping a more explicitly tree-like IR lets us do something simpler), so write now it's just "the number of Pieces whose state isn't zero". But it's not documented yet because it's not fully baked yet. I'll fill it in more when more of the language is in place.

lib/src/back_end/solver.dart
munificent: Oh good catch. Done.

lib/src/dart_formatter.dart
munificent: dart_formatter.dart is exported by the package and I don't want the flag to be part of the public API (expect as a string you can pass in) so that I can remove it without it being a breaking change.

Moved it to constants.dart and made it public there.

lib/src/front_end/piece_writer.dart
munificent: Oh, God, I went back and forth on how to name these damn classes for *ages*.

I don't love any of the names, but my general thinking was to use "Writer" for classes that incrementally build something up. They are internally mutable and "own" the thing being built (which they generally yield at the end). So `PieceWriter` is used to incrementally build the piece tree. (I suppose I could have called it `PieceTreeWriter`, but that seems unnecessary.) And `CodeWriter` incrementally builds the final output code.

I considered "Builder" for these, but that felt even more ambiguous with "Factory".

I used "Factory" to refer to functionality that produces new objects but doesn't have any interesting state or mutability itself. It's a little odd here because usually factory methods *return* the thing they make. I tried to do that here, but it doesn't work well because the `Piece` tree node boundaries often don't align well with the AST node boundaries. That's why `PieceWriter` has the `push()` and `pop()` API.

(I could have forced the piece tree to align with the AST node boundaries so that every AST node is a piece. But that would lead to a much larger piece tree with many places where no split can occur between a piece boundary anyway. I worry that would bleed perf for little benefit. My goal is to make the piece tree as coarse-grained as possible because that gives the back end, which is where all the perf challenges are, less to chew on.)

I had `PieceMaker` for a while too, which I loved for the homophonic pun, but otherwise didn't seem to clarify. Naming is hard.

The other option would be to lump it all together. There's nothing preventing everything from being right inside AstNodeVisitor but then it becomes a giant monster of a class.

lib/src/front_end/piece_writer.dart
munificent: Fixed.

lib/src/piece/import.dart
munificent: Libraries are directives too. :-/ The Analyzer API calls the union of "import and export directive" a "namespace directive", but that feels a little... arbitrary.

The comment in [PieceFactory] is relevant here:

```dart
/// Naming these functions can be hard. For example, there isn't an obvious
/// word for "import or export directive" or "named thing with argument list".
/// To avoid that, we pick one concrete construct formatted by the function,
/// usually the most common, and name it after that, as in [createImport()].
```

I ran into this problem all the time with the old IR. Collection literals, function calls with trailing commas, and lambdas get similar formatting. What is a word that describes all of those? (I started using "block", but that word already means a couple of other things, so it gets confusing.) Or a word that describes an asserts and a function call?

So with the new one, the idea is that sometimes I'll just pick one member of the set of things that share formatting logic, call it canonical, and use that for the name. When I can, I try to pick a more general name, like "Sequence" or "Infix". But otherwise, I just pick a specific name and then document the set of things it applies to.

munificent:
Thanks!

lib/src/piece/piece.dart
```diff
@@ -0,0 +1,82 @@
...
+import '../back_end/code_writer.dart';
+
+/// Base class for the formatter's internal representation used for line
+/// splitting.
+///
+/// We visit the source AST and convert it to a tree of [Piece]s. This tree
+/// roughly follows the AST but includes comments and is optimized for
+/// formatting and line splitting. The final output is then determined by
+/// deciding which pieces split and how.
+abstract class Piece {
```
natebosch, line 14: How about `abstract interface class` since there is no inherited behavior?

natebosch (approved):

lib/src/piece/piece.dart
munificent: Within a single codebase, I tend to default to `extends`. That way if I do end up defining some inherited behavior in the base class (which is fairly often for me), it just works.

natebosch: I think it's an OK choice to make, but I do think this usage of `extends` prioritizes writing over reading. When I read `implements` I need look no further to understand how the class behaves - when I read `extends` I have to make sure I understand parent class behavior. I also need to keep in mind that some of the behavior is intended for consumption by other callers, and some behavior may be intended for consumption by the parent class. In this case it's a quick hop to see no behavior exists, but the `implements` would have saved me even that hop.

lib/src/piece/import.dart
natebosch: Just thought of another option, what do you think about `UriDirectivePiece`? Library directives may have a name, but not a URI, and imports/exports both always have an URI.

> I ran into this problem all the time with the old IR. Collection literals, function calls with trailing commas, and lambdas get similar formatting. What is a word that describes all of those?

`DelimitedCommaBlockPiece`?

I do see the tension between descriptive and terse names - I think a name based an a representative example has an opportunity cost and it might be worth bikeshedding a bit.

PR #1268, opened 2023-09-14
Format blocks and labeled statements.

munificent:
This doesn't support blank lines between statements in blocks. That will happen when comments are supported, which this is leading up to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.3.3-dev
+## 2.3.3
 
 * Always split enum declarations containing a line comment (#1254).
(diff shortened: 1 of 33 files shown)


natebosch:
I think the merge base and the commits on this branch got out of alignment - do you want me to try to find a commit range to review a diff in this PR, or should we wait until other commits settle and we can merge in the right commits to have a clean diff here?

munificent:
I'll resync the branch. The `format-binary` is the next one in the chain to review and this one comes after that, followed by `format-comments`. (The PR number order should tell you what order to review them.)

munificent:
OK, this should be in a reviewable state now.

natebosch (approved):

PR #1269, opened 2023-09-14
Include comments and discretionary newlines in output.

munificent:
Handling comments is the next big piece of the new representation. My
goal was to come up with something simpler than the old very hard to
maintain code in ChunkBuilder.

This seems to look pretty good at least in the subset of the language
that is supported so far.

### DIFF
--- a/lib/src/chunk.dart
+++ b/lib/src/chunk.dart
@@ -5,43 +5,7 @@ import 'fast_hash.dart';
 import 'marking_scheme.dart';
 import 'nesting_level.dart';
 import 'rule/rule.dart';
-
-/// Tracks where a selection start or end point may appear in some piece of
-/// text.
-abstract class Selection {
-  /// The chunk of text.
-  String get text;
-
-  /// The offset from the beginning of [text] where the selection starts, or
-  /// `null` if the selection does not start within this chunk.
-  int? get selectionStart => _selectionStart;
-  int? _selectionStart;
-
-  /// The offset from the beginning of [text] where the selection ends, or
-  /// `null` if the selection does not start within this chunk.
-  int? get selectionEnd => _selectionEnd;
-  int? _selectionEnd;
-
-  /// Sets [selectionStart] to be [start] characters into [text].
-  void startSelection(int start) {
-    _selectionStart = start;
-  }
-
-  /// Sets [selectionStart] to be [fromEnd] characters from the end of [text].
-  void startSelectionFromEnd(int fromEnd) {
-    _selectionStart = text.length - fromEnd;
-  }
-
-  /// Sets [selectionEnd] to be [end] characters into [text].
-  void endSelection(int end) {
-    _selectionEnd = end;
-  }
-
-  /// Sets [selectionEnd] to be [fromEnd] characters from the end of [text].
-  void endSelectionFromEnd(int fromEnd) {
-    _selectionEnd = text.length - fromEnd;
-  }
-}
+import 'selection.dart';
 
 /// A chunk of non-breaking output text that may begin on a newline.
 ///
@@ -308,45 +272,4 @@ class Span extends FastHash with Markable {
 
   @override
   String toString() => '$id\$$cost';
-}
-
-enum CommentType {
-  /// A `///` or `/**` doc comment.
-  doc,
-
-  /// A non-doc line comment.
-  line,
-
-  /// A `/* ... */` comment that should be on its own line.
-  block,
-
-  /// A `/* ... */` comment that can share a line with other code.
-  inlineBlock,
-}
-
-/// A comment in the source, with a bit of information about the surrounding
-/// whitespace.
-class SourceComment extends Selection {
-  /// The text of the comment, including `//`, `/*`, and `*/`.
-  @override
-  final String text;
-
-  /// What kind of comment this is.
-  final CommentType type;
-
-  /// The number of newlines between the comment or token preceding this comment
-  /// and the beginning of this one.
-  ///
-  /// Will be zero if the comment is a trailing one.
-  int linesBefore;
-
-  /// Whether this comment starts at column one in the source.
-  ///
-  /// Comments that start at the start of the line will not be indented in the
-  /// output. This way, commented out chunks of code do not get erroneously
-  /// re-indented.
-  final bool flushLeft;
-
-  SourceComment(this.text, this.type, this.linesBefore,
-      {required this.flushLeft});
 }
--- a/lib/src/front_end/comment_writer.dart
+++ b/lib/src/front_end/comment_writer.dart
@@ -12,7 +12,11 @@ import 'piece_writer.dart';
 /// comment tokens between meaningful tokens used by AST nodes.
 ///
 /// Also handles preserving discretionary blank lines in places where they are
-/// allowed.
+/// allowed. These are handled with comments because both comments and
+/// whitespace are found between the linear series of [Token]s produced by the
+/// analyzer parser. Likewise, both are output as whitespace (in the sense of
+/// not being executable code) interleaved with the [Piece]-building code that
+/// walks the actual AST and processes the code tokens.
 ///
 /// Comments are a challenge because they confound the intuitive tree-like
 /// structure of the code. A comment can appear between any two tokens, and a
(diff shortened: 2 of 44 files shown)


lib/src/chunk.dart
```diff
@@ -5,43 +5,7 @@ import 'fast_hash.dart';
 import 'marking_scheme.dart';
 import 'nesting_level.dart';
 import 'rule/rule.dart';
-
```
munificent, line 8: This was just moved to a separate file so that the new code doesn't import `chunk.dart`, which is part of the old IR.

lib/src/chunk.dart
```diff
@@ -309,44 +273,3 @@ class Span extends FastHash with Markable {
   @override
   String toString() => '$id\$$cost';
 }
-
```
munificent, line 312: Likewise this too.

lib/src/front_end/comment_writer.dart
```diff
@@ -0,0 +1,277 @@
...
+
+import '../piece/sequence.dart';
+import '../source_comment.dart';
+import 'piece_writer.dart';
+
+/// Functionality used by [AstNodeVisitor] to build text and pieces from the
+/// comment tokens between meaningful tokens used by AST nodes.
+///
+/// Also handles preserving discretionary blank lines in places where they are
+/// allowed.
```
natebosch, line 15: This is surprising - can you expand this a bit to indicate why comments and blank lines are similar and worth implementing together?

lib/src/front_end/comment_writer.dart
```diff
@@ -0,0 +1,277 @@
...
+        flushLeft = false;
+
+        // Always add a blank line (if possible) before a doc comment block.
+        if (comment == token.precedingComments) linesBefore = 2;
+      }
+
+      var type = CommentType.block;
+      if (text.startsWith('///') && !text.startsWith('////') ||
+          text.startsWith('/**') && text != '/**/') {
+        // TODO(rnystrom): Check that the comment isn't '/**/' because some of
```
natebosch, line 197: Why is this a TODO - what do we expect to change here? The current code looks correct to me

lib/src/front_end/comment_writer.dart
```diff
@@ -0,0 +1,277 @@
...
+
+    return (comments, tokenLine - previousLine);
+  }
+
+  /// Whether [comment] should be attached to the preceding token.
+  bool _isHangingComment(SourceComment comment) {
+    // Don't move a comment to a preceding line.
+    if (comment.linesBefore != 0) return false;
+
+    // Multi-line comments are always pushed to the next line.
```
natebosch, line 230: Doc and block comments can be a single line. `inlineBlock` comments can be multiple lines.

What is this "Multi-line" referring to? I don't think it can mean either of comments which _are_ or _can be_ multiple lines if the criteria is `doc || block`.

lib/src/front_end/comment_writer.dart
```diff
@@ -0,0 +1,277 @@
...
+        // refactoring the Chunk representation to move splits to the front of
+        // Chunk, I want to preserve the current test behavior. The fact that
+        // those tests pass with the old representation is a buggy quirk of the
+        // comment handling.
+        type = CommentType.doc;
+      } else if (comment.type == TokenType.SINGLE_LINE_COMMENT) {
+        type = CommentType.line;
+      } else if (commentLine == previousLine || commentLine == tokenLine) {
+        type = CommentType.inlineBlock;
+      }
```
natebosch, line 208: [optional] I missed this defaulting on my first read-through. WDYT about making the default a trailing `else`?

```suggestion
      CommentType type;
      if (text.startsWith('///') && !text.startsWith('////') ||
          text.startsWith('/**') && text != '/**/') {
        // TODO(rnystrom): Check that the comment isn't '/**/' because some of
        // the dart_style tests use that to mean inline block comments. While
        // refactoring the Chunk representation to move splits to the front of
        // Chunk, I want to preserve the current test behavior. The fact that
        // those tests pass with the old representation is a buggy quirk of the
        // comment handling.
        type = CommentType.doc;
      } else if (comment.type == TokenType.SINGLE_LINE_COMMENT) {
        type = CommentType.line;
      } else if (commentLine == previousLine || commentLine == tokenLine) {
        type = CommentType.inlineBlock;
      } else {
        type = CommentType.block
      }
```

lib/src/selection.dart
```diff
@@ -0,0 +1,39 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// Where a selection start or end point may appear in some piece of text.
+abstract class Selection {
+  /// The chunk of text.
+  String get text;
+
+  /// The offset from the beginning of [text] where the selection starts, or
+  /// `null` if the selection does not start within this chunk.
+  int? get selectionStart => _selectionStart;
```
natebosch, line 12: This file has the word "selection" a lot...

Can we avoid repeating the class name context in it's members?

```dart
int? get start;
int? get end;
void startAt(int start);
void startBackFromTextEnd(int fromEnd);
void endAt(int end);
void endBackFromTextEnd(int fromEnd);
```

lib/src/selection.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Where a selection start or end point may appear in some piece of text.
```
natebosch, line 5:
```suggestion
/// A region of text selected by an optional start and end index.
```

Could consider calling out that the indices are for UTF-16 code units, but that's probably what most readers would assume.

lib/src/selection.dart
```diff
@@ -0,0 +1,39 @@
...
+  /// `null` if the selection does not start within this chunk.
+  int? get selectionStart => _selectionStart;
+  int? _selectionStart;
+
+  /// The offset from the beginning of [text] where the selection ends, or
+  /// `null` if the selection does not start within this chunk.
+  int? get selectionEnd => _selectionEnd;
+  int? _selectionEnd;
+
+  /// Sets [selectionStart] to be [start] characters into [text].
```
natebosch, line 20: Dart `length` and string indexes do not handle "characters" the way that authors think of them. Can we replace all mentions of "characters" in this class with "code units" or "UTF-16 code units"?

lib/src/source_comment.dart
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/// A comment in the source, with a bit of information about the surrounding
+/// whitespace.
+class SourceComment extends Selection {
+  /// The text of the comment, including `//`, `/*`, and `*/`.
+  @override
+  final String text;
+
+  /// What kind of comment this is.
```
natebosch, line 27: [nit] This doc is not adding information. Can it be removed?

lib/src/front_end/comment_writer.dart
munificent: Good point, done. (The old formatter worked this way too, so I didn't think to write it down.)

lib/src/front_end/comment_writer.dart
munificent: Good point. Updated the comment.

lib/src/front_end/comment_writer.dart
munificent: Rewrote the comment to clarify. Also added some docs to the enum values.

The distinction between `CommentType.block` and `CommentType.inlineBlock` is a little weird. You get the latter for any C-style comment where there is also non-comment code on the same line as it. Whether that's a *good* distinction to make... I'm not sure. It's been like this for many years, so I'm leaving it alone for now.

This part of the code was basically lifted directly from the old formatter. (I didn't reuse it because eventually the old code will be deleted.)

lib/src/front_end/comment_writer.dart
munificent: Good idea. Done in my commit.

lib/src/selection.dart
munificent: Done.

lib/src/front_end/comment_writer.dart
```diff
@@ -0,0 +1,280 @@
...
+
+        // Always add a blank line (if possible) before a doc comment block.
+        if (comment == token.precedingComments) linesBefore = 2;
+      }
+
+      CommentType type;
+      if (text.startsWith('///') && !text.startsWith('////') ||
+          text.startsWith('/**') && text != '/**/') {
+        // We decide, somewhat arbitrarily, to treat `/**/` as a regular block
+        // comment and not a JavaDoc-style doc comment.
```
natebosch, line 202: I don't think this is arbitrary at all. `/**/` is not a JavaDoc-Style doc comment. `/** */` and `/***/` _might_ be, but I don't know whether any tool would consider an empty doc meaningfully different from an omitted doc. `/**/` is to few characters to even be an empty doc.

lib/src/selection.dart
munificent: It looks a little weird here, but this class is used as a base class for `Chunk` and `SourceComment` and in the context of those, it makes sense for the members to have more explicit selection-related names.

lib/src/front_end/comment_writer.dart
munificent: Fair. Deleted comment.

natebosch (approved):

PR #1271, opened 2023-09-15
Bump pubspec now that there are unpublished changes.

munificent:
(I forgot to do this when I merged the first new IR PR onto main.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.3.4-dev
+## 2.3.4-wip
 
 * Add `tall-style` experiment flag to enable the in-progress unstable new
   formatting style (#1253).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.3.4-dev
```
natebosch, line 1: [nit] We recently changed the guidance to use `-wip` as a suffix. `-dev` can cause some confusion because there are legitimate workflows (that the Dar team usually doesn't use) for publishing with a `-dev` suffix, while `-wip` is more clearly not intended for publish.

natebosch (approved):

CHANGELOG.md
munificent: SGTM. Done.

PR #1276, opened 2023-09-26
Add formatting for `extension type` declarations.

lrhn:
Attempt to format extension types. Since they are experimental, the syntax and formatting can still change before launch,
but this should make it *possible* to format code containing extension type declarations.

Which will make writing tests much nicer.

### DIFF
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -1294,8 +1294,10 @@ class SourceVisitor extends ThrowingAstVisitor {
     space();
     token(node.name);
 
+    builder.nestExpression();
     visit(node.typeParameters);
     visit(node.representation);
+    builder.unnest();
 
     builder.startRule(CombinatorRule());
     visit(node.implementsClause);
--- a/lib/src/source_visitor.dart
+++ b/lib/src/source_visitor.dart
@@ -2633,48 +2633,8 @@ class SourceVisitor extends ThrowingAstVisitor {
   @override
   void visitRepresentationDeclaration(RepresentationDeclaration node) {
     // Consider having `node.asFormalParameterList()`, and format based on that.
-    // Inline `visitFormalParameterList`, remove all that doesn't apply.
-    void writeAsParameter() {
-      // Inline `visitSimpleFormalParameter`.
-      visitParameterMetadata(node.fieldMetadata, () {
-        // Inline `_beginFormalParameter`
-        builder.startLazyRule(Rule(Cost.parameterType));
-        builder.nestExpression();
-
-        visit(node.fieldType);
-        _separatorBetweenTypeAndVariable(node.fieldType);
-        token(node.fieldName);
-
-        // Inline `_endFormalParameter`
-        builder.unnest();
-        builder.endRule();
-      });
-    }
 
     visit(node.constructorName);
-    if (node.commaAfter != null) {
-      // Trailing comma is not currently allowed, so this code is not tested.
-      // Inline _visitTrailingCommaParameterList
-      builder.startRule(Rule.hard());
-      token(node.leftParenthesis);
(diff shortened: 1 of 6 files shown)


lib/src/source_visitor.dart
```diff
@@ -2599,6 +2623,53 @@ class SourceVisitor extends ThrowingAstVisitor {
...
 
+  @override
+  void visitRepresentationConstructorName(RepresentationConstructorName node) {
+    token(node.period);
+    token(node.name);
+  }
+
+  @override
+  void visitRepresentationDeclaration(RepresentationDeclaration node) {
+    // Consider having `node.asFormalParameterList()`, and format based on that.
```
natebosch, line 2634: [nit] I don't think this is a useful comment unless it's a TODO with an issue attached.

lib/src/source_visitor.dart
```diff
@@ -2599,6 +2623,53 @@ class SourceVisitor extends ThrowingAstVisitor {
...
+    final rule = PositionalRule(null, argumentCount: 1);
+
+    builder.startRule(rule);
+    rule.beforeArgument(zeroSplit());
+
+    // Make sure record and function type parameter lists are indented.
+    builder.startBlockArgumentNesting();
+    builder.startSpan();
+
+    // Inline `visitFormalParameterList`, remove all that doesn't apply
```
natebosch, line 2649: [nit] This and the following comments are useful during the review but will lose relevance because they are so temporally specific. I think they can be removed from the source and added to the commit message.

natebosch (approved):

test/whitespace/extension_types.unit
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+}
+>>> all tokens
+@Anno<int, int>() extension type const A<S, T>.name(@required Map<int, int> a) implements I<S>, J<T> {}
+<<<
+@Anno<int, int>()
+extension type const A<S, T>.name(
+    @required Map<int, int> a)
+    implements I<S>, J<T> {}
```
munificent, line 24: This doesn't look great to me. There's an implicit principle in the formatter that adjacent lines shouldn't have the same indentation but for different reasons. In this case, line three is implemented +4 because of the representation type's parameter list and the next line is indented +4 because it's a split `implements` clause.

If the representation type splits, I would indent it past the `implements` clause, like:

```dart
@Anno<int, int>()
extension type const A<S, T>.name(
        @required Map<int, int> a)
    implements I<S>, J<T> {}
```

(Eventually, with the new style, it will look more like, maybe:

```dart
@Anno<int, int>()
extension type const A<S, T>.name(
  @required Map<int, int> a,
) implements I<S>, J<T> {}
```

)

To implement this, try moving:

```dart
builder.startRule(CombinatorRule());
```

Above:

```dart
    visit(node.typeParameters);
    visit(node.representation);
```

And then you might need to wrap another `nestExpression()` `unnest()` pair around:

```dart
    builder.startRule(CombinatorRule());
    visit(node.implementsClause);
    builder.endRule();
```

(I could be wrong about this. There tends to be some trial and error with hacking on the formatter. That's one of the reasons I want to move to a better internal representation.)

munificent (approved):
Couple of changes then LGTM.

test/whitespace/extension_types.unit
lrhn: We want the "primary construtor parameter" to be indented more *iff* there is a following `implements` clause, and they are all on separate lines.

I seem to have succeeded in that using:
```dart
    builder.nestExpression();
    visit(node.typeParameters);
    visit(node.representation);
    builder.unnest();

    builder.startRule(CombinatorRule());
    visit(node.implementsClause);
    builder.endRule();
```

PR #1279, opened 2023-10-12
Allow non-contiguous state values.

munificent:
Requiring the states to be a zero-based and contiguous makes it harder for some pieces to support different sets of different states based on what they contain. For example, an ImportPiece only supports states for splitting the names inside a single combinator when there is more than one combinator.

To model that with contiguous states requires the piece to carefully track how state indexes are shifted or splitting pieces into different classes to handle the different sets of states.

Instead, with this change, we simply allow the piece to return a list of states that may be non-contiguous. They can then omit states that don't make sense given the piece's contents. This makes ImportPiece simpler and I think will be useful for other pieces that have a variety of ways they can split.

Also wrap state values in a class to make things a little clearer and more type safe.

### DIFF
--- a/lib/src/piece/piece.dart
+++ b/lib/src/piece/piece.dart
@@ -12,16 +12,19 @@ import '../back_end/code_writer.dart';
 /// formatting and line splitting. The final output is then determined by
 /// deciding which pieces split and how.
 abstract class Piece {
-  /// The number of different ways this piece can be split.
+  /// The ordered list of indexes identifying each way this piece can split.
   ///
-  /// States are numbered incrementally starting at zero. State zero should
-  /// always be the lowest cost state with the fewest line splits. Lower states
-  /// should generally be preferred over higher states.
-  int get stateCount;
+  /// Each piece determines what each integer value in the list represents. The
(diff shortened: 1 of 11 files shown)


lib/src/piece/piece.dart
```diff
@@ -12,16 +12,19 @@ import '../back_end/code_writer.dart';
...
 /// deciding which pieces split and how.
 abstract class Piece {
-  /// The number of different ways this piece can be split.
+  /// The ordered list of indexes identifying each way this piece can split.
   ///
-  /// States are numbered incrementally starting at zero. State zero should
-  /// always be the lowest cost state with the fewest line splits. Lower states
-  /// should generally be preferred over higher states.
-  int get stateCount;
+  /// Each piece determines what each integer value in the list represents. The
```
natebosch, line 17:
```suggestion
  /// Each piece determines what each value in the list represents. The
```

natebosch (approved):

lib/src/piece/piece.dart
munificent: I'll fix this locally so that I can re-wrap the rest of the comment.

PR #1280, opened 2023-10-12
Format top-level and local variable declarations.

munificent:
This required tweaking the way partial solutions are ordered so that the solver doesn't avoid overflowing solutions that will ultimately lead to non-overflowing lower cost ones.

### DIFF
--- a/lib/src/piece/variable.dart
+++ b/lib/src/piece/variable.dart
@@ -1,0 +1,99 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../back_end/code_writer.dart';
+import '../constants.dart';
+import 'piece.dart';
+
+/// A variable declaration.
+///
+/// Used for local variable declaration statements, top-level variable
+/// declarations and field declarations.
+///
+/// Typed and untyped variables have slightly different splitting logic.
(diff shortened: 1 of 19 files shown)


lib/src/piece/variable.dart
```diff
@@ -0,0 +1,99 @@
...
+/// VeryLongTypeName
+/// longVariableName = initializer,
+/// anotherVariable = anotherInitializer;
+/// ```
+class VariablePiece extends Piece {
+  /// Split between each variable in a multiple variable declaration.
+  static const State _betweenVariables = State(1);
+
+  /// Split after the type annotation and between each variable.
+  static const State _afterType = State(2);
```
natebosch, line 42: [optional]

```suggestion
  static const _betweenVariables = State(1);

  /// Split after the type annotation and between each variable.
  static const _afterType = State(2);
```

natebosch (approved):
The commit message could use much more detail.

lib/src/piece/variable.dart
munificent: I'm not sure if it's habit or an actually good idea, but my consistent practice is to always type annotate members. <shrug>

natebosch: > consistent practice

https://github.com/dart-lang/dart_style/pull/1279/files#diff-dc088a2965c2b639564656129400eaad08b4e1c424d03bd272d941431146aaceR85-R91

😝

PR #1281, opened 2023-10-13
Format list literals (and some other things).

munificent:
I'm trying to keep these PRs mostly well-contained and not just adding a random pile of formatting. But I'm also trying to make sure I don't drop stuff on the floor. For this one, I filled in a few more things that lists touch on:

- Handle type argument lists in lists and function calls.
- Support block-like formatting for right-hand sides of `=`.
- Format parenthesized expressions.
- Format named type annotations with type arguments and nullable types.
- Add support for states with different costs so we can prioritize splitting choices.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -176,7 +176,7 @@ class CodeWriter {
 
     // TODO(tall): Support pieces with different split costs, and possibly
     // different costs for each state value.
-    if (state != 0) _cost++;
+    if (state != State.initial) _cost++;
 
     // TODO(perf): Memoize this. Might want to create a nested PieceWriter
     // instead of passing in `this` so we can better control what state needs
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -174,9 +174,7 @@ class CodeWriter {
 
     var state = _pieceStates.pieceState(piece);
 
-    // TODO(tall): Support pieces with different split costs, and possibly
-    // different costs for each state value.
-    if (state != State.initial) _cost++;
+    _cost += state.cost;
 
     // TODO(perf): Memoize this. Might want to create a nested PieceWriter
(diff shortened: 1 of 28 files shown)


lib/src/back_end/code_writer.dart
```diff
@@ -174,9 +174,7 @@ class CodeWriter {
 
     var state = _pieceStates.pieceState(piece);
 
-    // TODO(tall): Support pieces with different split costs, and possibly
-    // different costs for each state value.
-    if (state != State.initial) _cost++;
+    _cost += state.cost;
```
natebosch, line 177: This is another impactful change worthy of mention in the changelog.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -526,7 +526,30 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+    //   4, 5, 6,
+    // ];
+    // ```
+    //
+    // The formatter will preserve the newline after element 3 and the lack of
+    // them after the other elements.
+
+    for (var element in node.elements) {
+      builder.add(element);
+    }
```
natebosch, line 549: [optional]
```suggestion
    node.elements.forEach(builder.add);
```

lib/src/piece/assign.dart
```diff
@@ -0,0 +1,83 @@
...
+/// [State.split] Split after the `=`:
+///
+/// ```
+/// var name =
+///   longValueExpression;
+/// ```
+class AssignPiece extends Piece {
+  /// Split inside the value but not at the `=`.
+  ///
+  /// This is only allowed when the value is a delimitet expression.
```
natebosch, line 37:
```suggestion
  /// This is only allowed when the value is a delimited expression.
```

natebosch (approved):

lib/src/back_end/code_writer.dart
munificent: Good point, done.

PR #1284, opened 2023-10-16
Format if and return statements.

munificent:
I also tweaked the CodeWriter API to make it easier to update the indentation when inserting a newline. This doesn't do anything radically new, but makes the piece formatting code a little shorter.

Made a field in SequencePiece private that didn't need to be public.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -7,6 +7,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import '../ast_extensions.dart';
 import '../piece/assign.dart';
 import '../piece/block.dart';
+import '../piece/if.dart';
 import '../piece/import.dart';
 import '../piece/infix.dart';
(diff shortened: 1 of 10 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -123,6 +133,69 @@ mixin PieceFactory implements CommentWriter {
     }
   }
 
+  // TODO(tall): Generalize this to work with if elements too.
```
natebosch, line 136: Why would this be useful? Do they share a representation in the analyzer?

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: They don't have the same analyzer representation (alas), but they should have the same formatting logic. So what I'll do is write some sort of shim/conversion layer that lets me use the same IfPiece and (hopefully) same lowering code for both if statements and if elements.

PR #1288, opened 2023-10-19
Format function types.

munificent:
This involves formatting parameter lists which is a big piece of work. They work similar to argument lists and collection literals... except for the special syntax around `[ ... ]` and `{ ... }` for optional parameters.

Getting that working right involved adding some more functionality to DelimitedListBuilder and ListPiece.

Getting comments working with that syntax was particularly difficult. I ended up deciding that the formatter can move comments before or after the optional parameter delimiters if it needs to. Doing so produces better looking output, as in:

```
// Before:
f(// weird comment location
[param] // another
);

// After:
f([
  // weird comment location
  param, // another
]);
```

It also, I think is cleaner to implement than trying to laboriously keep the comment where the user authored it.

In the process of reworking how comments are handled in lists, I also tweaked how inline block comments are handled before commas. Previously, they would get moved after the comma, like:

```
// Before:
[element /* comment */, another];

// After:
[element, /* comment */ another];
```

Now, they stay where they were, which is I think what users want. A common use case for block comments in argument/collection lists is describing the meaning of a preceding null argument and this keeps that.

Line after an element but before the comma continue to be moved after the comma:

```
// Before:
f(param // comment
,);

// After:
f(
  param, // comment
);
```

Also:

- Added tests for the loosened behavior of equalIgnoringWhitespace().
- Added a test for a corner case of variable splitting that I ran into with parameters, which use the same VariablePiece.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -81,6 +81,10 @@ class CodeWriter {
   /// comments and multi-line string literals.
   void handleNewline() {
     if (!_options.allowNewlines) _containsInvalidNewline = true;
+
+    // Note that this piece contains a newline so that we can propagate that
+    // up to containing pieces too.
+    _options.hasNewline = true;
   }
 
   /// Appends [text] to the output.
@@ -158,7 +162,7 @@ class CodeWriter {
   }
 
   /// Sets whether newlines are allowed to occur from this point on for the
-  /// current piece or any of its children.
+  /// current piece.
   void setAllowNewlines(bool allowed) {
     _options.allowNewlines = allowed;
   }
@@ -186,7 +190,12 @@ class CodeWriter {
     // to be used as the key in the memoization table.
     piece.format(this, state);
 
-    _pieceOptions.removeLast();
+    var childOptions = _pieceOptions.removeLast();
+
+    // If the child [piece] contains a newline then this one transitively does.
+    // TODO(tall): At some point, we may want to provide an API so that pieces
+    // can block this from propagating outward.
+    if (childOptions.hasNewline) handleNewline();
   }
 
   /// Format [piece] if not null.
@@ -220,6 +229,9 @@ class _PieceOptions {
   /// If a newline is written while this is `false`, the entire solution is
   /// considered invalid and gets discarded.
   bool allowNewlines;
+
+  /// Whether any newlines have occurred in this piece or any of its children.
+  bool hasNewline = false;
 
   _PieceOptions(this.indent, this.nesting, this.allowNewlines);
 }
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -267,7 +267,11 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitDefaultFormalParameter(DefaultFormalParameter node) {
-    throw UnimplementedError();
+    visit(node.parameter);
+
+    // TODO(tall): Implement default values when function declarations are
+    // implemented.
+    if (node.separator != null) throw UnimplementedError();
   }
 
   @override
@@ -357,7 +361,35 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFormalParameterList(FormalParameterList node) {
-    throw UnimplementedError();
+    // Find the parameter immediately preceding the optional parameters (if
+    // there are any).
+    FormalParameter? lastRequired;
+    for (var i = 0; i < node.parameters.length; i++) {
+      if (node.parameters[i] is DefaultFormalParameter) {
+        if (i > 0) lastRequired = node.parameters[i - 1];
+        break;
+      }
+    }
+
+    // If all parameters are optional, put the `[` or `{` right after `(`.
+    var builder = DelimitedListBuilder(this);
+    if (node.parameters.isNotEmpty &&
+        node.parameters.first is DefaultFormalParameter) {
+      builder.leftBracket(node.leftParenthesis, delimiter: node.leftDelimiter);
+    } else {
+      builder.leftBracket(node.leftParenthesis);
+    }
+
+    for (var parameter in node.parameters) {
+      builder.add(parameter);
+
+      // If the optional parameters start after this one, put the delimiter
+      // at the end of its line.
+      if (parameter == lastRequired) builder.leftDelimiter(node.leftDelimiter!);
+    }
+
+    builder.rightBracket(node.rightParenthesis, delimiter: node.rightDelimiter);
+    writer.push(builder.build());
   }
 
   @override
@@ -432,12 +464,15 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFunctionTypedFormalParameter(FunctionTypedFormalParameter node) {
-    throw UnimplementedError();
+    startFormalParameter(node);
+    createFunctionType(node.returnType, node.name, node.typeParameters,
+        node.parameters, node.question);
   }
 
   @override
   void visitGenericFunctionType(GenericFunctionType node) {
-    throw UnimplementedError();
+    createFunctionType(node.returnType, node.functionKeyword,
+        node.typeParameters, node.parameters, node.question);
   }
 
   @override
@@ -586,15 +621,8 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
     visit(node.methodName);
     visit(node.typeArguments);
 
-    var builder = DelimitedListBuilder(this);
-    builder.leftBracket(node.argumentList.leftParenthesis);
-
-    for (var argument in node.argumentList.arguments) {
-      builder.add(argument);
-    }
-
-    builder.rightBracket(node.argumentList.rightParenthesis);
-    writer.push(builder.build());
+    createDelimited(node.argumentList.leftParenthesis,
+        node.argumentList.arguments, node.argumentList.rightParenthesis);
   }
 
   @override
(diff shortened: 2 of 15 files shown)


lib/src/string_compare.dart
```diff
@@ -2,13 +2,20 @@
...
 /// This mostly follows the same rules as `String.trim()` because that's what
-/// dart_style uses to trim trailing whitespace as well as considering `,` to
-/// be a whitespace character since the formatter will add and remove trailing
-/// commas.
+/// dart_style uses to trim trailing whitespace.
+///
+/// This function treats `,` as a whitespace character since the formatter will
+/// add and remove trailing commas. It treats, `[`, `]`, `{`, and `}` as
+/// whitespace characters because the formatter may move a comment if it
+/// appears near the closing delimiter of an optional parameter section.
```
lrhn, line 10: Consider having a separate stack of open `[{`s and check that they all occur in both strings, in the correct nesting, even if we don't know that they occur in the correct places.

Or considering recognizing comments instead of `[]{}`, and ignoring the position of them, while checking that they still all occur, in the same order. Seems less error-prone than ignoring order of significan syntax.

munificent: I agree this function is getting weirder. Added a TODO comment to look into it more. There is some tension in that:

* This function exists entirely as a sanity check for the formatter itself. It is only useful when there are bugs in the formatter: it can sometimes prevent user code from getting discarded. This *has* occurred several times in the wild, so it's a useful sanity check. But 99.999% of the time, it doesn't add any user value.

* It is fairly heavy in terms of performance. It walks every character of the original and formatted output, so it adds a noticeable amount of overhead to every run of the formatter.

Finding the right balance between a sanity check that does catch bugs and one that doesn't add too much overhead is hard.

lib/src/front_end/piece_factory.dart
```diff
@@ -342,6 +373,15 @@ mixin PieceFactory implements CommentWriter {
     writer.push(InfixPiece(operands));
   }
 
+  /// Writes the parts of a formal parameter shared by all formal parameter
+  /// types: metadata, `coviariant`, etc.
```
kallentu, line 377: Nit: Typo on 'covariant'.

lib/src/front_end/comment_writer.dart
```diff
@@ -263,7 +263,7 @@ class CommentSequence extends ListBase<SourceComment> {
     // Doc comments and non-inline `/* ... */` comments are always pushed to
```
kallentu, line 263: Nit: It would be nice if the comment for this was also updated with the condition changed. It seems to be like the complement of the condition now.

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -31,7 +31,7 @@ class DelimitedListBuilder {
 
   late final Piece _rightBracket;
 
-  bool _trailingComma = true;
+  bool _isTypeList = false;
```
kallentu, line 34: Can we also add a little comment here like we have for `ListPiece`?

test/type/function_comment.stmt
```diff
@@ -0,0 +1,202 @@
+40 columns                              |
+>>> Block comment before first parameter.
```
kallentu, line 2: Do we have tests for inline block comments?
(And less exciting, but doc comments also)

lib/src/piece/list.dart
```diff
@@ -11,8 +11,17 @@ import 'piece.dart';
...
+/// Usually constructed using [createDelimited()] or a [DelimitedListBuilder].
 class ListPiece extends Piece {
+  /// State used for a split list containing type arguments or type parameters.
+  ///
+  /// Has a higher cost than [State.split] since splitting type arguments and
+  /// type parameters tends to look worse than splitting at other places.
+  // TODO(rnystrom): Having to use a different state for this is a little
+  // cumbersome. Maybe it would be better to most costs out of [State] and
+  // instead have the [Solver] ask each [Piece] for the cost of its state.
+  static const _splitTypes = State(1, cost: 2);
```
kallentu, line 23: Would it make more sense for this to be in `State` instead of `ListPiece`? To keep it together.
If not, maybe name it `_splitTypesState`?

It seems interesting that cost has been just a number to be assigned to a `State` haha. Maybe there's some other reasoning I don't understand. Would you have to update this if you're adding more states with different costs in between other types of states.

kallentu (approved):

kallentu:
Forgot to say :D 
Thank you for the comments and documentation everywhere. It's the only way I can understand all of this right now lol 🚀

lib/src/back_end/code_writer.dart
```diff
@@ -186,7 +190,12 @@ class CodeWriter {
     // to be used as the key in the memoization table.
     piece.format(this, state);
 
-    _pieceOptions.removeLast();
+    var childOptions = _pieceOptions.removeLast();
+
+    // If the child [piece] contains a newline then this one transitively does.
+    // TODO(tall): At some point, we may want to provide an API so that pieces
+    // can block this from propagating outward.
```
natebosch, line 197: Why would a piece want to block the propagation?

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -357,7 +361,35 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFormalParameterList(FormalParameterList node) {
-    throw UnimplementedError();
+    // Find the parameter immediately preceding the optional parameters (if
+    // there are any).
+    FormalParameter? lastRequired;
```
natebosch, line 366: [optional] Consider using the index instead

```dart
    var lastRequired =
        node.parameters.indexWhere((p) => p is DefaultFormalParameter) - 1;

    // ...

    for (var i = 0; i < node.parameters.length; i++) {
      builder.add(node.parameters[i]);

      // If the optional parameters start after this one, put the delimiter
      // at the end of its line.
      if (i == lastRequired) builder.leftDelimiter(node.leftDelimiter!);
    }
```

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -83,27 +123,68 @@ class DelimitedListBuilder {
...
       leading: leadingComments
-    ) = _splitCommaComments(commentsBeforeElement,
-        hasElementAfter: hasElementAfter);
+    ) = _splitCommaComments(comments, hasElementAfter: hasElementAfter);
 
-    // Add any hanging comments to the previous argument.
+    // Add any hanging inline block comments to the previous element before the
+    // subsequent ",".
+    if (inlineComments.isNotEmpty) {
+      for (var comment in inlineComments) {
```
natebosch, line 180: The for-in loop should already handle `isEmpty` input. Can we avoid the extra condition and indentation?

natebosch (approved):

lib/src/front_end/comment_writer.dart
munificent: Documented both.

test/type/function_comment.stmt
munificent: These ones are all inline block comments. Updated the descriptions to match and added tests of non-inline block and doc comments. (Deciding how comprehensive to be in tests is always a challenge.)

lib/src/piece/list.dart
munificent: > Would it make more sense for this to be in `State` instead of `ListPiece`? To keep it together.
> If not, maybe name it `_splitTypesState`?

I'm honestly not sure. I've gone back and forth on this question in my mind. If you look in the old IR, the costs used by all the different rules are [in one file](https://github.com/dart-lang/dart_style/blob/main/lib/src/constants.dart). That can help you see how the various constructs have costs relative to each other. On the other hand, it means the costs are not as encapsulated in the rules that actually use them.

I'm not sure what the right trade-off is yet. In practice, it probably doesn't matter too much one way or the other. One nice thing about keeping the piece-specific states in their respective pieces is that I think it makes it easier to document them because the reader already understands the kind of syntax the state is being used to format.

> It seems interesting that cost has been just a number to be assigned to a `State` haha.

I'm not sold on that design either. In the old IR, each Rule (which is roughly analogous to a Piece) [knows its cost](https://github.com/dart-lang/dart_style/blob/main/lib/src/rule/rule.dart#L47). I'm considering moving cost out of State and instead having a method on Piece where you can ask it "what should the cost of this state be":

```dart
int stateCost(State state);
```

That would give pieces more fine-grained control to choose different costs for states that are used by multiple pieces. Also, it would eliminate the need to create different states that represent conceptually the same thing but just with a different cost.

None of this is carved in stone yet. :)

> Would you have to update this if you're adding more states with different costs in between other types of states.

Costs don't have to be unique, so that's fine. We rely on distinguishable states having distinct `_value`s for that.

lib/src/back_end/code_writer.dart
munificent: I don't know if we'll need it, but the old IR has something similar to handle cases like:

```dart
function(argument, () {
  closure;
});
```
Usually, a newline inside an argument forces the argument list to split. But in this case, because it's a "block argument" we want to allow the newlines in the closure to *not* cause the surrounding argument list to split.

The new Piece architecture will, I think, have a better way of handling block arguments so we probably won't need to block propagation for this specific case, but there may be other places in the language where it's useful. The comment is basically a note to my future self in case it ends up being useful. Once more of the language is supported, if it turns out we didn't need it, I'll just remove the comment.

lib/src/front_end/ast_node_visitor.dart
munificent: Done, with some tweaks.

lib/src/front_end/delimited_list_builder.dart
munificent: Good catch, yes. (This was copied from the subsequent section for `hangingComments` which has some extra code outside of the loop.)

PR #1290, opened 2023-10-19
Turn top-library comment into non-dartdoc.

lrhn:
Prepare for lints 3.0.0 (PR #1289), which will report this as a dangling library comment, when it's probably just not intended a doc comment at all.

### DIFF
--- a/lib/src/debug.dart
+++ b/lib/src/debug.dart
@@ -2,7 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// Internal debugging utilities.
+/// Internal debugging utilities.
+library;
+
 import 'dart:math' as math;
 
 import 'chunk.dart';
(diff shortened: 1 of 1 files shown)


munificent:
It's not the most useful comment in the world, but it is intended to document the purpose of that library. Why is it becoming a lint violation?

lib/src/debug.dart
```diff
@@ -2,7 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// Internal debugging utilities.
+// Internal debugging utilities.
 import 'dart:math' as math;
```
natebosch, line 6:
```suggestion
/// Internal debugging utilities.
library;

import 'dart:math' as math;
```

natebosch (commented):
> Why is it becoming a lint violation?

Library level doc comments are expected to be on a `library;` directive. Using a doc comment on the first `import` (or other directive) is no longer recommended.

lrhn:
What Nate says, and I saw this was the only thing blocking #1289.

munificent (approved):

PR #1291, opened 2023-10-20
Support eagerly pinning a piece to a specific state.

munificent:
The BlockPiece had some bespoke support for being in an "always split" configuration that overrode the two different ways it could be formatted. While working on conditional expressions, I found myself needing something similar to InfixPiece.

Instead of adding similar code to that and likely other pieces in the future, I figured it made more sense to bake support for locking a piece into a specific state directly in the Piece base class.

### DIFF
--- a/lib/src/back_end/solution.dart
+++ b/lib/src/back_end/solution.dart
@@ -26,7 +26,7 @@ class PieceStateSet {
   PieceStateSet._(this._pieces, this._pieceStates);
 
   /// The state this solution selects for [piece].
-  State pieceState(Piece piece) => _pieceStates[piece] ?? State.initial;
+  State pieceState(Piece piece) => _pieceStates[piece] ?? piece.defaultState;
 
   /// Gets the first piece that doesn't have a state selected yet, or `null` if
   /// all pieces have selected states.
@@ -85,11 +85,18 @@ class Solution implements Comparable<Solution> {
     var piece = _state.firstUnsolved();
     if (piece == null) return const [];
 
-    return [
-      // All pieces support a default state.
-      Solution(root, pageWidth, _state.cloneWith(piece, State.initial)),
+    var states = [
+      if (piece.pinnedState case var state?)
+        state
+      else ...[
+        // All pieces support a default state.
+        State.initial,
+        ...piece.states,
+      ]
+    ];
 
-      for (var state in piece.states)
+    return [
+      for (var state in states)
         Solution(root, pageWidth, _state.cloneWith(piece, state))
     ];
   }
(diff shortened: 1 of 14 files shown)


lib/src/back_end/solution.dart
```diff
@@ -85,11 +85,18 @@ class Solution implements Comparable<Solution> {
...
     if (piece == null) return const [];
 
-    return [
-      // All pieces support a default state.
-      Solution(root, pageWidth, _state.cloneWith(piece, State.initial)),
+    var states = [
+      if (piece.pinnedState case var state?)
+        state
+      else ...[
+        // All pieces support a default state.
```
natebosch, line 92: I think it would be OK to ship this solution, but it feels a little off to me, and I think this comment highlights why. Is it fair to say that this comment should read "All _non-pinned pieces_ support a default state"?

It's also interesting that we _aren't_ reading `piece.defaultState` - instead we bake in the assumption that all non-pinned pieces have a default state of `State.initial`. The coupling seems tighter than it needs to be - and I think `Piece` could own more of this knowledge.

What would you think of changing `Piece.states` to be the API used here, and subclasses override something else like `stateOptions`?

```dart
class Piece {
  List<State> get states => _pinnedState ?? [State.initial, ...stateOptions];

  List<State> get stateOptions;

  List<State>? _pinnedState;

  void pin(State state) {
    _pinnedState = [state];
  }
}
```

lib/src/piece/block.dart
```diff
@@ -19,24 +19,19 @@ class BlockPiece extends Piece {
   /// The closing delimiter.
   final Piece rightBracket;
 
-  /// Whether the block should always split its contents.
-  ///
-  /// True for most blocks, but false for enums and blocks containing only
-  /// inline block comments.
```
kallentu, line 25: Can we still retain this comment information somehow by moving the comment somewhere with constructor?

lib/src/back_end/solution.dart
kallentu: Maybe this is along the same lines as what Nate is talking about, but it does seem a little confusing to me how pinned states and `states` itself is separate in `BlockPiece`. It makes sense to me that a pinned state would be a part of `states` like in Nate's example.

kallentu (approved):

lib/src/back_end/solution.dart
munificent: Yeah, I agree this API is getting hairy. Having every Piece implicitly support `State.initial` (whose meaning is somewhat piece dependent) feels kind of hacky too. It avoids having to copy/paste `State.initial` (or something similar) in every piece subclass, but I'm not sure that's really that useful.

I took a stab at doing something similar to what you suggested and I think it looks better. PTAL.

lib/src/piece/block.dart
munificent: Oops, didn't mean to lose that. Done.

PR #1292, opened 2023-10-20
Format conditional expressions.

munificent:
Also, add a test for variable splitting to cover a corner case that was missed.

### DIFF
--- a/test/expression/condition_comment.stmt
+++ b/test/expression/condition_comment.stmt
@@ -1,0 +1,62 @@
+40 columns                              |
+>>> Inline block comment before `?`.
+cond /* c */ ? 1 : 2;
+<<<
+cond /* c */ ? 1 : 2;
+>>> Inline block comment after `?`.
+cond ? /* c */ 1 : 2;
+<<<
+cond ? /* c */ 1 : 2;
+>>> Inline block comment before `:`.
+cond ? 1 /* c */ : 2;
+<<<
+cond ? 1 /* c */ : 2;
+>>> Inline block comment after `:`.
+cond ? 1 : /* c */ 2;
+<<<
+cond ? 1 : /* c */ 2;
+>>> Inline block comment after else.
+cond ? 1 : 2 /* c */;
+<<<
+cond ? 1 : 2 /* c */;
+>>> Line comment before `?`.
+cond // c
+? 1 : 2;
+<<<
+cond // c
+    ? 1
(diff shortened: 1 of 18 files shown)


test/statement/variable.stmt
```diff
@@ -193,4 +193,14 @@ var variableName = ([
...
+>>> Split all variables if an initializer has a split internally.
+var a = 1, b = [element, element, element, element];
+<<<
+var a = 1,
+    b = [
+      element,
+      element,
+      element,
+      element,
+    ];
```
kallentu, line 206: Nice! Sometimes the formatting result is so pleasant to look at compared to the original.

test/expression/condition_comment.stmt
```diff
@@ -0,0 +1,62 @@
...
+    ? 1 // c
+    : 2;
+>>> Line comment after `:`.
+cond ? 1 : // c
+2;
+<<<
+### Looks weird, but users generally won't put comments here.
+cond
+    ? 1
+    : // c
```
kallentu, line 52: Just me being picky, but does this make more sense for the comment to go with the 1 and maybe same with some of the other examples? I guess you don't do comment moving for conditionals?

kallentu (approved):

test/statement/variable.stmt
munificent: The really cool one for me is:

```dart
var longVariableName =
    veryLongFunctionName(
      argument,
      another,
    );
```

The old formatter isn't able to produce that output at all. Instead, it gives you:

```dart
var longVariableName =
    veryLongFunctionName(
  argument,
  another,
);
```

Because it isn't able to change the indentation of subsequent lines based on how it decided to split a previous construct.

test/expression/condition_comment.stmt
munificent: Yeah, I just didn't want to bother moving comments around for conditionals. Usually, it's easiest to just leave the comment where it is and the user will figure it out and move it themselves.

Lists and related constructs get more sophisticated comment handling since they really need it because the formatter may be adding and removing commas.

Eventually, I could see us doing more comment normalization and making stuff nicer like here, but for now I'm trying to not do too many style changes at the same time.

test/statement/variable.stmt
kallentu: That's awesome. Love to see it!

PR #1294, opened 2023-10-24
Format switch statements and expressions.

munificent:
Switch statements build on the existing SequencePiece but need some additional flexibility in there to handle the fact that switch case lines are indented +2 and then case body lines are indented +4.

Switch expressions build on the existing ListPiece since the body is sort of like a collection literal. There are a few tweaks needed here too since an unsplit switch expression body gets spaces inside the curly brackets.

Also, switch values (the part in the leading "( ... )") are formatted using ListPiece as well, since they are formatted essentially like a single-element argument list, except that they don't get a trailing comma.

This does not implement all of the kinds of patterns that can appear in cases. It just implements basic constants needed for the tests. It also doesn't support guard clauses yet. Those will come later.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -959,7 +959,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
     createSwitchValue(node.switchKeyword, node.leftParenthesis, node.expression,
         node.rightParenthesis);
 
-    // Attach the ` {` after the `(` in the [ListPiece] created by
+    // Attach the ` {` after the `)` in the [ListPiece] created by
     // [createSwitchValue()].
     writer.space();
     token(node.leftBracket);
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -5,7 +5,9 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/source/line_info.dart';
 
+import '../constants.dart';
 import '../dart_formatter.dart';
+import '../piece/block.dart';
 import '../piece/do_while.dart';
 import '../piece/if.dart';
 import '../piece/infix.dart';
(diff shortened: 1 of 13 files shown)


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -927,17 +930,87 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+    finishAssignment(node.arrow, node.expression);
   }
 
   @override
   void visitSwitchStatement(SwitchStatement node) {
-    throw UnimplementedError();
+    createSwitchValue(node.switchKeyword, node.leftParenthesis, node.expression,
+        node.rightParenthesis);
+
+    // Attach the ` {` as part of the `)` in the switch value ListPiece.
```
natebosch, line 962: I don't understand the mention of  `ListPiece`. Can you expand this?

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -31,24 +31,80 @@ class DelimitedListBuilder {
...
+  /// This is true for most lists but false for type parameters, type arguments,
+  /// and switch values.
+  final bool _trailingComma;
+
+  /// The cost of splitting this list. Normally 1, but higher for some lists
+  /// that look worse when split.
+  final int _splitCost;
+
+  /// Whether this list should have spaces inside the bracket when it doesn't
+  /// split. This is false for most lists, but true for switch expression
```
natebosch, line 45: [nit] Blank line after header.

lib/src/piece/list.dart
```diff
@@ -74,30 +121,31 @@ class ListPiece extends Piece {
...
-        // Each argument on its own line with a trailing comma after the last.
-        writer.newline(indent: Indent.block);
-        for (var i = 0; i < _arguments.length; i++) {
-          var argument = _arguments[i];
-          argument.format(writer,
-              omitComma: _isTypeList && i == _arguments.length - 1);
-          if (i < _arguments.length - 1) {
-            writer.newline(blank: _blanksAfter.contains(argument));
-          }
+        // Don't write a trailing comma.
```
natebosch, line 132: This has been stated in a comment - does it need to be repeated?

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: Done.

lib/src/front_end/piece_factory.dart
```diff
@@ -373,6 +355,44 @@ mixin PieceFactory implements CommentWriter {
     writer.push(InfixPiece(operands));
   }
 
+  /// Creates a [ListPiece] for the given bracket-delimited set of elements.
+  void createList(
```
kallentu, line 359: I like this name a lot better than before.

kallentu (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -927,17 +930,88 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+    finishAssignment(node.arrow, node.expression);
   }
 
   @override
   void visitSwitchStatement(SwitchStatement node) {
-    throw UnimplementedError();
+    createSwitchValue(node.switchKeyword, node.leftParenthesis, node.expression,
+        node.rightParenthesis);
+
+    // Attach the ` {` after the `(` in the [ListPiece] created by
```
natebosch, line 962: The first part made more sense when it followed a closing paren.

```suggestion
    // Attach the ` {` after the `)` in the [ListPiece] created by
```

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: Oops, thanks for noticing that.

PR #1295, opened 2023-10-26
Refactor ListPiece and DelimitedListBuilder.

munificent:
These classes started out just supporting argument lists, then grew to collection literals, then parameter lists, then switch expressions, and soon for loop headers.

Each time, I need to tweak how it works a little. (But not enough to not use it completely, since the comment handling is complex and is the same across all of those constructors.)

There was getting to be an unwieldy set of options being passed around. Cleaned that up by defining a separate ListStyle that contains those options and can be shuttled directly from DelimitedListBuilder to ListPiece.

I made the before and after bracket pieces optional since I'll be using ListPiece in a future PR for the updaters list in a for loop, which isn't delimited.

Did a bunch of other minor clean-up.

Also, many of the docs were out of date. Fixed those.

### DIFF
--- a/lib/src/front_end/delimited_list_builder.dart
+++ b/lib/src/front_end/delimited_list_builder.dart
@@ -20,7 +20,8 @@ import 'piece_factory.dart';
 class DelimitedListBuilder {
   final PieceFactory _visitor;
 
-  late final Piece _leftBracket;
+  /// The opening bracket before the elements, if any.
+  Piece? _leftBracket;
 
   /// The list of elements in the list.
   final List<ListElement> _elements = [];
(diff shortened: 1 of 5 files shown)


natebosch:
> I made the before and after bracket pieces optional since for

Can you complete this thought in the commit message when you merge?

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -163,24 +102,40 @@ class DelimitedListBuilder {
...
+    _addComments(commentsBeforeElement, hasElementAfter: true);
+  }
+
   /// Adds [element] to the built list.
   ///
   /// Includes any comments that appear before element. Also includes the
   /// subsequent comma, if any, and any comments that precede the comma.
-  void add(AstNode element) {
+  void visit(AstNode element) {
     // Handle comments between the preceding element and this one.
```
natebosch, line 129: This comment feels redundant against the clearly named method on the next line, and against the copied one inside the body of that method. We're in "repeat what the code does" territory with these comments (arguably with both copies).

Consider taking a pass find comments that repeat the code - noise like this makes the useful signal harder to find.

https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html

natebosch (approved):

lib/src/piece/list.dart
```diff
@@ -245,3 +206,91 @@ final class ListElement {
...
+/// be either all on one line or can be each split to their own line with no
+/// extra indentation: argument lists, parameter lists, collection literals,
+/// type arguments, switch expression cases, etc.
+///
+/// These have similar enough formatting to use the same class. And, in
+/// particular, they all handle comments between elements the same way. But
+/// they vary in whether or not a trailing comma is allowed, whether there
+/// should be spaces inside the delimiters when the elements aren't split, etc.
+/// This class captures those options.
+class ListStyle {
```
kallentu, line 236: Love this! Stream-lines the styles a little bit better.

kallentu (approved):

PR #1298, opened 2023-10-27
Revamp the PieceWriter API some.

munificent:
The current API is pretty confusing around when you should use `pop()`, `split()`, or both. If you get it wrong, sometimes it silently works but may leave a bug for later, or sometimes it fails it seemingly unrelated ways. Also, it's not clear whether you should call `pop()` first or `split()`. (It actually didn't matter.)

I tried to simplify the API some. It's still more subtle and brittle than I would like, but it's not clear to me if the complexity is essential. I'm trying to build Piece trees that don't have unnecessary nodes, so some slippage between AST node boundaries and Piece tree boundaries is essential.

I hope this API is better. The changes are:

- Make `split()` implicitly `pop()` the previous piece. This means you never need to call both `split()` and `pop()`. Just call `split()`. You do still have to know when to prefer `pop()` instead of `split()`, but I added some documentation to try to clarify that.

- Rename the main PieceWriter field from `writer` to `pieces`. I think this reads a little better at the callsites and makes it clearer what is being pushed, popped, and split. "Writer" doesn't really mean much.

- Rename `space()` to `writeSpace()`. Then add a helper method in PieceFactory called `space()`. This way, almost all calls on `pieces` deal with pieces and calls that recurse into or write pieces of an AST are bare function calls: `token()`, `visit()`, and `space()`.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -60,8 +60,7 @@ mixin PieceFactory implements CommentWriter {
   /// ```
   void createBlock(Block block, {bool forceSplit = false}) {
     token(block.leftBracket);
-    var leftBracketPiece = writer.pop();
-    writer.split();
+    var leftBracketPiece = pieces.split();
 
     var sequence = SequenceBuilder(this);
     for (var node in block.statements) {
@@ -72,9 +71,9 @@ mixin PieceFactory implements CommentWriter {
     sequence.addCommentsBefore(block.rightBracket);
 
     token(block.rightBracket);
-    var rightBracketPiece = writer.pop();
+    var rightBracketPiece = pieces.pop();
 
-    writer.push(BlockPiece(
+    pieces.push(BlockPiece(
         leftBracketPiece, sequence.build(), rightBracketPiece,
         alwaysSplit: forceSplit || block.statements.isNotEmpty));
   }
@@ -83,7 +82,7 @@ mixin PieceFactory implements CommentWriter {
   void createBreak(Token keyword, SimpleIdentifier? label, Token semicolon) {
     token(keyword);
     if (label != null) {
-      writer.space();
+      space();
       visit(label);
     }
     token(semicolon);
@@ -141,8 +140,7 @@ mixin PieceFactory implements CommentWriter {
     Piece? returnTypePiece;
     if (returnType != null) {
       visit(returnType);
-      returnTypePiece = writer.pop();
-      writer.split();
+      returnTypePiece = pieces.split();
     }
 
     token(functionKeywordOrName);
@@ -152,8 +150,8 @@ mixin PieceFactory implements CommentWriter {
 
     // Allow splitting after the return type.
     if (returnTypePiece != null) {
-      var parametersPiece = writer.pop();
-      writer.push(FunctionTypePiece(returnTypePiece, parametersPiece));
+      var parametersPiece = pieces.pop();
+      pieces.push(FunctionTypePiece(returnTypePiece, parametersPiece));
     }
   }
 
@@ -166,12 +164,11 @@ mixin PieceFactory implements CommentWriter {
     // chain handled by a single [IfPiece].
     void traverse(IfStatement node) {
       token(node.ifKeyword);
-      writer.space();
+      space();
       token(node.leftParenthesis);
       visit(node.expression);
       token(node.rightParenthesis);
-      var condition = writer.pop();
-      writer.split();
+      var condition = pieces.split();
 
       // Edge case: When the then branch is a block and there is an else clause
       // after it, we want to force the block to split even if empty, like:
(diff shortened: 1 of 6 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -189,83 +186,80 @@ mixin PieceFactory implements CommentWriter {
...
           // the last section just being the `else` keyword.
           token(node.elseKeyword);
-          var header = writer.pop();
-          writer.split();
+          var header = pieces.split();
 
           visit(elseStatement);
-          var statement = writer.pop();
-          writer.split();
+          var statement = pieces.pop();
```
natebosch, line 207: Lost a call to `split()` here. Did any tests fail? Any tests we need to add?

```suggestion
          var statement = pieces.split();
```

lib/src/front_end/piece_writer.dart
```diff
@@ -98,33 +167,38 @@ class PieceWriter {
     return _currentText!;
   }
 
-  /// Ends the current text piece and (lazily) begins a new one.
+  /// Pops the previous piece and begins a new one.
```
natebosch, line 170: [nit] Between the two docs we describe the same piece as "the previous piece" and "the most recent piece". I think the latter makes the most sense.

```suggestion
  /// Pops the most recent piece and begins a new one.
```

lib/src/front_end/piece_writer.dart
```diff
@@ -76,19 +133,31 @@ class PieceWriter {
   PieceWriter(this._formatter, this._source);
 
   /// Gives the builder a newly completed [piece], to be taken by a later call
-  /// to [pop] from some surrounding piece.
+  /// to [pop()] or [split()] from some surrounding piece.
   void push(Piece piece) {
```
natebosch, line 137: I wonder if there are any other names that would convey that this is a single slot. Maybe something like `offer` and `take`?

The `push` and `pop` are too engrained with collections, and I'm finding it hard to reason about this CL with those wrong interpretations so accessible. Some of the code changes look like bugs if it behaved like a collection.

lib/src/front_end/piece_factory.dart
munificent: For the trailing chunk of code owned by some AST node, it often doesn't matter whether you capture it with `pop()` or `split()`. The distinction is only visible if some *surrounding* AST node wants to append more code directly that trailing node.

That's very common for expressions (think the `;` in an expression statement). But it's less common for statements.

In this case, both `pop()` and `split()` pass all the tests, and I can't offhand think of any place you could stuff an if statement where the distinction would matter. I picked `pop()` because that's the "friendliest" default way to grab the last chunk of code for a node: it lets the caller decide whether or not it wants to append more code to it.

lib/src/front_end/piece_writer.dart
munificent: Ugh, yeah. I've agonized over what to call them. I did have `give()` and `take()` for a while. I tried using `end()` for `pop()`, but that felt weird because there wasn't really a corresponding `begin()`. I considered calling `split()` `begin()`, but then it's weird that it returns the *previous* piece and not the piece it just begun.

Maybe `send()` and `receive()`?

Naming is hard. @kallentu, what do you think of `give()` for `push()` and `take()` for `pop()`?

kallentu: I thought Nate's `shelve` and `unshelve` kinda made sense. It does almost feel like staging a piece for later.

I don't know how I feel about `give` and `take`. I feel like `push` and `pop` almost reads better than those.

munificent: `shelve()` and `unshelve()` feel weird for me because it sounds like you're putting the piece aside for some arbitrary amount of time when the shelved piece really will be the next one yielded up by the PieceWriter (which is why I initially went with `push()` and `pop()`).

Maybe `add()` for `push()` and `complete()` for `pop()`?

kallentu: An additional thought: I was thinking of `split()` and I was wondering if we could name `split` in a way that references `pop`. It's not clear immediately that `split` also `pop`s. 

`give()` `take()` `takeAndBegin()` => or something? I couldn't make a better example with `add()` and `complete()`. Take with a grain of salt I suppose haha. I just thought it'd help me read the code better.

Naming is hard :/

munificent: OK, I spent some more time mulling over this and trying out a few options. Looking at the callsites and what reads most naturally in the docs, I think `give()`/`take()` are my favorites. I went ahead and changed it to that.

We can always iterate on it later if it doesn't grow on us. But I think this PR overall improves things by having `split()` be a combined operation for splitting and returning the previous piece.

natebosch (approved):

kallentu (approved):

PR #1299, opened 2023-10-27
Update test running code in example script.

munificent:
The example script in example/format.dart has long been my sandbox script for debugging the formatter. One thing it has is a little helper function to run one of the format tests.

That helper code was copy/pasted from an ancient version of the test running code. Cleaned it up here to actually use the TestFile class now that there is some abstraction for running tests.

Also added support to example/format.dart to handle selections in tests.

(I'm starting to work on handling selections in the new formatter.)

### DIFF
--- a/lib/src/testing/test_file.dart
+++ b/lib/src/testing/test_file.dart
@@ -169,6 +169,17 @@ class FormatTest {
   }
 }
 
+extension SourceCodeExtensions on SourceCode {
+  /// If the source code has a selection, returns its text with `‹` and `›`
+  /// inserted at the selection begin and end points.
+  ///
+  /// Otherwise, returns the code as-is.
+  String get textWithSelectionMarkers {
+    if (selectionStart == null) return text;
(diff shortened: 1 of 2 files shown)


lib/src/testing/test_file.dart
```diff
@@ -169,6 +169,17 @@ class FormatTest {
   }
 }
 
+extension SourceCodeExtensions on SourceCode {
```
natebosch, line 172: Why does this live in lib/src/testing instead of the example directly?

natebosch (approved):

lib/src/testing/test_file.dart
munificent: I wanted it to be next to the code in test_file.dart that parses the selection markers out of the string. Also, at some point, it wouldn't hurt if the test code used this to produce better error messages when tests fail because the selection is wrong.

kallentu (approved):

PR #1300, opened 2023-10-31
Get selections running through the new back end.

munificent:
This mostly ports similar code directly over from the old back end, including a couple of quirks in its behavior around whitespace. I considered simplifying that, but I think it's better to have fewer changes during this already huge migration.

Because the new back end adds and removes trailing commas, it treats all commas as whitespace. When a selection is in whitespace, it will be snapped to the nearest non-whitespace code. These two interact in that a selection right around a comma may snap over it because the comma is considered whitespace.

I thought about making it smarter but I didn't want to rathole too long. Instead, I left a comment in CodeWriter.write() with some context and we can revisit it when the overall rewrite is further along. We may end up just deleting that comment later if we decide the behavior here is good enough (I think it is), but this way to remember to think about it.


kallentu (approved):

lib/src/back_end/code_writer.dart
```diff
@@ -59,16 +59,29 @@ class CodeWriter {
   /// The options for the current innermost piece being formatted.
   _PieceOptions get _options => _pieceOptions.last;
 
+  /// The offset in the formatted code where the selection starts, or `null` if
```
natebosch, line 62: This pair of fields makes sense when either can be null independently. These comments make it seem like if one is non-null the other will be as well. Consider changing the docs to say what it means when one is null but not the other.

lib/src/front_end/piece_writer.dart
```diff
@@ -202,11 +229,13 @@ class PieceWriter {
...
     _write(comment.text,
-        containsNewline: comment.containsNewline, hanging: hanging);
+        offset: comment.offset,
+        containsNewline: comment.containsNewline,
+        hanging: hanging);
   }
 
   void _write(String text,
-      {bool containsNewline = false, bool hanging = false}) {
+      {bool containsNewline = false, bool hanging = false, int? offset}) {
```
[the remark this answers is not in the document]
munificent: Good call, done.

lib/src/back_end/code_writer.dart
munificent: Done!

natebosch (approved):

PR #1301, opened 2023-10-31
Format record literals.

kallentu:
Essentially the same piece logic used for other iterables.

- The `TypeLiteral` in `createCollection` has been broken down into its components for record literal re-use.
- Added a new `Comma` style which essentially describes that we need to preserve the comma after each element, regardless of splitting or not.
- Basic record tests.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -664,10 +664,10 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   void visitListLiteral(ListLiteral node) {
     createCollection(
       node.constKeyword,
+      typeArguments: node.typeArguments,
       node.leftBracket,
       node.elements,
       node.rightBracket,
-      typeArguments: node.typeArguments,
     );
   }
 
@@ -865,8 +865,9 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   @override
   void visitRecordLiteral(RecordLiteral node) {
     ListStyle style;
-    if (node.fields.length == 1) {
-      // Single-element records always have a trailing comma.
+    if (node.fields.length == 1 && node.fields[0] is! NamedExpression) {
+      // Single-element records always have a trailing comma, unless the single
+      // element is a named field.
       style = const ListStyle(commas: Commas.alwaysTrailing);
     } else {
       style = const ListStyle(commas: Commas.trailing);
@@ -943,10 +944,10 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   void visitSetOrMapLiteral(SetOrMapLiteral node) {
     createCollection(
       node.constKeyword,
+      typeArguments: node.typeArguments,
       node.leftBracket,
       node.elements,
       node.rightBracket,
-      typeArguments: node.typeArguments,
     );
   }
 
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -662,7 +662,13 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitListLiteral(ListLiteral node) {
-    createCollection(node, node.leftBracket, node.elements, node.rightBracket);
+    createCollection(
+      node.constKeyword,
+      node.leftBracket,
+      node.elements,
+      node.rightBracket,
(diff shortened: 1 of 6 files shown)


kallentu:
@munificent I'll remove the record tests from the temp repo when this lands, in case anything changes.

natebosch (commented):
LGTM. I'll let Bob approve

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -863,7 +869,20 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitRecordLiteral(RecordLiteral node) {
-    throw UnimplementedError();
+    ListStyle style;
+    if (node.fields.length == 1) {
+      // Single-element records always have a trailing comma.
```
munificent, line 874: You'll also need to check that the field is positional. Records with only a single *named* field don't get a trailing comma unless they split:

```dart
var x = (name: value);
```

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -667,7 +667,13 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitListLiteral(ListLiteral node) {
-    createCollection(node, node.leftBracket, node.elements, node.rightBracket);
+    createCollection(
+      node.constKeyword,
+      node.leftBracket,
+      node.elements,
+      node.rightBracket,
+      typeArguments: node.typeArguments,
```
munificent, line 675: Nit: I'd move the type arguments argument up to reflect the order that these appear in the source syntax:

```dart
    createCollection(
      node.constKeyword,
      typeArguments: node.typeArguments,
      node.leftBracket,
      node.elements,
      node.rightBracket,
  );
```

test/expression/record.stmt
```diff
@@ -0,0 +1,183 @@
...
+<<<
+var record = (
+  veryLongRecordField________________,
+);
+>>> Single-element named record.
+var record = (longFieldName: longRecordFieldValue);
+<<<
+var record = (
+  longFieldName: longRecordFieldValue,
+);
```
munificent, line 38: Yeah, this is the test that should change. Here, we don't need a trailing comma and shouldn't insert one.

test/expression/record.stmt
```diff
@@ -0,0 +1,183 @@
...
+    nine,
+    tenth,
+    eleventh,
+  ),
+);
+>>> Trailing comma in single-element does not split.
+(1,);
+<<<
+(1,);
+>>> Trailing comma in multi-element does force split.
```
munificent, line 94: You can just remove this test. It's migrated from the old tests where a trailing comma forces a split, but that's no longer true.

test/expression/record.stmt
```diff
@@ -0,0 +1,183 @@
...
+(1,2,);
+<<<
+(1, 2);
+>>> Ignore line comment after the "]".
```
munificent, line 98: Nit: "]" -> ")".

munificent, line 98: Can you move the comment-related tests to a separate `expression/record_comment.stmt` file?

test/expression/record.stmt
```diff
@@ -0,0 +1,183 @@
...
+);
+<<<
+(
+  element,
+
+  // comment
+  element,
+  element,
+);
+>>> Format on one line in an argument list.
```
munificent, line 132: These tests for how record literals are formatted inside function calls should get moved over to `tests/invocation/`. There's enough special sauce around argument lists and method chains that I made a whole directory for those tests.

Actually, since we don't have block argument formatting implemented yet (see [here](https://github.com/dart-lang/dart_style/blob/main/lib/src/piece/list.dart#L54-L61)). It might make the most sense to just leave these tests of record literals in argument list in the unmigrated repo for now.

munificent (changes requested):
This looks great! There are a few changes here and there, but overall this is looking really good.

test/expression/record.stmt
kallentu: Will leave the ones with function calls out.

kallentu (commented):
@munificent If you could take another look, that'd be great.
(I messed up the commit history because I'm too used to my gerrit commands, but I'll try to avoid that for the next PR 😭 )

munificent (approved):
Looks great! :D

PR #1303, opened 2023-11-01
Format function declarations and expressions.

munificent:
This covers:

- Top-level function declarations
- Local function declaration statements
- Anonymous function expressions

It handles all three kinds of bodies: `;`, `{ ... }`, and `=> ... ;`.

Most of the implementation work for functions is in:

- Their parameter lists, which is most already handled by the existing support (and tests) for function type annotations.
- Their `{ ... }` bodies, which are just block statements and are already implemented and tested.
- Their `=> ...` bodies, which are formatted like assignments so need little additional implementation.

This PR mostly just stitches that together, adds tests, and covers some of the bits that can't appear in those existing constructs:

- The `external` modifier on function declarations.
- Function names.
- Default values.
- The `var` and `final` modifiers on parameters.

This PR does not do getters and setters. That will come later.

### DIFF
--- a/test/function/default_value.unit
+++ b/test/function/default_value.unit
@@ -1,0 +1,48 @@
+40 columns                              |
+>>> Named parameter.
+f(  {  optional  =  null  }  ) {}
+<<<
+f({optional = null}) {}
+>>> Named parameter with old separator.
+### This syntax is no longer supported by new versions of Dart, but we want to
+### support formatting older versions if possible.
+f(  {  optional  :  1  +  2  }  ) {}
+<<<
+f({optional: 1 + 2}) {}
+>>> Optional positional parameter.
+f(  [  optional  =  1  +  2  ]  ) {}
+<<<
+f([optional = 1 + 2]) {}
+>>> Split on positional default.
+doStuff([parameter = veryLongDefaultValueThatSplits, another =
+veryLongDefaultValue, third = alsoQuiteLongDefaultValue]) {}
+<<<
+doStuff([
+  parameter =
+      veryLongDefaultValueThatSplits,
+  another = veryLongDefaultValue,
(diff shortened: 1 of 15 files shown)


test/function/expression_body.unit
```diff
@@ -0,0 +1,71 @@
...
+myFunction() => argument + anotherArgument;
+<<<
+myFunction() =>
+    argument + anotherArgument;
+>>> Split inside expression forces split at `=>`.
+veryLongFunction() => argument + anotherArgument + aThirdArgument;
+<<<
+veryLongFunction() =>
+    argument +
+        anotherArgument +
```
kallentu, line 29: It's a little odd how these don't line up properly.

test/function/default_value.unit
```diff
@@ -0,0 +1,48 @@
...
+doStuff({
+  parameter =
+      veryLongDefaultValueThatSplits,
+  another = veryLongDefaultValue,
+  third = alsoAQuiteLongDefaultValue,
+}) {}
+>>> Prefer block-like splitting for collection default values.
+function([param = [element, element, element, element]]) { body; }
+<<<
+function([
```
kallentu, line 39: 🤌 lovely

kallentu (approved):

test/function/expression_body.unit
munificent: Yeah, that's a fair observation. The old style [handles that case explicitly](https://github.com/dart-lang/dart_style/blob/main/lib/src/source_visitor.dart#L344-L352), at least for the RHS of expression-bodied functions. I'm up for doing that in the new style too. Added a TODO to AssignPiece (which uniformly handles `=>`, `=`, and `:` in the new style).

natebosch (approved):

PR #1307, opened 2023-11-02
Format instance creation expressions (constructor calls).

munificent:
Simple constructor calls behave exactly like normal function calls and use the same formatting code for their argument list.

Prefixed and/or named constructor calls are formatted sort of like method/property chains where they may split at the ".". We don't have full support for method/property chains yet, but this starts sketching that out enough to get the instance creation expression tests passing.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -9,6 +9,7 @@ import 'package:analyzer/source/line_info.dart';
 import '../constants.dart';
 import '../dart_formatter.dart';
 import '../piece/block.dart';
+import '../piece/chain.dart';
 import '../piece/do_while.dart';
 import '../piece/for.dart';
 import '../piece/if.dart';
@@ -279,7 +280,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitConstructorName(ConstructorName node) {
-    throw UnimplementedError();
+    assert(false, 'This node is handled by visitInstanceCreationExpression().');
   }
 
   @override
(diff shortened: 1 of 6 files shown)


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -618,7 +619,38 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
-    throw UnimplementedError();
+    token(node.keyword, after: space);
+
+    // If there is an import prefix and/or constructor name, then allow
+    // splitting before the `.`. This doesn't look good, but is consistent with
+    // constructor calls that don't have `new` or `const`. We allow splitting
+    // in the latter because there is no way to distinguish syntactically
+    // between a named constructor call and any other kind of method call.
+    var calls = <Piece>[];
```
natebosch, line 629: [nit] I wouldn't name the type following an import prefix a "call". WDYT about `memberAccesses` or `memberReads`?

lib/src/piece/chain.dart
```diff
@@ -0,0 +1,52 @@
...
+
+/// A series of property access or method calls, like:
+///
+/// ```
+/// target.getter.method().another.method();
+/// ```
+///
+/// This piece handles splitting before the `.`.
+class ChainPiece extends Piece {
+  /// The series of calls.
```
natebosch, line 21: Similar here. A function tearoff isn't a "call" for instance.

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -618,7 +619,38 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+
+    if (node.constructorName.type.importPrefix case var importPrefix?) {
+      token(importPrefix.name);
+      calls.add(pieces.split());
+      token(importPrefix.period);
+    }
+
+    token(node.constructorName.type.name2);
```
kallentu, line 637: Nit: Maybe make a local for `node.constructorName.type` to make it easier to read this section?

kallentu (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: Those are both a little too long and too specific for my taste. Eventually, I expect this code to get a decent amount of complexity, based on how much complexity there is in the prototype and the old style around method call chains. So I want a fairly terse and generic term for the things with a chain.

I changed it to "operation". I think that is suitably generic.

lib/src/front_end/ast_node_visitor.dart
munificent: Good call. Did the same for `node.constructorName` too.

PR #1309, opened 2023-11-02
Format class declarations.

munificent:
This handles all of the stuff that can appear outside of the braces in a class declaration:

- The `class` keyword and any class modifiers.
- The name.
- Type parameters, if any.
- The `extends`, `with`, and/or `implements` clauses.
- The `native` clause.

It doesn't handle members yet.

It also handles ensuring there is a blank line before and after class declarations. In the process of doing that, I also migrated over similar code for function declarations where we ensure there is a blank line after non-empty functions.


lib/src/piece/type.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../back_end/code_writer.dart';
+import 'clause.dart';
+import 'piece.dart';
+
+/// Piece for a type declaration with a body containing members.
```
natebosch, line 8: Can you expand this, maybe with a list of some of the constructs this covers?

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -74,8 +74,29 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+      var needsBlank = true;
       for (var declaration in node.declarations) {
+        // Add a blank line before types with bodies.
+        var hasBody = declaration is ClassDeclaration ||
+            declaration is EnumDeclaration ||
+            declaration is ExtensionDeclaration;
+
+        if (hasBody) needsBlank = true;
+
+        if (needsBlank) sequence.addBlank();
```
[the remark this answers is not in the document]
munificent: Good catch. This was ported over from the old style which isn't quite as resilient about writing blank lines. I managed to simplify it a bunch more thanks to how SequenceBuilder works.

lib/src/piece/type.dart
munificent: Done. I'm tempted to actually delete this Piece entirely. It doesn't really *do* anything except concatenate a few pieces. The only reason I'm not using AdjacentPiece is because it needs to insert a space between two of them. I might just add support to AdjacentPiece for that.

But I wanted to keep this class for now in case it becomes more specialized over time as more class/enum/extension features are added.

kallentu (approved):

PR #1310, opened 2023-11-03
Format mixin application classes.

munificent:
This mostly reuses the existing createType() code with a couple of tweaks since mixin application classes don't have bodies and have the "= Superclass" part before the "with" clause.

You'll notice that the modifiers test case here takes a couple of seconds to run. That's because the new back end's line splitter doesn't do any memoization or divide and conquering, so it's brute-forcing the whole compilation unit to figure out which of the mixin applications need splitting.

I don't think it's intolerable yet as far as how long it takes the tests to run, but this means it's probably a good time for me to start looking into optimization soon.


natebosch:
> but this means it's probably a good time for me to start looking into optimization soon.

+1 I would avoid adding more complexity until we know we can optimize what we have

lib/src/front_end/piece_factory.dart
```diff
@@ -408,29 +408,46 @@ mixin PieceFactory implements CommentWriter {
...
-      TypeParameterList? typeParameters,
+  /// Creates a class, enum, extension, mixin, or mixin application class
+  /// declaration.
+  ///
+  /// For all but a mixin application class, [body] should a record containing
+  /// the bracket delimiters and the list of member declarations for the type's
+  /// body.
+  ///
+  /// For mixin application classes, [body] is `null` and instead [equals],
+  /// [superclass], and [semicolon] are provided.
```
natebosch, line 419: can these be split into different calls with more statically useful signatures (that might forward to a shared private implementation)

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: I thought about that, but most of the parameters are valid in both so there would be a lot of pointless forwarding. And the function for handling mixin application classes would only be called once. This seemed the least boilerplate-y to me.

munificent:
> +1 I would avoid adding more complexity until we know we can optimize what we have

I'm basically always worried about performance of the formatter, but in this case, I have a pretty decent confidence that there is a big optimization hammer that will work: memoizing the result of formatting child pieces.

That's what the old formatter does for the contents of BlockChunks and is one of the main parts of the algorithm that makes performance tractable. The new back end and line splitter has always been designed around applying that same optimization. One way to think of the new back end is that's basically "What if we just made *every* Chunk a BlockChunk?" So I hope (but can't promise) that performance will end up better than the old one.

But, yeah, I think there's enough different kinds of pieces in place that now I can start working on the performance side of things.

munificent:
> I have a pretty decent confidence that there is a big optimization hammer that will work: memoizing the result of formatting child pieces.

After spending a bunch of time on Friday working on that optimization, it turns out I was wrong and that isn't the big hammer. The big hammer is [only expanding pieces that help fix problems with the current solution](https://github.com/dart-lang/dart_style/pull/1311).

kallentu (approved):

PR #1311, opened 2023-11-06
Less branchy expanding

munificent:
When expanding, only use the first unsolved piece in the first bad line.

The line splitting solver works by incrementally binding one piece to a state at a time, expanding out a tree of partial solutions until it finds one that fits.

Currently, the solver uses no logic to choose which pieces to bind to states when expanding. It just brute forces them all from beginning to end. This means it spends a bunch of time adding line splits to places that don't need them and generating losing solutions that later get discarded.

Instead, with this change, we take the same optimization that the old line splitter uses: When expanding a solution, we only look at the first solvable piece in the first problematic line.

A "solvable" piece is one that allows multiple states and hasn't already been bound in the current solution. A problematic line is one that overflows the page width or contains an invalid newline.

This optimization lets the solver skip past all of the non-problematic pieces that are in regions of code that already fit.

There will probably be more optimization work to do later, but this gets the tests back down to <1 second.

I split this into a couple of commits to hopefully make it easier to review. The first two just simplify the solver some. There was some unnecessary complexity in there that I wanted to get rid of before adding any more. Then the third commit is the actual optimization.


lib/src/back_end/solution.dart
```diff
@@ -112,8 +154,19 @@ class Solution implements Comparable<Solution> {
   /// code.
   @override
   int compareTo(Solution other) {
-    var scoreComparison = score.compareTo(other.score);
-    if (scoreComparison != 0) return scoreComparison;
+    // TODO(perf): This check is necessary to not make some of the tests not
+    // timeout, but I'm not sure why. Investigate what this does and if it's
+    // still necessary once other optimizations are in place.
+    // An invalid solution is worse than all others.
+    if (!isValid) return 1;
```
natebosch, line 161: Should we try to make this symmetric even when comparing two invalid solutions?

lib/src/back_end/solution.dart
```diff
@@ -69,8 +59,57 @@ class Solution implements Comparable<Solution> {
...
-  /// The score resulting from the selected piece states.
-  final Score score;
+  /// Whether this score is for a valid solution or not.
+  ///
+  /// An invalid solution is one where a hard newline appears in a context
+  /// where splitting isn't allowed. This is considered worse than any other
+  /// solution.
+  final bool isValid;
+
+  /// The number of characters that do not fit inside the page width.
```
natebosch, line 69: [nit]
```suggestion
  /// The total number of characters that do not fit inside the page width.
```

lib/src/back_end/code_writer.dart
```diff
@@ -119,6 +145,19 @@ class CodeWriter {
 
     _buffer.write(text);
     _column += text.length;
+
+    // If we haven't found an overflowing line yet, then this line might be one
+    // so keep track of the pieces we've encountered.
+    if (!_foundExpandLine) {
+      // TODO(perf): This [addAll()] call might be slow. If each Piece had a
```
natebosch, line 152: Which `addAll` call are you referring to?

Also, [nit] avoid using the square braces for non-doc comments

natebosch (approved):

lib/src/back_end/solution.dart
munificent: Actually, it looks like this check is no longer needed at all now that I've optimized it. Removed the validity checks completely.

lib/src/back_end/solution.dart
munificent: Done locally.

lib/src/back_end/code_writer.dart
munificent: Oops! That's a relic of an intermediate implementation that I ended up improving on. Removed.

lib/src/back_end/code_writer.dart
```diff
@@ -54,7 +53,39 @@ class CodeWriter {
   ///
   /// This is used to increase the cumulative nesting as we recurse into pieces
   /// and then unwind that as child pieces are completed.
-  final List<_PieceOptions> _pieceOptions = [_PieceOptions(0, 0, true)];
+  final List<_PieceOptions> _pieceOptions = [_PieceOptions(0, true)];
+
+  /// True if we have already found the first line where whose piece should be
```
kallentu, line 58: Nit: `Whether` instead of `True if` ?

kallentu (approved):
Nice! Removing nesting is kinda nice. Clears up indent vs nesting and how that works.

But also, thanks for making the commits that you did. It helped a lot with reviewing this big PR.

PR #1312, opened 2023-11-06
Format yield and yield*.

kallentu:
Small CL for yield behaviour.
No tests were migrated (but also, none existed for the temp dir lol), just added some simple ones.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -1307,12 +1307,8 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   @override
   void visitYieldStatement(YieldStatement node) {
     token(node.yieldKeyword);
-
-    if (node.star case var star?) {
-      token(star);
-      space();
-    }
-
+    token(node.star);
+    space();
     visit(node.expression);
     token(node.semicolon);
   }
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -1306,7 +1306,15 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitYieldStatement(YieldStatement node) {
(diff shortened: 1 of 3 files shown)


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1306,7 +1306,15 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitYieldStatement(YieldStatement node) {
-    throw UnimplementedError();
+    token(node.yieldKeyword);
+
+    if (node.star case var star?) {
+      token(star);
+      space();
```
munificent, line 1313: Huh, I'm surprised this does the right thing when there's no `*`. I would expect you to end up with `yieldi;` since there's no `space()` call when `star` is `null`.

Any idea where the space is coming from?

I would expect this to be:

```dart
  void visitYieldStatement(YieldStatement node) {
    token(node.yieldKeyword);
    token(node.star);
    space();
    visit(node.expression);
    token(node.semicolon);
  }
```

(The `token()` function will silently do nothing if you pass it `null`, which ends up being pretty handy.)

munificent (approved):

lib/src/front_end/ast_node_visitor.dart
kallentu: Okay, you're right in that this was kinda weird. Turns out `yield i;` on its own just parses as a variable declaration and not a yield statement. The parser needs more context than that to determine that it's a yield.

I've modified the tests and they both should be parsing into yield statements and they should work as intended now.

Thanks for bringing this up and if you could PTAnotherL, that'd be great!

munificent: *OOOH* that makes. I totally forgot that `yield` is a contextual keyword and won't be parsed as a keyword unless it's in a `sync*` or `async*` function.

munificent (approved):
Looks great! Thank your for tracking down what was going on.

PR #1313, opened 2023-11-07
Format fields and methods.

munificent:
Almost all of the real work was already done to handle variables and functions, so this just extends that to fields and methods defined in classes and covers a few loose ends:

- External declarations in their various forms.
- The covariant modifier on parameters and fields.
- Operator declarations (which apparently were never tested on the old style!).
- Getters and setters.
- The `static`, `abstract`, and `late` modifiers.

I also tweaked the cost heuristics around splitting between return types, parameter lists, and `=>` expression bodies. It looks pretty weird to split after a return type, so I made that more costly than splitting at the body or parameter list. That in turn led me to tweaking the cost to split a type parameter list since those also look bad when split.

I also decided to put the tests for member declarations inside declaration/ instead of member/. My impression is that there won't be a huge number of separate files in declaration/ and member/, so it's easier to just merge them together. Let me know what you think about that.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -415,7 +415,12 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFieldDeclaration(FieldDeclaration node) {
-    throw UnimplementedError();
+    modifier(node.externalKeyword);
+    modifier(node.staticKeyword);
+    modifier(node.abstractKeyword);
+    modifier(node.covariantKeyword);
+    visit(node.fields);
+    token(node.semicolon);
   }
 
   @override
@@ -556,19 +561,14 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFunctionDeclaration(FunctionDeclaration node) {
-    modifier(node.externalKeyword);
-
-    Piece? returnType;
-    if (node.returnType case var returnTypeNode?) {
-      visit(returnTypeNode);
-      returnType = pieces.split();
-    }
-
-    // TODO(tall): Get or set keywords for getters and setters.
-    if (node.propertyKeyword != null) throw UnimplementedError();
-    token(node.name);
-
-    finishFunction(returnType, node.functionExpression);
+    createFunction(
+        externalKeyword: node.externalKeyword,
+        returnType: node.returnType,
+        propertyKeyword: node.propertyKeyword,
+        name: node.name,
+        typeParameters: node.functionExpression.typeParameters,
+        parameters: node.functionExpression.parameters,
+        body: node.functionExpression.body);
   }
 
   @override
@@ -578,7 +578,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFunctionExpression(FunctionExpression node) {
-    finishFunction(null, node);
+    finishFunction(null, node.typeParameters, node.parameters, node.body);
   }
 
   @override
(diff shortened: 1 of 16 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -483,7 +515,7 @@ mixin PieceFactory implements CommentWriter {
         leftBracket: leftBracket,
         elements,
         rightBracket: rightBracket,
-        style: const ListStyle(commas: Commas.nonTrailing, splitCost: 2));
+        style: const ListStyle(commas: Commas.nonTrailing, splitCost: 3));
```
natebosch, line 518: Can this change be mentioned in the commit message?

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: It was a knock-on part of: "I also tweaked the cost heuristics around splitting between return types, parameter lists, and => expression bodies." But, yeah, I'll update it to mention that.

test/function/getter.unit
```diff
@@ -0,0 +1,6 @@
+40 columns                              |
+>>> Split before `get`.
+VeryLongTypeAnnotation get veryLongGetter => null;
+<<<
+VeryLongTypeAnnotation
+get veryLongGetter => null;
```
kallentu, line 6: Does this getter not have any indentation when split?

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -415,7 +415,12 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitFieldDeclaration(FieldDeclaration node) {
-    throw UnimplementedError();
+    modifier(node.externalKeyword);
+    modifier(node.staticKeyword);
```
kallentu, line 419: Are we handling `const` modifiers? For static const fields

kallentu (approved):

test/function/getter.unit
munificent: This is a style change I made in the prototype based on what I saw in a few places in the Flutter repo.

It's really rare, but not impossible, for a field or variable declaration to split after the type annotation, or for a function declaration to split after the return type. It mostly happens when you have a really long generic type. When that happens, it's not really clear what the indentation should look like.

Many years ago, I basically just picked when I first implemented it, and decided it would indent the member, like:

```dart
VeryLongTypeAnnotation
    get veryLongGetter => null;
```

I don't think I ever particularly liked this style, and I don't think many people ran into it. When I was tweaking style rules during the prototype, I tried just not indenting there, and I think it looks better, especially in the context of a long parameter list:

```dart
function(
  SomeReallyLongParameterType
  aParameter,
  int anotherParameter,
) {
  ...
}
```

It's definitely up for discussion, though, if you think it looks weird.

lib/src/front_end/ast_node_visitor.dart
munificent: Oh, good catch. It is working correctly (I think `const` uses the same `.keyword` token in `VariableDeclarationList`. Added tests.

test/function/getter.unit
kallentu: I don't have a very strong opinion about this. I just thought it looked different from how I expected a member to look, that's all.

If it's what we've had for a while, let's keep it as it is.

natebosch: I think we should expect some pushback if folks do notice it. It makes the second line look like it could return `dynamic`, but I do agree that it looks nicer, especially looking at the first line of the body against the signature line(s) above.

PR #1314, opened 2023-11-07
Format for-in loops.

munificent:
I ended up refactoring the code for handling the for parts and inlining it all into visitForStatement() because it turns out that there isn't as much sharing between C-style for loops and for-in loops as I was expecting.

Thanks to the magic of pattern matching and shared case bodies, I was still able to get it all into a single flattened switch statement for all the kinds of for parts.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -339,7 +339,9 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
 
   @override
   void visitDeclaredIdentifier(DeclaredIdentifier node) {
-    throw UnimplementedError();
+    modifier(node.keyword);
+    visit(node.type, after: space);
+    token(node.name);
   }
 
   @override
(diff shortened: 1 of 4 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -526,22 +526,34 @@ mixin PieceFactory implements CommentWriter {
...
-  ///
-  /// This method assumes the code to the left of the `=` or `:` has already
+  /// This method assumes the code to the left of the operator has already
   /// been visited.
-  void finishAssignment(Token operator, Expression rightHandSide) {
-    if (operator.type == TokenType.EQ) space();
-    token(operator);
-    var target = pieces.split();
+  void finishAssignment(Token operator, Expression rightHandSide,
+      {bool splitBeforeOperator = false}) {
```
natebosch, line 546: [nit] can you describe the new argument in the doc?

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: Good catch, done.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -476,66 +478,114 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+        // for (var variable in [
+        //   initializer,
+        // ]) {
+        //   body;
+        // }
+        // ```
+        token(node.leftParenthesis);
+        visit(variable);
+
+        finishAssignment(forEachParts.inKeyword, forEachParts.iterable,
```
kallentu, line 580: I'm not a fan of having `finishAssignment` or `AssignPiece` to describe this splitting type. It's kinda confusing to read and I wish we had a better way to describe it like `finishOperatorExpression` or `OperatorExpressionPiece`. Maybe not that, but it seems odd to read it as an assign or an "assignment" style.

kallentu (approved):
Looks great overall :D
Thanks for also adding the await-fors.

lib/src/front_end/ast_node_visitor.dart
munificent: I'm going to go ahead and land this, but I'm definitely up for iterating on names.

PR #1316, opened 2023-11-08
Format await statements.

kallentu:
Even smaller CL to handle await statements.
Await fors are handled in https://github.com/dart-lang/dart_style/pull/1314 which made this even simpler.

Also, there are no other existing tests for `await` because it's honestly too small of a piece. So no tests have been migrated.

I added one test to test the basic await + one for await with a comment.


munificent (approved):
Looks great, thanks!

PR #1317, opened 2023-11-09
Format enums

munificent:
Format enum declarations with and without members.

Enums are a little tricky because they are sort of like lists (i.e. ListPiece) and sort of like type declarations (i.e. BlockPiece). Enums without members should be formatted pretty much like any other delimited list where the declaration can collapse to one line (and discard any trailing comma) if it fits or expand to one element per line (and add a trailing comma) otherwise.

Enums with members are formatted pretty much like blocks where each constant  or member is on its own line.

There are a few wrinkles around trailing commas and the semicolon between the constants and members:

-   If the enum has no members, any trailing `;` is discarded. It doesn't do anything useful and I consider this morally equivalent to discarding other trailing commas.

-   If the enum has members, then a trailing comma after the last constant is redundant with the subsequent `;`. In that case, it discards the comma.

In both of these cases, we have to make sure we don't accidentally lose any comments around that discarded punctuation. So there are a lot of tests for all of those edge cases even though in practice no one writes code like that.

Discarding `;` on enums without members means that we treat `;` like whitespace, so I also loosened equalIgnoringWhitespace() to ignore semicolons too.

Since enums with and without members are formatted using different codepaths, there are a lot of mostly redundant-seeming tests for them.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -114,8 +114,11 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   }
 
   @override
-  void visitArgumentList(ArgumentList node, {bool nestExpression = true}) {
-    throw UnimplementedError();
+  void visitArgumentList(ArgumentList node) {
+    createList(
+        leftBracket: node.leftParenthesis,
+        node.arguments,
+        rightBracket: node.rightParenthesis);
   }
 
   @override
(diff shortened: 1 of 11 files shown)


natebosch:
> I also loosened equalIgnoringWhitespace() to ignore semicolons too.

Does this reduce our ability to catch a missed `token(semicolon);`? Should we make it an optional behavior and only use the looser check when we need it?

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -400,12 +403,81 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
+
+      var sequence = SequenceBuilder(this);
+      for (var constant in node.constants) {
+        sequence.addCommentsBefore(constant.firstNonCommentToken);
+        visit(constant);
+
+        if (constant != node.constants.last) {
+          commaAfter(constant);
+        } else {
+          // Discard the trailing comma if there is one since there is a
```
kallentu, line 449: Yes! Not having `,;` haha

kallentu (approved):

PR #1320, opened 2023-11-13
Format assert statements.

kallentu:
This looks really similar to how asserts were formatted in the old formatter.
It's also similar logic to assert initializers, but we need constructors done first and then I'll handle that after.

I assume we have test logic for comments in ListPieces and general statements so I didn't add any extra for that, but if you think otherwise, let me know!

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -131,15 +131,14 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
   @override
   void visitAssertStatement(AssertStatement node) {
     token(node.assertKeyword);
-
-    var arguments = [node.condition];
-    if (node.message case var message?) arguments.add(message);
     createList(
-      arguments,
+      [
+        node.condition,
+        if (node.message case var message?) message,
+      ],
       leftBracket: node.leftParenthesis,
       rightBracket: node.rightParenthesis,
     );
-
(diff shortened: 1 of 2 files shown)


kallentu:
Oh wait, before reviewing, let me push another commit. I forgot to push this one.

kallentu:
Okay, should be ready for review now!

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -130,7 +130,17 @@ class AstNodeVisitor extends ThrowingAstVisitor<void>
...
-    throw UnimplementedError();
+    token(node.assertKeyword);
+
+    var arguments = [node.condition];
+    if (node.message case var message?) arguments.add(message);
+    createList(
+      arguments,
+      leftBracket: node.leftParenthesis,
+      rightBracket: node.rightParenthesis,
+    );
```
munificent, line 141: If you really wanted to be clever, you could do:

```dart
    createList(
      [
        node.condition,
        if (node.message case var message?) message
      ],
      leftBracket: node.leftParenthesis,
      rightBracket: node.rightParenthesis,
    );
```

munificent (approved):
This is great!

lib/src/front_end/ast_node_visitor.dart
kallentu: Oh that's kinda wild. Done.

PR #1322, opened 2023-11-15
Support "block argument" formatting.

munificent:
Sometimes in an argument list, we don't want to force it to full split even if an argument contains an inner newline. In other words, we want to prefer:

```
test('hi', () {
  body;
});
```

Over:

```
test(
  'hi',
  () {
    body;
  },
);
```

The latter is what you would normally get when an argument contains a newline. This change enables the former. The heuristics for what kinds of arguments are allowed to have this block-like formatting, how many of them, and what other arguments are allowed are pretty fuzzy.

I picked a simple set of rules here (which follow the prototype), but left some TODOs because I suspect we will iterate on them later once we can run the new formatter on larger codebases.

The heuristic for what kinds of arguments are block-like is the same rule that we use for "delimited" expressions after an assignment. I think it's good that those two agree and helps make the style overall more consistent. Basically, any time the right-hand side of an `=` or `:` doesn't need to split, that same kind of expression can be block formatted:

```
var list = [
  element,
];

function([
  element,
]);
```

Block formatting is used for argument lists, asserts, and switch values.

I also simplified AssignPiece. It used an extra unnecessary to allow newlines inside delimited value expressions without splitting after "=".

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -126,17 +126,61 @@ extension ExpressionExtensions on Expression {
   ///   element,
   /// ];
   /// ```
-  bool get isDelimited => switch (this) {
-        FunctionExpression() => true,
-        InstanceCreationExpression() => true,
-        ListLiteral() => true,
-        MethodInvocation() => true,
-        ParenthesizedExpression(:var expression) => expression.isDelimited,
-        RecordLiteral() => true,
-        SetOrMapLiteral() => true,
-        SwitchExpression() => true,
-        _ => false,
-      };
+  ///
+  /// Likewise, in an argument list, delimited expressions can be given
+  /// block-like formatting:
+  ///
+  /// ```
+  /// function([
+  ///   element,
+  /// ]);
+  /// ```
+  bool get isDelimited {
+    // Unwrap named expressions to get the real expression inside.
+    var expression = this;
+    if (expression is NamedExpression) {
+      expression = expression.expression;
+    }
+
+    // TODO(tall): We should also allow multi-line strings to be formatted
+    // like block arguments, at least in some cases like:
+    //
+    // ```
+    // function('''
+    //   Lots of
+    //   text
+    // ''');
+    // ```
+
+    // TODO(tall): Consider whether immediately-invoked function expressions
+    // should be block argument candidates, like:
+    //
+    // ```
+    // function(() {
+    //   body;
+    // }());
+    // ```
+    return switch (expression) {
+      // A function expression can use either a non-empty parameter list or a
+      // non-empty block body for block formatting.
+      FunctionExpression(:var parameters!, :var body) =>
+        !parameters.parameters.isEmptyBody(parameters.rightParenthesis) ||
+            body is BlockFunctionBody &&
+                !body.block.statements.isEmptyBody(body.block.rightBracket),
+      ListLiteral(:var elements, :var rightBracket) ||
+      SetOrMapLiteral(:var elements, :var rightBracket) =>
+        !elements.isEmptyBody(rightBracket),
+      RecordLiteral(:var fields, :var rightParenthesis) =>
+        !fields.isEmptyBody(rightParenthesis),
+      SwitchExpression(:var cases, :var rightBracket) =>
+        !cases.isEmptyBody(rightBracket),
+      InstanceCreationExpression(:var argumentList) ||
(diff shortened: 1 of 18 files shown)


munificent:
This is a pretty big one, but it's a big step towards actually producing the output we want. Also, it's another example of the kind of formatting that the new formatter can handle but the old one can't. In the old formatter, if an argument could possibly be block formatted, it gets locked into that indentation. If it turns out that the block formatting doesn't fit, you end up with weird indentation like:

```dart
function(
    'some string', <SomeLongType>[
  element,
  element,
  element
]);
```

With this change in the new formatter, you get:

```dart
function(
  'some string',
  <SomeLongType>[
    element,
    element,
    element,
  ],
);
```

I've wanted to fix this for a loooong time.

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -44,8 +49,15 @@ class DelimitedListBuilder {
   /// literal, etc.
   DelimitedListBuilder(this._visitor, [this._style = const ListStyle()]);
 
-  ListPiece build() =>
-      ListPiece(_leftBracket, _elements, _blanksAfter, _rightBracket, _style);
+  /// Create the final [ListPiece] out of the added brackets, delimiters,
```
kallentu, line 52: Nit: `Creates`

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -44,8 +49,15 @@ class DelimitedListBuilder {
...
+  /// elements, and style.
+  ListPiece build() {
+    var blockElement = -1;
+    if (_style.allowBlockElement) blockElement = _findBlockElement();
+
+    return ListPiece(_leftBracket, _elements, _blanksAfter, _rightBracket,
```
kallentu, line 58: Are `_blockCandidates` different from `_elements`, if the `_style.allowBlockElement` is `true`?

If they are, could `blockElement` (which maybe should be named `blockElementIndex`?) potentially point to the wrong element when used in `ListPiece`? 


And if they aren't different, why do we need `_blockCandidates`?

lib/src/ast_extensions.dart
```diff
@@ -126,17 +126,61 @@ extension ExpressionExtensions on Expression {
...
+    //
+    // ```
+    // function(() {
+    //   body;
+    // }());
+    // ```
+    return switch (expression) {
+      // A function expression can use either a non-empty parameter list or a
+      // non-empty block body for block formatting.
+      FunctionExpression(:var parameters!, :var body) =>
```
natebosch, line 166: Is the comment intended to explain the `!`? I don't think I understand it.

lib/src/ast_extensions.dart
```diff
@@ -126,17 +126,61 @@ extension ExpressionExtensions on Expression {
...
+        !parameters.parameters.isEmptyBody(parameters.rightParenthesis) ||
+            body is BlockFunctionBody &&
+                !body.block.statements.isEmptyBody(body.block.rightBracket),
+      ListLiteral(:var elements, :var rightBracket) ||
+      SetOrMapLiteral(:var elements, :var rightBracket) =>
+        !elements.isEmptyBody(rightBracket),
```
natebosch, line 172: [optional] there are so many `!` it makes me curious what it would look like having only 1 outside the switch...

lib/src/ast_extensions.dart
munificent: I changed it to `?`. In practice it doesn't matter. The `parameters` field in `FunctionExpression` is nullable because `FunctionExpression` is used for both anonymous functions and as an inner AST node for function declarations. In the latter case, you might be defining a getter where there is no parameter list, so `parameters` is nullable.

Dart doesn't have anonymous getters, so any actual function expression will have a `parameters`. And we only ever call `isDelimited` on expressions, not declarations. So in practice, `parameters` will always be present.

But... it's not hard to define `isDelimited` to correctly handle an absent parameter list. In that case, it can't split at the parameter list. So I just did that.

lib/src/ast_extensions.dart
munificent: Good idea. I went ahead and turned that function into its negation and gave it a hopefully better name. That led to me also renaming `isDelimited` to something I hope is better.

kallentu (approved):
Thank you!

PR #1324, opened 2023-11-15
Better debug output for piece trees.

munificent:
The old formatter has pretty nice debug output, but the new one is still pretty rudimentary. This improves that situation somewhat from:

```
Sequence(Type(`class Foo` Block(`{` Sequence(Var(`static late final int` `i;`) Var(`static late int` `i;`) Var(`static late var` `i;`) Var(`covariant late var` `i;`) Var(`covariant late int` `i;`) Var(`late final int` `i;`) Var(`late int` `i;`) Var(`late var` `i;`)) `}`)))
```

To:

```
Seq(
  Type(
    `class Foo`
    Block(
      `{`
      Seq(
        Var(`static late final int` `i;`)
        Var(`static late int` `i;`)
        Var(`static late var` `i;`)
        Var(`covariant late var` `i;`)
        Var(`covariant late int` `i;`)
        Var(`late final int` `i;`)
        Var(`late int` `i;`)
        Var(`late var` `i;`)
      )
      `}`
    )
  )
)
```

It also eliminates a bunch of pointless `toString()` implementations that are only used for debugging and can be inferred from runtimeType.

### DIFF
--- a/lib/src/debug.dart
+++ b/lib/src/debug.dart
@@ -268,25 +268,72 @@ void dumpLines(List<Chunk> chunks, SplitSet splits) {
 /// Build a string representation of the [piece] tree.
 String pieceTree(Piece piece) {
   var buffer = StringBuffer();
+  _PieceDebugTree(piece).write(buffer, 0);
+  return buffer.toString();
+}
 
-  void traverse(Piece piece) {
-    buffer.write(piece);
+/// A stringified representation of a tree of pieces for debug output.
+class _PieceDebugTree {
+  final String label;
+  final List<_PieceDebugTree> children = [];
 
-    if (piece is! TextPiece) {
-      var children = <Piece>[];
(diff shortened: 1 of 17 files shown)


lib/src/debug.dart
```diff
@@ -268,25 +268,72 @@ void dumpLines(List<Chunk> chunks, SplitSet splits) {
...
+    });
+  }
 
-      buffer.write('(');
-      for (var child in children) {
-        if (child != children.first) buffer.write(' ');
-        traverse(child);
-      }
-      buffer.write(')');
+  /// The approximate number of characters needed to print this tree.
```
natebosch, line 286: Can you expand this? I was confused by the intention, but I think it would be more clear if it said "needed to print this tree on a single line"

natebosch (approved):

kallentu (approved):

PR #1326, opened 2023-11-17
Improve how expressions in control flow headers are handled.

munificent:
In the process of writing tests for if statements, I realized that the way if conditions are formatted was different than switch values. And it turns out that, based on looking at how all of the control flow in Flutter is hand-formatted, the way I was handling switches was more complex than needed. (This also means that this PR isn't dependent on #1322 like I thought it would be.)

It seems like the right behavior we want is to just visit the expression directly and let the parentheses attach directly to it.

So I updated if, while, and switch to all do that.

Also added tests for comments in if statements.

### DIFF
--- a/test/statement/if_comment.stmt
+++ b/test/statement/if_comment.stmt
@@ -1,0 +1,80 @@
+40 columns                              |
+>>> Line comment after `if`.
+if // comment
+(c) { body; }
+<<<
+if // comment
+(c) {
+  body;
+}
+>>> Line comment before condition.
+if (// comment
+c) { body; }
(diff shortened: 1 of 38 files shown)


natebosch (approved):

test/statement/if_comment.stmt
```diff
@@ -0,0 +1,80 @@
+40 columns                              |
+>>> Line comment after `if`.
+if // comment
```
kallentu, line 3: Woo, it works!

kallentu (approved):

PR #1330, opened 2023-11-23
Parse with language version 3.3

parlough:
Since extension types are now enabled by default (https://github.com/dart-lang/sdk/commit/a19ae4ee688e5297ef95970eb1077cba3fddb5eb) and `dart_style` supports formatting extension types, the language version passed to the analyzer should be updated.

Reference: https://github.com/dart-lang/sdk/issues/54139#issuecomment-1824826027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.3.5-wip
+
+* Use language version `3.3` to parse so that
+  code with extension types can be formatted.
+
 ## 2.3.4
 
 * Add `tall-style` experiment flag to enable the in-progress unstable new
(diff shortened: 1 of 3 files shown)


munificent:
Ah, yes, good catch! Would you mind also bumping the version number in the pubspec.yaml to 2.3.5-wip? That way we know this fix hasn't been released yet.

parlough:
Good point. Updated :)

munificent (approved):
Thank you for the fix! I'll merge this in and roll it into the SDK today.

babbage:
It will be great when this update is released to production.

munificent:
It's published to pub now and the latest bleeding edge Dart SDK has it.

PR #1335, opened 2023-11-28
Update to and fix latest lints

kevmoo:


analysis_options.yaml
```diff
@@ -1,6 +1,5 @@
...
-linter:
-  rules:
-    - avoid_dynamic_calls
-    - directives_ordering
-    - unawaited_futures
+include: package:dart_flutter_team_lints/analysis_options.yaml
+
+analyzer:
+  errors:
+    comment_references: ignore
```
kevmoo, line 5: LOTS of failures here!

lib/src/line_splitting/solve_state_queue.dart
```diff
@@ -14,7 +14,9 @@ import 'solve_state.dart';
 /// overflow characters. When a new state is added to the heap, it will be
 /// discarded, or a previously enqueued one will be discarded, if two overlap.
 class SolveStateQueue {
-  /// Initial capacity of a queue when created, or when added to after a [clear].
+  /// Initial capacity of a queue when created, or when added to after a
+  /// `clear`.
```
kevmoo, line 18: Zero idea where "clear" went.

analysis_options.yaml
```diff
@@ -1,6 +1,5 @@
-include: package:lints/recommended.yaml
-linter:
-  rules:
-    - avoid_dynamic_calls
```
kevmoo, line 4: all of these are in `dart_flutter_team_lints`

lib/src/front_end/comment_writer.dart
```diff
@@ -106,6 +106,7 @@ mixin CommentWriter {
     // just override the script tag's line.
     if (token.previous!.type == TokenType.SCRIPT_TAG) previousLine = tokenLine;
 
+    // ignore: prefer_const_constructors
```
munificent, line 109: Does the linter really yell at you to use `const` here?! It's creating a mutable object. That's a bananas lint. I definitely don't want to enable it for dart_style.

lib/src/front_end/comment_writer.dart
```diff
@@ -312,7 +313,7 @@ class CommentSequence extends ListBase<SourceComment> {
   SourceComment operator [](int index) => _comments[index];
 
   @override
-  operator []=(int index, SourceComment value) =>
+  void operator []=(int index, SourceComment value) =>
```
munificent, line 316: This return type is pointless. Index setters are like setters: they can never return a value anyway.

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -3,8 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
-import 'package:dart_style/src/ast_extensions.dart';
```
munificent, line 6: Nice catch. I wish VS Code didn't automatically use package imports when it auto-adds an import from the same package.

lib/src/line_splitting/solve_state_queue.dart
munificent: This class was copied from HeapPriorityQueue and then heavily modified, so this comment is just referring to something that no longer applies to this class. You can just remove everything after "created".

munificent (changes requested):

lib/src/front_end/comment_writer.dart
kevmoo: Generally agree, so maybe the lint needs to be updated for this case!

lib/src/front_end/delimited_list_builder.dart
kevmoo: agreed!

lib/src/front_end/comment_writer.dart
kevmoo: Yep!

If we don't want it enabled, there's a lot of churn to delete.

example/format.dart
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: unreachable_from_main
```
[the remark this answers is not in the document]
kevmoo: okay to just delete top-level functions that aren't used?

lib/src/line_splitting/solve_state_queue.dart
kevmoo: done

lib/src/argument_list_visitor.dart
```diff
@@ -96,10 +96,11 @@ class ArgumentListVisitor {
...
       this._arguments,
       this._functions,
-      this._argumentsAfterFunctions) {
-    assert(_functions == null || _argumentsAfterFunctions != null,
-        'If _functions is passed, _argumentsAfterFunctions must be too.');
-  }
+      this._argumentsAfterFunctions)
+      : assert(
+          _functions == null || _argumentsAfterFunctions != null,
+          'If _functions is passed, _argumentsAfterFunctions must be too.',
```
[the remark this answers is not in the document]
kevmoo: done

lib/src/dart_formatter.dart
```diff
@@ -82,8 +82,9 @@ class DartFormatter {
       throw ArgumentError('uri must be `null`, a Uri, or a String.');
     }
 
-    return formatSource(SourceCode(source, uri: uri, isCompilationUnit: true))
-        .text;
+    return formatSource(
+      SourceCode(source, uri: uri as String?, isCompilationUnit: true),
```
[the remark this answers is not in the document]
kevmoo: done

lib/src/front_end/comment_writer.dart
munificent: I believe the lint is doing the right thing. I enabled always_declare_return_types and tested it on:

```dart
class Foo {
  get getter => 1;
  set setter(int value) => 1;
  method() => 1;
  operator [](int index) => 1;
  operator []=(int index, int value) => 1;
}
```

I get lint warnings on `getter`, `method()`, and `operator []`, but not `setter` or `operator []=`. Maybe the fix is out of sync with the lint? Or maybe this was just a mistake fixing the lint? Either way, I think you can delete this `void` and it should still be lint clean.

munificent (approved):
This LGTM, but note that I have a very large PR out (#1336) that will probably collide with this in nasty ways and I really don't want that PR (which is blocking any other progress on the new back end) any harder to land than it already is.

Would you be OK with waiting until that's in and updating this PR with those changes first?

PR #1338, opened 2023-12-07
Format constructor declarations.

munificent:
Includes everything that can go in a constructor:

- Factory and redirecting constructors.
- Initializing formals ("this." parameters).
- Super parameters.
- Constructor initializers.
- Assert initializers.

Constructors are tricky because there are some constraints between how the parameter list and initializer list are allowed to split. Up until now, aside from setAllowNewlines(), there aren't really any ways that pieces interact.

To support this, I added a pretty simple API where when a piece calls CodeWriter.format() on a child piece, it can pass it what state the piece is required to be in. If the solution doesn't have the child in that state, it invalidates the solution.

It seems to work pretty well. There is probably a more optimal way to implement that in the solver so that we test these constraints even before formatting and discard the solution more eagerly. I left a TODO comment about that and we can look into it later when we're benchmarking.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -13,7 +13,9 @@ import '../piece/adjacent.dart';
 import '../piece/assign.dart';
 import '../piece/block.dart';
 import '../piece/chain.dart';
+import '../piece/constructor.dart';
 import '../piece/for.dart';
+import '../piece/function.dart';
 import '../piece/if.dart';
 import '../piece/infix.dart';
 import '../piece/list.dart';
@@ -130,7 +132,17 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
 
   @override
   Piece visitAssertInitializer(AssertInitializer node) {
-    throw UnimplementedError();
+    return buildPiece((b) {
+      b.token(node.assertKeyword);
+      b.add(createList(
+        leftBracket: node.leftParenthesis,
+        [
+          node.condition,
+          if (node.message case var message?) message,
+        ],
+        rightBracket: node.rightParenthesis,
+      ));
+    });
   }
 
   @override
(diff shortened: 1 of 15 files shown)


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1310,7 +1405,30 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
   @override
   Piece visitSuperFormalParameter(SuperFormalParameter node) {
-    throw UnimplementedError();
+    if (node.parameters case var parameters?) {
+      // A function-typed super parameter like:
+      //
+      // ```
+      // C(super.fn(parameter));
+      // ```
+      return createFunctionType(
```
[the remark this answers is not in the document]
munificent: Yeah, this is a fair point. I started using backticks during the rewrite because I thought it was maybe a little easier to write than using indentation. But in retrospect the backticks require an extra pair of lines which is lame.

How about I fix all of them in a separate PR?

lib/src/piece/piece.dart
```diff
@@ -55,6 +55,14 @@ abstract class Piece {
   /// Invokes [callback] on each piece contained in this piece.
   void forEachChild(void Function(Piece piece) callback);
 
+  /// The cost that this piece should apply to the solution when in [state].
+  ///
+  /// This is usually just the state's cost, but some pieces may want to tweak
+  /// the cost in certain circumstances.
+  // TODO(tall): Given that we have this API now, consider whether it makes
+  // sense to remove the cost field from State entirely.
+  int stateCost(State state) => state.cost;
```
natebosch, line 64: This API feels iffy to me. Is there any other way we can handle this? Would it work to do `if (_mustSplit) piece.pin(State(2, cost: _style.splitCost);` instead of `pin(State.split)`?

lib/src/front_end/piece_factory.dart
```diff
@@ -253,23 +286,37 @@ mixin PieceFactory {
...
+      b.visit(body);
+    });
   }
 
   /// Creates a function type or function-typed formal.
   ///
   /// If creating a piece for a function-typed formal, then [parameter] is the
   /// formal parameter.
+  ///
+  /// If this is a function-typed initializing formal (`this.foo()`, then
```
kallentu, line 308: Nit: Missing a closing `)` in the comment.

lib/src/piece/constructor.dart
```diff
@@ -0,0 +1,176 @@
...
+    callback(_header);
+    callback(_parameters);
+    if (_redirect case var redirect?) callback(redirect);
+    if (_initializerSeparator case var separator?) callback(separator);
+    if (_initializers case var initializers?) callback(initializers);
+    callback(_body);
+  }
+
+  @override
+  String get debugName => 'Ctor';
```
kallentu, line 175: Do we still need this override with how `Piece` handles the debug name?

lib/src/piece/constructor.dart
```diff
@@ -0,0 +1,176 @@
...
+/// the parameter list is allowed to split and how the initializer list is
+/// allowed to split. Only a few combinations of splits are allowed:
+///
+/// [State.unsplit] No splits at all, in the parameters or initializers.
+///
+/// ```
+///   SomeClass(param) : a = 1, b = 2;
+/// ```
+///
+/// [_splitBeforeInitializers] Split before the `:` and between the
```
kallentu, line 20: These examples are great to visualize everything. Thanks for including them.

kallentu (approved):
Generally looks good to me. I don't have too much of an opinion with the API that Nate commented on, but assume my approval is with that thread resolved.

lib/src/piece/piece.dart
natebosch: > 1. That would work for this example, but I don't know if we'll run into cases in the future

I generally have a strong preference to lean on YAGNI and go for the least possible API surface until the other use case shows up.

I'm fine going forward with this as is and reconsidering the coupling between states, pieces, and the code writer after more of the language is implemented.

natebosch (approved):

lib/src/piece/piece.dart
munificent: If it's OK with you, I'm going to keep this as-is with a TODO comment still in there, and I'll think more about better ways to express relative priority between formatting choices and constraints between pieces. I have some ideas, but it's hard to evaluate them until we have some more complex pieces implemented that need them.

PR #1343, opened 2023-12-16
Refactor how piece constraints are applied.

munificent:
When I added support for constructors, I introduced a little API to let pieces enforce constraints between their state and the states of other child pieces. It worked but felt a little bolted on. In particular, enforcing the constraints during CodeWriter formatting meant we didn't prune invalid solutions as efficiently as we could.

I'm working on if elements now and I needed a better way to handle constraints between pieces for that, so I went ahead and refactored the way this works to be, I think, generally better:

- Enforce constraints between pieces during solution expansion instead of during formatting. This means that we eagerly discard any solution (and the whole tree of solutions that could have come from it) if it doesn't meet the constraints.

- Remove the `_pieces` fields in PieceStateSet. With the previous optimization to track the next unsolved piece in Solution, that list wasn't doing much of anything. It's only use (aside from debug output) was for determining the order that pieces are compared when comparing two solutions. But we can get that from iterating over the _pieceStates map since maps in Dart track their key insertion order.

- Update ConstructorPiece to use this new API.

### DIFF
--- a/lib/src/back_end/solution.dart
+++ b/lib/src/back_end/solution.dart
@@ -7,23 +7,19 @@ import 'code_writer.dart';
 /// A possibly incomplete set of selected states for a set of pieces being
 /// solved.
 class PieceStateSet {
-  // TODO(perf): Looking up and expanding the set of chunk states was a
-  // performance bottleneck in the old line splitter. If that turns out to be
-  // true here, then consider a faster representation for this list and the
-  // subsequent map field.
-  /// The in-order flattened list of all pieces being solved.
+  /// The states that pieces have been bound to.
   ///
-  /// This doesn't include pieces like text that have only a single value since
-  /// there's nothing to solve for them.
-  final List<Piece> _pieces;
-
+  /// Note that order that keys are inserted into this map is significant. When
+  /// ordering solutions, we use the order that pieces are bound in here to
+  /// break ties between solutions that otherwise have the same cost and
+  /// overflow.
   final Map<Piece, State> _pieceStates;
 
   /// Creates a new [PieceStateSet] with no pieces set to any state (which
   /// implicitly means they have state 0).
-  PieceStateSet(this._pieces) : _pieceStates = {};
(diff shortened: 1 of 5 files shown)


lib/src/back_end/solution.dart
```diff
@@ -137,10 +158,15 @@ class Solution implements Comparable<Solution> {
...
-      ];
+      var result = <Solution>[];
+      for (var state in piece.states) {
+        var stateSet = _state.tryBind(piece, state);
+        if (stateSet != null) {
+          result.add(Solution._(root, pageWidth, stateSet));
+        }
+      }
+
+      return result;
```
natebosch, line 169: [optional]

```suggestion
      return [
        for (var state in piece.states)
          if (_state.tryBind(piece, state) case final stateSet?)
            Solution._(root, pageWidth, stateSet)
      ];
```

lib/src/back_end/solution.dart
```diff
@@ -137,10 +158,15 @@ class Solution implements Comparable<Solution> {
   /// piece and yields further solutions for each state that piece can have.
   List<Solution> expand(Piece root, int pageWidth) {
     if (_nextPieceToExpand case var piece?) {
-      return [
-        for (var state in piece.states)
-          Solution._(root, pageWidth, _state.cloneWith(piece, state))
-      ];
+      var result = <Solution>[];
+      for (var state in piece.states) {
+        var stateSet = _state.tryBind(piece, state);
```
natebosch, line 163: [out of scope for this PR] Can we consider renaming `_state` to `_stateSet` or similar? The distinction between `_state` and `state` here is too subtle IMO.

natebosch (approved):

kallentu (approved):

lib/src/back_end/solution.dart
munificent: Yeah, I agree that's much better. I'll just do it now.

PR #1344, opened 2023-12-16
Format if elements.

munificent:
I was hoping to be able to reuse more of the createIf() code in PieceFactory for both if elements and if statements, but there are enough differences between the two that that didn't work out, so I just inlined the code for both of them directly in AstNodeVisitor.

I was able to reuse the existing IfPiece, though, which is nice.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -93,6 +93,23 @@ extension AstNodeExtensions on AstNode {
 
     return null;
   }
+
+  /// If this is a spread of a non-empty collection literal, then returns `this`
+  /// as a [SpreadElement].
+  ///
+  /// Otherwise, returns `null`.
+  SpreadElement? get spreadCollection {
+    var node = this;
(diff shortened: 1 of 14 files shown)


lib/src/ast_extensions.dart
```diff
@@ -93,6 +93,23 @@ extension AstNodeExtensions on AstNode {
...
+  ///
+  /// Otherwise, returns `null`.
+  SpreadElement? get spreadCollection {
+    var node = this;
+    if (node is! SpreadElement) return null;
+
+    return switch (node.expression) {
+      ListLiteral(:var elements, :var rightBracket) ||
+      SetOrMapLiteral(:var elements, :var rightBracket)
+          when elements.canSplit(rightBracket) =>
```
natebosch, line 108: Consider adding a comment explaining why `element.canSplit` indicates a non-empty collection.

natebosch (approved):

kallentu (approved):

PR #1346, opened 2023-12-27
Format record type annotations.

kallentu:
As title says, formats record type annotations.
- Used a `DelimitedListBuilder` for these since they work similarly to parameter lists, but not quite the same.
- Added a few comment tests, just to make sure they work.

The comments show how the splitting has changed for these, specifically in the case of:
```
>>> Split positional types.
(TypeName,TypeName,TypeName,TypeName) x;
<<<
(
  TypeName,
  TypeName,
  TypeName,
  TypeName,
) x;
```
which no longer splits like:
```
(
  TypeName,
  TypeName,
  TypeName,
  TypeName,
)
x;
```


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1336,19 +1336,70 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
+      builder.leftBracket(node.leftParenthesis);
+    }
+
+    for (var positionalField in positionalFields) {
+      builder.visit(positionalField);
+    }
+
+    Token? rightDelimiter;
+    if (namedFields != null) {
+      // If this is the first optional parameter, put the delimiter before it.
```
munificent, line 1364: "optional parameter" -> "named field".

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1336,19 +1336,70 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
+        builder.visit(namedField);
+      }
+      rightDelimiter = namedFields.rightBracket;
+    }
+
+    builder.rightBracket(node.rightParenthesis, delimiter: rightDelimiter);
+    return buildPiece((b) {
+      b.add(builder.build());
+      if (node.question case var question?) {
+        b.add(tokenPiece(question));
```
munificent, line 1378:
```dart
b.token(question);
```

AdjacentBuilder has helper functions for `token()` and `visit()` since adding a token or visited AST node is so common.

In fact, you can even just do:

```dart
b.token(node.question);
```

And get rid of the surrounding `if` because `AdjacentBuilder.token()` silently does nothing if you pass it `null`, since optional tokens are so common.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1336,19 +1336,70 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
 
   @override
   Piece visitRecordTypeAnnotationNamedField(
       RecordTypeAnnotationNamedField node) {
-    throw UnimplementedError();
+    // TODO(tall): Format metadata.
+    if (node.metadata.isNotEmpty) throw UnimplementedError();
+    return buildPiece((b) {
+      b.visit(node.type);
+      b.token(node.name, spaceBefore: true);
```
munificent, line 1390: I think we'll want to allow splitting between the type and name if both are present, since we do for formal parameters and record type fields sort of mirror those.

Take a look at `PieceFactory.createFormalParameter()` and either copy from there or, if it makes sense, refactor out a little helper function?

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1336,19 +1336,70 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
   @override
   Piece visitRecordTypeAnnotationPositionalField(
       RecordTypeAnnotationPositionalField node) {
-    throw UnimplementedError();
+    // TODO(tall): Format metadata.
+    if (node.metadata.isNotEmpty) throw UnimplementedError();
+    return buildPiece((b) {
+      b.visit(node.type);
+      b.token(node.name, spaceBefore: true);
+    });
```
munificent, line 1402: Probably worth making `PieceFactory.createRecordTypeField()` and have both `visitRecordTypeAnnotationNamedField()` and `visitRecordTypeAnnotationPositionalField()` call it.

test/declaration/class.unit
```diff
@@ -73,4 +73,12 @@ abstract base mixin class C13 {}
...
 }
 <<<
-class SomeClass native "Zapp" {}
\ No newline at end of file
+class SomeClass native "Zapp" {}
+>>> Record type with multiple fields in parameter has no trailing comma.
+class C { C((TypeName,TypeName,) this.record) {;} }
+<<<
+class C {
+  C((TypeName, TypeName) this.record) {
```
munificent, line 81: Is this mostly testing that we can handle record types on initializing formals? If it's just about using a record type in a parameter, you could make a simpler example like:

```dart
function((TypeName, TypeName) parameter) { ; }
```

(And put it in probably "type/function.stmt".)

test/type/record.stmt
```diff
@@ -0,0 +1,184 @@
...
+someLongFunctionName__________________(
+  () x,
+) {}
+>>> Unsplit short single positional field.
+(TypeName,
+) x;
+<<<
+(TypeName,) x;
+>>> Unsplit single positional field.
+function((VeryLongTypeName____________,) x) {;}
```
munificent, line 64: Unless you specifically think it's helpful to test these in parameter position, you can simplify them by just using a variable declaration instead of a function:

```dart
(VeryLongTypeName____________,) x;
```

munificent (approved):
This is fantastic. The tests are excellent. I made a few suggests on the code, but it's all minor stuff. This is a nice Christmas present to come back to. :)

test/declaration/class.unit
kallentu: Moved that and a bunch of other tests to `type/function.stmt`.

lib/src/front_end/ast_node_visitor.dart
kallentu: Amazing. Will do and will remember for next time.

lib/src/front_end/ast_node_visitor.dart
kallentu: Good call. I added a commit + a test commit that changed this behaviour if you want to take a look.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1336,19 +1336,58 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
+      builder.leftBracket(node.leftParenthesis);
+    }
+
+    for (var positionalField in positionalFields) {
+      builder.visit(positionalField);
+    }
+
+    Token? rightDelimiter;
+    if (namedFields != null) {
+      // If this is the first named field, put the delimiter before it.
```
natebosch, line 1364: [nit] The comment and conditional are both "if" but the condition does not match the comment - it feel a little misleading.

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
kallentu: Updated the comment to try to explain it a little more clearly.

kallentu:
Sending to @sigmundch for readability comments. Please review as a readability CL and add your feedback at [go/dart-team-readability-feedback](http://goto.google.com/dart-team-readability-feedback)

lib/src/front_end/piece_factory.dart
```diff
@@ -542,6 +532,17 @@ mixin PieceFactory {
     return builder.build();
   }
 
+  /// Creates an [AdjacentPiece] for a given record type field.
+  Piece createRecordTypeField(RecordTypeAnnotationField node) {
+    // TODO(tall): Format metadata.
```
sigmundch, line 537: sounds like something that is already done generally in this repo, but what is `TODO(tall)`? I was expecting a `TODO(your-ldap-here)` instead.

sigmundch (approved):
LGTM!

lib/src/front_end/piece_factory.dart
munificent: We are essentially writing an entirely new formatter to support the new style (while still supporting the old one in the interim). So in the new formatter code, like PieceFactory here, no existing language feature is supported until we implement it again. Metadata isn't implemented yet.

We're using `TODO(tall)` as marker comments to keep track of work left to do in the new implementation. The idea is that the new implementation isn't done until all of those (and `throw UnimplementedError` and `TODO(perf)`) are addressed and removed.

sigmundch: totally makes sense, thanks for clarifying!

Long ago we had a `TODO(redemption)` in dart2js for the same purpose... to my surprise, we still have 6 of those left :open_mouth:

PR #1347, opened 2023-12-28
Format throw and rethrow.

kallentu:
Pretty basic token and expression formatting and some new tests I've added. :D

### DIFF
--- a/test/statement/other.stmt
+++ b/test/statement/other.stmt
@@ -55,3 +55,25 @@ foo() async {
 foo() async {
   await i(1 + 2);
 }
+>>> Throw.
+throw   'error'
+;
+<<<
+throw 'error';
+>>> Throw with long string literal.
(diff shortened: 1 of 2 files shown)


test/statement/other.stmt
```diff
@@ -55,3 +55,25 @@ foo() async {
...
+throw   'error'
+;
+<<<
+throw 'error';
+>>> Throw with long string literal.
+throw new FormatException('This is a long exception message.');
+<<<
+throw new FormatException(
+  'This is a long exception message.',
+);
```
munificent, line 68: Probably worth testing that it never splits after `throw`:

```
throw 'Some extremely long error message.';
```

munificent (approved):
Looks great!

test/statement/other.stmt
kallentu: Yes! Added.

PR #1349, opened 2024-01-02
Format function references.

kallentu:
Small PR to format function references -- really just builds upon type argument formatting.
I also migrated previous function reference tests over.


munificent (approved):
\o/

natebosch (approved):

PR #1350, opened 2024-01-04
Format for elements.

munificent:
This mostly uses the existing code for formatting the `for (...)` part of for statements, which I moved into PieceFactory so that it can be reused.

In the process of adding tests for this, I noticed that a newline in the initializer of a for-in loop forced the `in` to split, which looks weird, so I fixed that too, for both for statements and for elements.

I also added trivial support for postfix expressions because some of the for element tests use `i++`. I didn't migrate the tests for those yet since they aren't really related to this change, so added a TODO to do that in a later PR.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -695,8 +695,8 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
     var forPiece = ForPiece(forKeyword, forPartsPiece, body,
         hasBlockBody: node.body.isSpreadCollection);
 
-    // If the body of a for element is itself a control flow element, then
-    // force the outer for to always split.
+    // It looks weird to have multiple nested control flow elements on the same
+    // line, so force the outer one to split if there is an inner one.
     if (node.body.isControlFlowElement) {
       forPiece.pin(State.split);
     }
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -371,7 +371,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
     if (node.redirectedConstructor case var constructor?) {
       redirect = AssignPiece(
           tokenPiece(node.separator!), nodePiece(constructor),
-          isValueDelimited: false);
+          allowInnerSplit: false);
     } else if (node.initializers.isNotEmpty) {
       initializerSeparator = tokenPiece(node.separator!);
       initializers = createList(node.initializers,
@@ -583,7 +583,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
       var expression = nodePiece(node.expression);
 
       b.add(AssignPiece(operatorPiece, expression,
-          isValueDelimited: node.expression.canBlockSplit));
+          allowInnerSplit: node.expression.canBlockSplit));
       b.token(node.semicolon);
     });
   }
@@ -683,7 +683,25 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
 
   @override
   Piece visitForElement(ForElement node) {
-    throw UnimplementedError();
+    var forKeyword = buildPiece((b) {
+      b.modifier(node.awaitKeyword);
+      b.token(node.forKeyword);
+    });
+
+    var forPartsPiece = createForLoopParts(
+        node.leftParenthesis, node.forLoopParts, node.rightParenthesis);
+    var body = nodePiece(node.body);
+
+    var forPiece = ForPiece(forKeyword, forPartsPiece, body,
+        hasBlockBody: node.body.isSpreadCollection);
+
+    // If the body of a for element is itself a control flow element, then
+    // force the outer for to always split.
+    if (node.body.isControlFlowElement) {
+      forPiece.pin(State.split);
+    }
+
+    return forPiece;
   }
 
   @override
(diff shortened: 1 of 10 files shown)


test/statement/for_in.stmt
```diff
@@ -58,6 +58,13 @@ for (var identifier
     in iteratableExpression) {
   body;
 }
+>>> Split inside initializer.
+for (var identifier in someVeryLong + iterableExpression) { body; }
+<<<
+for (var identifier in someVeryLong +
+    iterableExpression) {
```
kallentu, line 65: Is this what you meant by `newline in the initializer of a for-in loop forced the in to split`?
Honestly, I thought the split was going to be 
```
for (var identifier in
     someVeryLong + iterableExpression) {
  body;
}
```

kallentu (approved):

test/statement/for_in.stmt
munificent: There are two separate style choices:

1. If a newline occurs inside the expression after `in`, does that force us to split before the `in` as well?

In many places, a newline inside a subexpression will also force the outer expression to split. For example, the formatter will never produce:

```dart
expression1 + (expression2 *
    expression3);
```

If either operand to an infix operator contains a newline, we force the infix operator to split too (even if doing so isn't necessary to fit in the page width), so you get:

```dart
expression1 +
    (expression2 * expression3);
```

Or:

```dart
expression1 +
    (expression2 *
        expression3);
```

Depending on whether splitting at the outer infix operator is enough to not need to also split the inner infix operator.

For "assignment" pieces (`=`, `=>`, `in`, `:`), there is similar logic. For example, if you have a really long map key or value (we use AssignPiece for map entries too), you can get:

```dart
var map = {
  someReallyLong + keyExpression:
    theValue
}
```

But not:

```dart
var map = {
  someReallyLong +
      keyExpression: theValue
}
```

Before this PR, `in` had the same behavior, which meant you could get:

```dart
for (var identifier
      in someVeryLong +
          iterableExpression) {
  body;
}
```

Or:

```dart
for (var identifier
      in someVeryLong + iterableExpression) {
  body;
}
```

But never:

```dart
for (var identifier in someVeryLong +
    iterableExpression) {
  body;
}
```

Whether we want to allow the latter is basically a style choice. The old formatting style tries to be pretty dense and avoid newlines when possible, but my impression is that users generally like a sort of taller, more open style. So when I started working on the new style, one of the changes I made is that for assignment-like things, it generally does have the infix-operator-like rule that a split inside an operand forces the assignment to split.

But I think that choice looks pretty funny for `in`. Honestly, I'm not sure it's the right fit for `=` either. I think many users would be OK with:

```dart
variable = someLongExpression +
    thatWraps;
```

So it may be that we end up changing other assignment-like pieces to allow splitting in the RHS without forcing the assignment operator to split. But, for now, I just did it for `in`. 

There's a long TODO(tall) comment in AssignPiece about this so that our future selves can play with that style choice once we can run the formatter on a corpus.

2. If the whole for clause doesn't fit on one line because the initializer is too long, do we prefer to split before the `in`, or inside the initializer itself?

Assuming that we do allow a split inside the RHS after `in` (style choice 1) and also allow splitting before the `in`, there is a separate style choice for which of those should be *preferred* when either is sufficient to fit in the page width. I think that's the question you're asking.

In general, the formatter usually prefer to split at outer constructs before inner ones. In fact, one of the main ways it models that preference is the above style choice where in many contexts, splitting in a subexpression *forces* the outer one to split. That leads the formatter to only be able to choose between "split at outer and not inner" and "split at *both* outer and inner". Since fewer splits are always better, it will always then prefer to split at the outer expression and not the inner if that fits.

But in places where we don't have that hard constraint (like with `in` here), we now have a choice as to which should be preferred.

I don't have a strong sense of what the right answer is. We'll probably want to iterate on it. For what it's worth, the PR here follows what the old style does and I can't recall people complaining about how newlines for-in loop initializers get handled. I added a TODO comment so that we reconsider it when things are further along.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -683,121 +683,36 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
+    var forPiece = ForPiece(forKeyword, forPartsPiece, body,
+        hasBlockBody: node.body.isSpreadCollection);
 
-        // The update clauses.
-        if (forParts.updaters.isNotEmpty) {
-          partsList.addCommentsBefore(forParts.updaters.first.beginToken);
-          partsList.add(createList(forParts.updaters,
-              style: const ListStyle(commas: Commas.nonTrailing)));
-        }
+    // If the body of a for element is itself a control flow element, then
```
natebosch, line 698: [nit] This repeats the code - are the any "why" details worth adding?

lib/src/front_end/piece_factory.dart
```diff
@@ -219,6 +219,142 @@ mixin PieceFactory {
...
+                condition: null,
+                rightSeparator: Token(precedingComments: null),
+                updaters: NodeList(isEmpty: true),
+              ) &&
+              var forParts
+          when rightParenthesis.precedingComments == null:
+        return buildPiece((b) {
+          b.token(leftParenthesis);
+          b.token(forParts.leftSeparator);
+          b.token(forParts.rightSeparator);
```
natebosch, line 241:
```suggestion
      case ForPartsWithExpression(
            initialization: null,
            leftSeparator: Token(precedingComments: null),
            condition: null,
            rightSeparator: Token(precedingComments: null),
            updaters: NodeList(isEmpty: true),
          )
          when rightParenthesis.precedingComments == null:
        return buildPiece((b) {
          b.token(leftParenthesis);
          b.token(forLoopParts.leftSeparator);
          b.token(forLoopParts.rightSeparator);
```

I don't think we should need `&& var forParts` here. The `forLoopParts` variable should be promoted right?

lib/src/front_end/piece_factory.dart
```diff
@@ -219,6 +219,142 @@ mixin PieceFactory {
...
+          b.token(rightParenthesis);
+        });
+
+      case ForParts forParts &&
+            ForPartsWithDeclarations(variables: AstNode? initializer):
+      case ForParts forParts &&
+            ForPartsWithExpression(initialization: AstNode? initializer):
```
natebosch, line 248: [optional] I find this a little easier to read with the shared part between the cases written once. WDYT?

```suggestion
      case ForParts forParts &&
            (ForPartsWithDeclarations(variables: AstNode? initializer) ||
                ForPartsWithExpression(initialization: AstNode? initializer)):
```

lib/src/front_end/piece_factory.dart
```diff
@@ -219,6 +219,142 @@ mixin PieceFactory {
...
+        partsList.rightBracket(rightParenthesis);
+        return partsList.build();
+
+      case ForPartsWithPattern():
+        throw UnimplementedError();
+
+      case ForEachParts forEachParts &&
+            ForEachPartsWithDeclaration(loopVariable: AstNode variable):
+      case ForEachParts forEachParts &&
+            ForEachPartsWithIdentifier(identifier: AstNode variable):
```
natebosch, line 307: [optional] 
```suggestion
      case ForEachParts forEachParts &&
            (ForEachPartsWithDeclaration(loopVariable: AstNode variable) ||
                ForEachPartsWithIdentifier(identifier: AstNode variable)):
```

lib/src/front_end/piece_factory.dart
```diff
@@ -696,11 +832,23 @@ mixin PieceFactory {
...
-      bool spaceBeforeOperator = true}) {
+      bool spaceBeforeOperator = true,
+      bool allowInnerSplit = false}) {
+    // If the right-hand side can have block formatting, then a newline in
+    // it doesn't force the operator to split, as in:
+    //
+    //    var list = [
+    //      element,
+    //    ];
+    allowInnerSplit |= rightHandSide.canBlockSplit;
```
natebosch, line 850: I don't have a problem with this but I don't recall how common it is - do we consider it idiomatic to use bitwise-or-assignment for booleans?

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: I do slightly prefer the separate cases here.

lib/src/front_end/ast_node_visitor.dart
munificent: Good point done.

lib/src/front_end/piece_factory.dart
munificent: Good catch! Fixed locally.

lib/src/front_end/piece_factory.dart
munificent: It's not as common as I expected, but it happens. Using `||=` would be the idiomatic self-assignment operator for this, but Dart doesn't have `||=` and `|=` does the right thing since `bool` implements `|`.

lib/src/front_end/piece_factory.dart
munificent: I think I like the slightly flatter original code.

PR #1351, opened 2024-01-04
Change code blocks to use indentation instead of backticks.

munificent:
Not exactly the most impactful change. :) But in a previous PR, Nate suggested that old-style Markdown indentation for code blocks would be better than backticks since it's shorter and we aren't actually running the comments through a Markdown formatter anyway.

This converts all of them over in on Go so the codebase is consistent.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -131,27 +131,21 @@ extension ExpressionExtensions on Expression {
   /// example, in an assignment, a split in the assigned value is usually
   /// indented:
   ///
-  /// ```
-  /// var variableName =
-  ///     longValue;
-  /// ```
+  ///     var variableName =
+  ///         longValue;
   ///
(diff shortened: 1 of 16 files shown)


kallentu (approved):

lib/src/ast_extensions.dart
```diff
@@ -131,27 +131,21 @@ extension ExpressionExtensions on Expression {
   /// example, in an assignment, a split in the assigned value is usually
   /// indented:
   ///
-  /// ```
```
kevmoo, line 134: I prefer backticks + `dart` – just to minimize the dart-specific magic. 🤷

munificent: Using indentation is actually [the original Markdown syntax for code blocks](https://daringfireball.net/projects/markdown/syntax#precode). Backticks were a non-standard extension added by GitHub.

natebosch (approved):

PR #1353, opened 2024-01-04
Format if case for if statements.

kallentu:
This only formats if cases for if statements. I'll work on if element cases later.

- Added one basic test for if cases. There's no existing test that doesn't need a pattern or two. I assume the coverage will be added as we implement more patterns.
- Dissected `startControlFlow` and removed that call in order to handle the case clause. If there's a better way to do this, let me know.
- Added a new `pattern` test folder to hold all the patterns tests.


natebosch (approved):

test/pattern/if_case.stmt
```diff
@@ -0,0 +1,7 @@
+40 columns                              |
```
munificent, line 1: I'd move this test to "statement/if_case.stmt" since it's about testing the if-case itself (I think) and not the patterns within it.

Also, can you add another test file ("expression/collection_if_case.stmt") for if-case elements?

```dart
var list = [
  if (obj case true) element
];
```

(And then eventually that will expand to the various ways that you can split inside if-case and patterns inside if-case elements.)

munificent (approved):
Two things in one comment then LGTM!

test/pattern/if_case.stmt
kallentu: For the part on moving it to `statement/if_case.stmt`, that sounds good. Done.

For the second part on if-case elements. This PR doesn't implement if-case elements yet. I'll do that once I implement it next PR.

PR #1355, opened 2024-01-09
Update TODO comment on for-in to consider how "in" should split.

munificent:
In the old and new style, if a newline occurs in the pattern of an if-case statement or element, it forces a split before the `case` too. That syntax is structurally similar to for-in loops, but for-in loops currently don't force that split. (And I had to add explicit support to *not* force a split there.)

Update the TODO comment to consider whether we want to make for-in loops format similar to if-case.


kallentu (approved):
Perfect. Thanks for tracking these discussions.

PR #1356, opened 2024-01-11
Format call chains!

munificent:
This handles unsplit call chains, fully split ones, as well as block-style calls, like:

```
target.call(
  argument,
);
```

This doesn't fully implement the proposal I sketched out because it only allows a single block call. There's a long TODO(tall) comment about whether we want to do that, but I figured it makes sense to do that as a separate iteration since there is so much in here.

But it implements most of it, and I've migrated all of the existing tests over. It ended up being both simpler than the old formatter while also producing better output.

The old formatter has a restriction where the indentation of a piece of code must be fixed regardless of which way it splits. It can't handle:

```
// Block style, so indent argument +2:
target.property.call(
  argument
);

// Can't do block style, so indent +6:
target
    .slightlyLongerProperty
    .call(
      argument
    );
```

That leads to it sometimes producing gross output like:

```
target
    .longProperty
    .anotherLongProperty
    .method(
  wtfIsGoingOnHere,
);
```

To try to avoid that, it has lots of heuristics during Chunk construction to try to not end up with output like that during line splitting.

The new formatter doesn't need any of those heuristics because it can freely choose indentation during line splitting. 🎉

### DIFF
--- a/lib/src/front_end/chain_builder.dart
+++ b/lib/src/front_end/chain_builder.dart
@@ -184,7 +184,8 @@ class ChainBuilder {
     }
   }
 
-  /// Creates a Piece for [target].
+  /// Creates and stores the resulting Piece for [target] as well as whether it
+  /// allows being split.
   void _visitTarget(Expression target) {
     _allowSplitInTarget = target.canBlockSplit;
     _target = _visitor.nodePiece(target);
--- a/lib/src/front_end/chain_builder.dart
+++ b/lib/src/front_end/chain_builder.dart
@@ -1,0 +1,206 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+
+import '../ast_extensions.dart';
+import '../piece/chain.dart';
+import '../piece/piece.dart';
+import 'piece_factory.dart';
+
+/// Creates [Chain] pieces from method calls and property accesses, along with
+/// postfix operations (`!`, index operators, and function invocation
+/// expressions) that follow them.
+///
+/// In the AST for method calls, selectors are nested bottom up such that this
+/// expression:
+///
+///     obj.a(1)[2].c(3)
+///
+/// Is structured like:
+///
+///           .c()
+///           /  \
+///          []   3
+///         /  \
+///       .a()  2
+///       /  \
+///     obj   1
+///
+/// This means visiting the AST from top down visits the selectors from right
+/// to left. It's easier to format if we organize them as a linear series of
+/// selectors from left to right. Further, we want to organize it into a
+/// two-tier hierarchy. We have an outer list of method calls and property
+/// accesses. Then each of those may have one or more postfix selectors
+/// attached: indexers, null-assertions, or invocations. This mirrors how they
+/// are formatted.
+///
+/// This lets us create a single [ChainPiece] for the entire series of dotted
+/// operations, so that we can control splitting them or not as a unit.
+class ChainBuilder {
+  final PieceFactory _visitor;
+
+  /// The left-most target of the chain.
+  late Piece _target;
+
+  /// Whether the target expression may contain newlines when the chain is not
+  /// fully split. (It may always contain newlines when the chain splits.)
+  ///
+  /// This is true for most expressions but false for delimited ones to avoid
+  /// ugly formatting like:
+  ///
+  ///     function(
+  ///       argument,
+  ///     )
+  ///         .method();
+  late final bool _allowSplitInTarget;
+
+  /// The dotted property accesses and method calls following the target.
+  final List<ChainCall> _calls = [];
+
+  ChainBuilder(this._visitor, Expression expression) {
+    _unwrapCall(expression);
+  }
+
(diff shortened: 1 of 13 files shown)


lib/src/front_end/chain_builder.dart
```diff
@@ -0,0 +1,206 @@
...
+  Piece build() {
+    // If there are no calls, there's no chain.
+    if (_calls.isEmpty) return _target;
+
+    // Count the number of contiguous properties at the beginning of the chain.
+    var leadingProperties = 0;
+    while (leadingProperties < _calls.length &&
+        _calls[leadingProperties].type == CallType.property) {
+      leadingProperties++;
+    }
```
natebosch, line 76: [optional] I usually check if we can avoid manual indexing, I don't know if the alternative is any more readable though.

```suggestion
    var leadingProperties =
        _calls.indexWhere((call) => call.type != CallType.property);
    if (leadingProperties < 0) leadingProperties = _calls.length;
```

lib/src/front_end/chain_builder.dart
```diff
@@ -0,0 +1,206 @@
...
+          });
+        });
+
+      default:
+        // Otherwise, it isn't a selector so we've reached the target.
+        _visitTarget(expression);
+    }
+  }
+
+  /// Creates a Piece for [target].
```
natebosch, line 187: [nit] Can you expand this comment to describe the side effects.

lib/src/piece/chain.dart
```diff
@@ -5,43 +5,205 @@ import '../back_end/code_writer.dart';
...
+///         .rotate();
+///
+/// [State.split] Split before every `.` and indent the chain, like:
+///
+///     target
+///         .getter
+///         .method(
+///           argument,
+///           argument,
+///         ).another.method(
```
natebosch, line 46: Comment says before every `.`, but there is no split before the second `.method`.

natebosch (approved):

lib/src/front_end/chain_builder.dart
munificent: I try to avoid indexing too but in this case the desired result is a number, so I think it makes sense. I think I like the current code a little more. If `Iterable` had something like `countWhile()`, I would have used that.

lib/src/piece/chain.dart
munificent: Oops, typo. :) Fixed.

lib/src/front_end/piece_factory.dart
```diff
@@ -573,6 +573,29 @@ mixin PieceFactory {
     return builder.build();
   }
 
+  /// Creates a [Piece] for an index expression whose [target] has already been
+  /// converted to a piece.
+  ///
+  /// The [target] may be `null` if [index] is an index expression for a
+  /// cascade section.
+  Piece createIndexExpression(Piece? target, IndexExpression index) {
+    // TODO(tall): Consider whether we should allow splitting between
```
kallentu, line 582: Just my two cents, but I think we should. At least reading the first postfix test, I find myself expecting a split.

If this doesn't happen, is there just no split at all for all chained index expressions?

test/invocation/chain_target.stmt
```diff
@@ -0,0 +1,161 @@
...
+[
+  element1,
+  element2,
+  element3,
+  element4,
+  element5,
+].method().prop.another(
+  argument1,
+  argument2,
+);
```
kallentu, line 64: Is this similar to the example you had in your method chains doc?
```
// Don't split the chain, mix split and unsplit argument lists:
target.prop1.method(
  arg,
).prop2.method2(
  arg,
);
// prop2 is hidden.
```

Why do we not put `method()` and `prop` on a newline?

kallentu (approved):

lib/src/front_end/piece_factory.dart
munificent: Yeah, if you have a long series of index operations, it is nice to be able to split. But, outside of weird contrived formatter tests, they're pretty rare. I think we probably will end up supporting in the new style (the old one does), but it was kind of tricky so I figured I would leave a TODO and kick it down the road a bit.

>  is there just no split at all for all chained index expressions?

Correct, it just keeps on going on one line.

test/invocation/chain_target.stmt
munificent: The difference here is that the leading block formatted thing—in this case a list literal—is the target expression and not part of the call chain itself (which is just the `.` parts following the target). These are very similar syntactically:

```dart
foo.bar(
  argument,
).prop.another(
  argument,
);

bar(
  argument,
).prop.another(
  argument,
);
```

In the former, the target is just `foo` and `.bar()` is the first call in the chain. In the latter, the target is `bar(argument,)` and that entire call isn't considered part of the call chain. We handle those differently because in the former case, we have the option of splitting between the target and the first call:


```dart
foo
    .bar(
      argument,
    )
    .prop
    .another(
      argument,
    );
```

But in the latter, there's no way to split *before* the `bar(...)`:

```dart
???
    bar(
      argument,
    )
    .prop
    .another(
      argument,
    );
```

The best we could do is like:

```dart
bar(
      argument,
    )
    .prop
    .another(
      argument,
    );
```

Which we do sometimes, but it looks pretty weird, so the formatter tries to avoid that.

The TL;DR: is that if the target of a call chain is itself a function call (with no leading target), it *looks* like it's part of the call chain, but it's not really and we handle it kind of differently. We end up handling it much more like like a target that's a collection literal.

PR #1358, opened 2024-01-12
Prefer splitting before case in an if-case.

kallentu:
This is previous behaviour from the old formatter that would prefer splitting at the `case` keyword.
We want to retain this behaviour and I added a few tests for this.
We also need this because it seems to be used by most pattern tests.


natebosch (approved):

munificent (approved):
Looks great!

PR #1359, opened 2024-01-12
Add macro flag note to CHANGELOG.

munificent:


kallentu (approved):

devoncarew:
It would be great to publish a version of the formatter that had support for extension types by default; right now to format `3.3` code w/ this package you have to know to pass in `DartFormatter(experimentFlags: ['inline-class'])`.

munificent:
Good catch. I'll try to kick a version out soon, but I want to land a fix (#1405) for #1404 first and roll that into the SDK.

munificent:
This is published now.

PR #1361, opened 2024-01-17
Format declared variable pattern.

kallentu:
- Added a `pattern` test directory.
- Implemented `visitDeclaredVariablePattern`'s splitting using an `AssignPiece`.
- Added `_indentInValue` to `AssignPiece` which adds extra indentation to the `value` when split and should be used for all patterns that have the similar splitting and indenting as declared variable patterns.

Indentation for some of the comment cases are a little weird, like this one:
```dart
if (obj
    case true // comment
        ) {
  ;
}
```
But this is probably fine for how much this occurs.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -173,7 +173,7 @@ class CodeWriter {
   ///
   /// Replaces any previous indentation set by this piece.
   ///
-  /// TODO(tall): Add another API that adds/subtracts existing indentation.
+  // TODO(tall): Add another API that adds/subtracts existing indentation.
   void setIndent(int indent) {
     _options.indent = _pieceOptions[_pieceOptions.length - 2].indent + indent;
   }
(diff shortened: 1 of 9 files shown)


lib/src/back_end/code_writer.dart
```diff
@@ -172,6 +172,8 @@ class CodeWriter {
   /// piece to [indent], relative to the indentation of the surrounding piece.
   ///
   /// Replaces any previous indentation set by this piece.
+  ///
+  /// TODO(tall): Add another API that adds/subtracts existing indentation.
```
munificent, line 176: Tiny nit: "///" -> "//" since the TODO isn't really part of the doc comment.

munificent (approved):
Perfect, I love it. :D

lib/src/back_end/code_writer.dart
kallentu: Oh yes, thanks for catching that. Fixed.

PR #1362, opened 2024-01-17
Format multi-line strings and string interpolation.

munificent:
In the old style, the formatter has some special code to discard line splits that occur inside string interpolation expressions. That's largely for historical reasons because the formatter initially didn't support formatting of string interpolation expressions *at all* and I didn't want too much churn when adding support for formatting them.

In the new style here, we don't do that: The contents of a string interpolation expression are split like any other expression. In practice, it doesn't matter much since users generally reorganize their code to avoid long strings and splits in string interpolation. This way leads to less special case code in the formatter.

This change is somewhat large because I also reorganized how newlines inside lexemes are handled in general. Previously, TextPiece stored a list of "lines" to handle things like line comments preceding or following a token. But it was also possible for a single "line" string in that list to internally contain newline characters because of multi-line strings or block comments.

But those internal newlines also need to force surrounding code to split, so there was this "_containsNewline" bit that had to be plumbed through and tracked. Even so, there were latent bugs where the column calculation in CodeWriter would be incorrect if a line contained internal newlines because it just used to the length of the entire "line" string.

With this change, the "_lines" list in TextPiece really is a list of lines. We eagerly split any incoming lexeme into multiple lines before writing it to the TextPiece. I think the resulting code is simpler, it fixes the column calculation in CodeWriter, and it means the formatter will correctly normalize line endings even when they occur inside block comments or multiline strings.

This was a good time to test the line ending code, so I copied those existing tests over from short_format_test.dart. I went ahead and copied all of the unit tests from that file, even the ones not related to line endings, since they're all working and passing now.

This PR does *not* handle adjacent strings. Those have a decent amount of special handling not related to what's going on here, so I'll do those separately.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -156,14 +156,6 @@ extension ExpressionExtensions on Expression {
       expression = expression.expression;
     }
 
-    // TODO(tall): We should also allow multi-line strings to be formatted
-    // like block arguments, at least in some cases like:
-    //
-    //     function('''
-    //       Lots of
-    //       text
-    //     ''');
-
     // TODO(tall): Consider whether immediately-invoked function expressions
     // should be block argument candidates, like:
     //
@@ -177,6 +169,8 @@ extension ExpressionExtensions on Expression {
         parameters.parameters.canSplit(parameters.rightParenthesis) ||
             (body is BlockFunctionBody &&
                 body.block.statements.canSplit(body.block.rightBracket)),
+
+      // Non-empty collection literals can block split.
       ListLiteral(:var elements, :var rightBracket) ||
       SetOrMapLiteral(:var elements, :var rightBracket) =>
         elements.canSplit(rightBracket),
@@ -184,9 +178,17 @@ extension ExpressionExtensions on Expression {
         fields.canSplit(rightParenthesis),
       SwitchExpression(:var cases, :var rightBracket) =>
         cases.canSplit(rightBracket),
+
+      // Function calls can block split if their argument lists can.
       InstanceCreationExpression(:var argumentList) ||
       MethodInvocation(:var argumentList) =>
         argumentList.arguments.canSplit(argumentList.rightParenthesis),
+
+      // Multi-line strings can.
+      StringInterpolation(isMultiline: true) => true,
+      SimpleStringLiteral(isMultiline: true) => true,
+
+      // Parenthesized expressions can if the inner one can.
       ParenthesizedExpression(:var expression) => expression.canBlockSplit,
       _ => false,
     };
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -199,6 +199,15 @@ class CodeWriter {
         flushLeft: flushLeft);
   }
 
+  /// Queues [whitespace] to be written to the output.
+  ///
+  /// If any non-whitespace is written after this call, then this whitespace
+  /// will be written first. Also handles merging multiple kinds of whitespace
+  /// intelligently together.
+  ///
+  /// If [flushLeft] is `true`, then the new line begins at column 1 and ignores
+  /// any surrounding indentation. This is used for multi-line block comments
+  /// and multi-line strings.
   void whitespace(Whitespace whitespace, {bool flushLeft = false}) {
     if (whitespace case Whitespace.newline || Whitespace.blankLine) {
       _handleNewline();
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -128,22 +128,6 @@ class CodeWriter {
         isValid: !_hasInvalidNewline, overflow: _overflow, cost: _cost);
   }
 
-  /// Notes that a newline has been written.
-  ///
-  /// If this occurs in a place where newlines are prohibited, then invalidates
-  /// the solution.
-  ///
-  /// This is called externally by [TextPiece] to let the writer know some of
-  /// the raw text contains a newline, which can happen in multi-line block
-  /// comments and multi-line string literals.
-  void handleNewline() {
-    if (!_options.allowNewlines) _hasInvalidNewline = true;
-
-    // Note that this piece contains a newline so that we can propagate that
-    // up to containing pieces too.
-    _options.hasNewline = true;
-  }
-
   /// Appends [text] to the output.
   ///
   /// If [text] contains any internal newlines, the caller is responsible for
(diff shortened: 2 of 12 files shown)


test/expression/string.stmt
```diff
@@ -0,0 +1,88 @@
...
+multiline
+""";
+}
+<<<
+{
+  """
+multiline
+""";
+}
+>>>
```
kallentu, line 75: Nit: Write a small description for this test?

kallentu (approved):

test/expression/string.stmt
munificent: Done. (I tended to omit the description if a test case is for the same reason as the preceding one, but I just realized I never once explained that convention and maybe it's not worth keeping. :) )

lib/src/ast_extensions.dart
```diff
@@ -177,16 +169,26 @@ extension ExpressionExtensions on Expression {
...
         cases.canSplit(rightBracket),
+
+      // Function calls can block split if their argument lists can.
       InstanceCreationExpression(:var argumentList) ||
       MethodInvocation(:var argumentList) =>
         argumentList.arguments.canSplit(argumentList.rightParenthesis),
+
+      // Multi-line strings can.
+      StringInterpolation(isMultiline: true) => true,
+      SimpleStringLiteral(isMultiline: true) => true,
```
natebosch, line 189: [nit] Lines 174 and 183 use `||`, consider doing the same for consistency.

lib/src/back_end/code_writer.dart
```diff
@@ -204,16 +188,21 @@ class CodeWriter {
...
+  void newline({bool blank = false, int? indent, bool flushLeft = false}) {
     if (indent != null) setIndent(indent);
 
-    whitespace(blank ? Whitespace.blankLine : Whitespace.newline);
+    whitespace(blank ? Whitespace.blankLine : Whitespace.newline,
+        flushLeft: flushLeft);
   }
 
-  void whitespace(Whitespace whitespace) {
+  void whitespace(Whitespace whitespace, {bool flushLeft = false}) {
```
natebosch, line 202: [nit] Can you add a doc comment?

lib/src/front_end/piece_writer.dart
```diff
@@ -180,14 +196,31 @@ class PieceWriter {
       _currentText = TextPiece();
     }
 
-    _write(lexeme ?? token.lexeme, offset: token.offset);
+    lexeme ??= token.lexeme;
+
+    _write(lexeme, offset: token.offset);
+  }
+
+  /// Writes multi-line [text] to the current [TextPiece].
```
natebosch, line 204: What `[text]` is this referencing? Should this be `[lexeme]`? Or IMO change the argument name back to `text`.

lib/src/front_end/piece_writer.dart
```diff
@@ -180,14 +196,31 @@ class PieceWriter {
...
+  ///
+  /// Handles breaking [text] into lines and adding them to the [TextPiece].
+  Piece _writeMultiLine(String lexeme, {required int offset}) {
```
natebosch, line 207: Can you mention `offset` in the doc?

lib/src/piece/piece.dart
```diff
@@ -210,7 +208,26 @@ class TextPiece extends Piece {
   }
 
   @override
-  String toString() => '`${_lines.join('¬')}`${_containsNewline ? '!' : ''}';
+  String toString() => '`${_lines.join('¬')}`';
+}
+
+/// A single line of text within a [TextPiece].
+class _Line {
+  String _text = '';
```
natebosch, line 216: [nit] I think it's overkill to use private fields in a private class. The rest of the library reads some of these directly, so making them private here does not indicate that they are only used within the class definition.

lib/src/piece/piece.dart
```diff
@@ -210,7 +208,26 @@ class TextPiece extends Piece {
...
+
+  /// Whether this line should start at column one or use the surrounding
+  /// indentation.
+  final bool _isFlushLeft;
+
+  _Line({required bool flushLeft}) : _isFlushLeft = flushLeft;
+
+  void append(String text) {
+    // TODO(perf): Consider a faster way of accumulating text.
```
natebosch, line 225: Adding the `StringBuffer` now should be trivial - I think the introduction of a new abstraction for a line is a good reason to handle this TODO.

test/tall_format_test.dart
```diff
@@ -20,7 +21,118 @@ void main() async {
...
+
+  test('FormatterException.message() does not throw', () {
+    // This is a regression test for #358 where an error whose position is
+    // past the end of the source caused FormatterException to throw.
+    try {
+      DartFormatter().format('library');
+    } on FormatterException catch (err) {
+      var message = err.message();
+      expect(message, contains('Could not format'));
+    }
```
natebosch, line 37: [nit] Prefer to avoid `try/catch` in a test.

```suggestion
    expect(
        () => DartFormatter().format('library'),
        throwsA(isA<FormatterException>().having(
            (e) => e.message(), 'message', contains('Could not format'))));
```

Unless you specifically want the test to pass when no exception is thrown at all. In which case I think the following expresses the intent more explicitly.

```dart
    expect(
        () => DartFormatter().format('library'),
        anyOf(
            throwsA(isA<FormatterException>().having(
                (e) => e.message(), 'message', contains('Could not format'))),
            not(throwsA(anything))));
```

test/tall_format_test.dart
```diff
@@ -20,7 +21,118 @@ void main() async {
...
+  test('FormatterException describes parse errors', () {
+    try {
+      DartFormatter().format('''
+
+      var a = some error;
+
+      var b = another one;
+      ''', uri: 'my_file.dart');
+
+      fail('Should throw.');
```
natebosch, line 49: Consider using `throwsA`

test/tall_format_test.dart
```diff
@@ -20,7 +21,118 @@ void main() async {
...
+
+  test('does not add newline to statement', () {
+    expect(DartFormatter().formatStatement('var x = 1;'), equals('var x = 1;'));
+  });
+
+  test('fails if anything is after the statement', () {
+    try {
+      DartFormatter().formatStatement('var x = 1;;');
+
+      fail('Should throw.');
```
natebosch, line 79: Consider using `throwsA`

natebosch (approved):

lib/src/ast_extensions.dart
munificent: I used `||` for those because they bind variables and then share a body that uses those variables, so there's some benefit to using the same body.

Here, though, it would actually be longer to share a body because then it moves the `=> true` to the next line.

lib/src/front_end/piece_writer.dart
munificent: Oops, yeah, copy/paste error. Changed to `text`.

lib/src/piece/piece.dart
munificent: I've gone back and forth on this. I used to just make everything public if the class was private (and non-escaping).

But over time, I've appreciated the "unused declaration" warnings more and more and you only get those for private members. So I tend to use private for anything I can that way the analyzer helps me clean up dead code better.

lib/src/piece/piece.dart
munificent: Trivial, yes, but I don't know whether it will help or harm performance. Every single lexeme in the source program will go through this function, so it's a performance bottleneck. 99+% of TextPieces will only ever have a single token added so the overhead of creating a StringBuffer might be net loss.

I left this comment in here so that I can circle back and investigate it once I have enough of the language supported to run real benchmarks.

test/tall_format_test.dart
munificent: Good idea. I'll apply this locally since I have other changes in progress.

PR #1363, opened 2024-01-19
Format list pattern and rest pattern.

kallentu:
Changes
- Add an extension to `DartPattern` so we can block split `ListPattern`s.
- Implemented `visitListPattern` and `visitRestPattern`.
- Avoid extra indentation in `AssignPiece` when we are (inner) block splitting.
- Migrated existing list pattern tests.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -317,11 +317,9 @@ extension PatternExtensions on DartPattern {
   /// expressions that allows "block-like" formatting in some contexts.
   ///
   /// See [ExpressionExtensions.canBlockSplit].
-  bool get canBlockSplit {
-    return switch (this) {
-      ListPattern(:var elements, :var rightBracket) =>
-        elements.canSplit(rightBracket),
-      _ => false,
-    };
-  }
+  bool get canBlockSplit => switch (this) {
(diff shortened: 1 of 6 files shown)


lib/src/ast_extensions.dart
```diff
@@ -311,3 +311,17 @@ extension CascadeExpressionExtensions on CascadeExpression {
...
   }
 }
+
+extension PatternExtensions on DartPattern {
+  /// Whether this expression is a non-empty delimited container for inner
+  /// expressions that allows "block-like" formatting in some contexts.
+  ///
+  /// See [ExpressionExtensions.canBlockSplit].
+  bool get canBlockSplit {
+    return switch (this) {
```
natebosch, line 321: [optional] Consider `=>` over `{return}`.

natebosch (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -119,9 +119,14 @@ mixin PieceFactory {
   }
 
   /// Creates a [ListPiece] for a collection literal.
```
munificent, line 121: Nit: "for a collection literal" -> "for a collection literal or pattern".

munificent (approved):
Ah, this looks so good!

Issue #684, opened 2018-04-25
Adjacent strings should never be... adjacent to each other


srawlins:
Hopefully not alone on this one:

```dart
var a = 'Multiline strings keep leading space\n' 'I write adjacent strings.\n';
```

I am using adjacent strings here, instead of a multiline string, because of leading whitespace. So I want the strings on separate lines. I'd expect this to format as:

```dart
var a = 'Multiline strings keep leading space\n'
    'I write adjacent strings.\n';
```

But it doesn't. I think 99% of adjacent string usage is where multiline strings would work, but the leading whitespace retention makes it unattractive.

The only use case I can think of for adjacent strings on the same line is mixing raw and not raw (or mixing double and single quotes):

```dart
var a = '$var' r'$money' 'newline\n' r'regexp \s*\d+\Z';
```

Is that a thing (not that it's super readable either)? Anyways, I can't see a way to force newlines between my two strings (no trailing comma trick here). Even comments in between don't force it :frowning_face: 

```dart
var b = 'text' /* comment */ 'more text' /* comment */;
```

a14n:
What about wrapping when a string ends with `\n`?

srawlins:
Bob, I'm assuming this is a super tiny change. If you can paste a patch here, I'd be happy to run it over (all) internal code and see what comes up.

munificent:
@a14n, that's a really interesting idea. Definitely worth investigating.

> Bob, I'm assuming this is a super tiny change.

I *think* so, but it may be a little tricky getting things like the indentation correct. Adjacent strings have been a difficult area of the formatter to work on because it's often not clear what kind of indentation is desired in what contexts or how to ensure that's what users get.

nshahan:
There is a pattern in DDC that tickles this issue every once and a while when I'm missing a comma between two adjacent string arguments and they accidentally become adjacent strings on the same line. I wish they would get formatted to separate lines because then I feel like the missing comma would be easier to notice.

Also, there is a violation of the reversible property of the formatter here that I think would go away if adjacent strings were always forced to separate lines.

munificent:
Your wish is granted (admittedly after a long delay). The [forthcoming](https://github.com/dart-lang/dart_style/issues/1403) [tall style](https://github.com/dart-lang/dart_style/issues/1253) always forces newlines between adjacent strings. In other words, it has the principle that the only reason to use adjacent strings is to split them across lines. With the new formatter, you get:

```dart
var a =
[...]

PR #1364, opened 2024-01-23
Format adjacent strings.

munificent:
Most of this was fairly straightforward, but the two tricky bits are:

### 1. Deciding whether or not to indent

There are some rules around whether subsequent strings in the adjacent strings get indented or not. The answer is yes in some cases to avoid confusion:

```
var list = function(
  'string 1',
  'adjacent'
      'string 2',
  'string 3',
];
```

But not in others since it looks nicer to line them up when possible:

```
var description =
    'some text '
    'more text';
```

### 2. Handling `test()` and `group()`

It's really important that test functions don't end up fully split because doing so would lead to the inner function expression getting indented +2:

```
test('this looks good', () {
  body;
});

test(
  'this looks bad',
  () {
    body;
  },
);
```

Test descriptions often span multiple lines using adjacent strings:

```
test('this is a very long test description '
    'spanning multiple lines, () {
  body;
});
```

Normally, the newline inside the adjacent strings would cause the entire argument list to split. The old style handles that (I think) by allowing multiple block-formatted arguments and then treating both the adjacent strings and the function expressions as block arguments.

The new style currently only allows a single block argument (because in almost all of the Flutter code I found using block formatting, one argument was sufficient). So I chose a more narrowly targeted rule here where we allow adjacent strings to not prevent block formatting only if the adjacent strings are the first argument and the block argument is a function expression as the next argument.

I left a TODO to see if we want to iterate on that rule, but I think it works pretty well.

### Other stuff

Unlike the old style, I chose to always split between adjacent strings. The old style will preserve newlines there but if a user chooses to deliberately put multiple adjacent strings on the same line and they fit, it will honor it. That didn't seem useful to me, so now they just always split. I don't think adjacent strings ever look good on the same line.

I ended up moving the state to track which elements in a ListPiece out of ListPiece and into the ListElements themselves. I think it's clearer this way and will be easier to evolve if we end up supporting multiple block formatted elements in a single list.

### DIFF
--- a/lib/src/front_end/delimited_list_builder.dart
+++ b/lib/src/front_end/delimited_list_builder.dart
@@ -52,7 +52,7 @@ class DelimitedListBuilder {
   /// Creates the final [ListPiece] out of the added brackets, delimiters,
   /// elements, and style.
   ListPiece build() {
-    _selectBlockElement();
+    _setBlockElementFormatting();
 
     var piece =
         ListPiece(_leftBracket, _elements, _blanksAfter, _rightBracket, _style);
@@ -154,7 +154,7 @@ class DelimitedListBuilder {
 
     // See if it's an expression that supports block formatting.
     var format = switch (element) {
-      AdjacentStrings() when element.indentStrings =>
+      AdjacentStrings(indentStrings: true) =>
         BlockFormat.indentedAdjacentStrings,
       AdjacentStrings() => BlockFormat.unindentedAdjacentStrings,
       FunctionExpression() when element.canBlockSplit => BlockFormat.function,
@@ -417,8 +417,8 @@ class DelimitedListBuilder {
   ///     );
   ///
   /// Stores the result of this calculation by setting flags on the
-  /// [ListElements].
-  void _selectBlockElement() {
+  /// [ListElement]s.
+  void _setBlockElementFormatting() {
     // TODO(tall): These heuristics will probably need some iteration.
     var functions = <int>[];
     var others = <int>[];
@@ -438,13 +438,12 @@ class DelimitedListBuilder {
       }
     }
 
-    // A function expression takes precedence over other block arguments.
-    if (functions.length == 1) {
-      // We only block formatting in an argument list containing adjacent
+    switch ((functions, others, adjacentStrings)) {
+      // Only allow block formatting in an argument list containing adjacent
       // strings when:
       //
       // 1. The block argument is a function expression.
-      // 2. If is the second argument, following an adjacent strings expression.
+      // 2. It is the second argument, following an adjacent strings expression.
       // 3. There are no other adjacent strings in the argument list.
       //
       // This matches the `test()` and `group()` and other similar APIs where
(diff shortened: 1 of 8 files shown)


srawlins:
> Unlike the old style, I chose to always split between adjacent strings.

Sounds like this fixes https://github.com/dart-lang/dart_style/issues/684. Not sure how bugs are being tracked between new and existing formatter...

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -155,6 +154,9 @@ class DelimitedListBuilder {
 
     // See if it's an expression that supports block formatting.
     var format = switch (element) {
+      AdjacentStrings() when element.indentStrings =>
```
natebosch, line 157:
```suggestion
      AdjacentStrings(indentStrings: true) =>
```

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -387,32 +389,88 @@ class DelimitedListBuilder {
...
+  ///     test(
+  ///       'some long description'
+  ///       'split across multiple lines',
+  ///       () {
+  ///         expect(1, 1);
+  ///       },
+  ///     );
+  ///
+  /// Stores the result of this calculation by setting flags on the
+  /// [ListElements].
```
natebosch, line 419: [nit] `ListElements` should not be linkable - do you mean `[ListElement]s`?

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -387,32 +389,88 @@ class DelimitedListBuilder {
...
+        case BlockFormat.unindentedAdjacentStrings:
+          adjacentStrings.add(i);
         case BlockFormat.none:
           break; // Not a block element.
       }
     }
 
     // A function expression takes precedence over other block arguments.
-    if (functions.length == 1) return functions.first;
+    if (functions.length == 1) {
```
natebosch, line 441: [optional] 
```suggestion
    if (functions case [var singleFunction]) {
```
I think the name `singleFunction` will be more informative than `functions.first` below.

lib/src/piece/adjacent_strings.dart
```diff
@@ -0,0 +1,35 @@
...
+
+/// Piece for a series of adjacent strings, like:
+///
+///     var message =
+///         'This is a long message '
+///         'split into multiple strings';
+class AdjacentStringsPiece extends Piece {
+  final List<Piece> _strings;
+
+  /// Whether string after the first should be indented.
```
natebosch, line 16: [nit] 
```suggestion
  /// Whether strings after the first should be indented.
```

natebosch (approved):

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -387,32 +389,88 @@ class DelimitedListBuilder {
...
+      // We only block formatting in an argument list containing adjacent
+      // strings when:
+      //
+      // 1. The block argument is a function expression.
+      // 2. If is the second argument, following an adjacent strings expression.
```
kallentu, line 446: Typo in "If is the second argument".

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -387,32 +389,88 @@ class DelimitedListBuilder {
...
+  ///       'some long description'
+  ///       'split across multiple lines',
+  ///       () {
+  ///         expect(1, 1);
+  ///       },
+  ///     );
+  ///
+  /// Stores the result of this calculation by setting flags on the
+  /// [ListElements].
+  void _selectBlockElement() {
```
kallentu, line 420: I feel like this name doesn't describe what's being done. Could we name it something like `_setBlockElementFormatting`?

kallentu (approved):

munificent:
> Sounds like this fixes #684. Not sure how bugs are being tracked between new and existing formatter...

I haven't really figured that out either. :) I'll probably just do a pass through the issue tracker once we've moved to the new style and close anything that's done. It feels weird to close them now because the new style is hidden behind an experiment flag.

lib/src/front_end/delimited_list_builder.dart
munificent: Good idea! I'll do this locally with the other changes.

I'm so glad we decided to let object pattern fields call extensions.

lib/src/front_end/delimited_list_builder.dart
munificent: Ooh, good idea. Went all the way:

```dart
    switch ((functions, others, adjacentStrings)) {
      case ([1], _, [0]):
        // The adjacent strings.
        _elements[0].allowNewlines = true;
        if (_elements[0].blockFormat == BlockFormat.unindentedAdjacentStrings) {
          _elements[0].indentWhenBlockFormatted = true;
        }

        // The block-formattable function.
        _elements[1].allowNewlines = true;

      // A function expression takes precedence over other block arguments.
      case ([var blockArgument], _, _):
      // Otherwise, if there is single block argument, it can be block formatted.
      case ([], [var blockArgument], _):
        _elements[blockArgument].allowNewlines = true;
    }
```

lib/src/front_end/delimited_list_builder.dart
munificent: I didn't love the name either, but couldn't come up with anything better. I like your suggestion. Done.

PR #1365, opened 2024-01-25
Format cascades.

munificent:
There were a couple of tricky corners, but this wasn't as bad as I'd feared. Most of ChainBuilder worked great for it.

In the process of doing this and migrating the tests, I noticed another kind of expression that should allow block formatting in an argument list: function expression invocations.

That's a `(...)` where the thing on the left is some kind of expression and not simply a name, like: `(foo + bar)(args)`.

Added support for that and added some tests of block formatting for those too.

I believe this is the last corner of the grammar where there might be some weird tricky formatting logic to implement. There's a lot still left to cover, but I think it's all syntactically similar to other already-implemented constructs. I'm breathing a  big sigh of relief because I think this means that the new Piece representation really is going to work out. 😌

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -185,7 +185,12 @@ extension ExpressionExtensions on Expression {
 
       // Function calls can block split if their argument lists can.
       InstanceCreationExpression(:var argumentList) ||
-      MethodInvocation(:var argumentList) ||
+      MethodInvocation(:var argumentList) =>
+        argumentList.arguments.canSplit(argumentList.rightParenthesis),
+
+      // Note: Using a separate case instead of `||` for this type because
+      // Dart 3.0 reports an error that [argumentList] has a different type
+      // here than in the previous two clauses.
       FunctionExpressionInvocation(:var argumentList) =>
         argumentList.arguments.canSplit(argumentList.rightParenthesis),
 
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -163,6 +163,10 @@ extension ExpressionExtensions on Expression {
     //       body;
     //     }());
     return switch (expression) {
+      // Allow the target of a single-section cascade to be block formatted.
+      CascadeExpression(:var target, :var cascadeSections) =>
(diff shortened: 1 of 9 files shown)


lib/src/ast_extensions.dart
```diff
@@ -184,6 +188,12 @@ extension ExpressionExtensions on Expression {
       MethodInvocation(:var argumentList) =>
         argumentList.arguments.canSplit(argumentList.rightParenthesis),
 
+      // Note: Using a separate case instead of `||` for this type because
+      // Dart 3.0 reports an error that [argumentList] has a different type
```
natebosch, line 192: Is this fixed in a more recent Dart SDK? Can we bump the constraint instead of hacking around the older SDK?

lib/src/front_end/chain_builder.dart
```diff
@@ -60,11 +68,40 @@ class ChainBuilder {
...
+        _unwrapCall(section);
+      }
+    } else {
+      _unwrapCall(_root);
+    }
   }
 
   Piece build() {
+    if (_root case CascadeExpression cascade) {
+      // If there is only a single section and it can block split, allow it:
```
natebosch, line 85: I don't see any code I can correlate to "if there is only a single section". Should this read "If the last section can block split, allow it"?

natebosch (approved):

kallentu (approved):

lib/src/ast_extensions.dart
munificent: > Is this fixed in a more recent Dart SDK?

I assume so? I only noticed this because it failed the 3.0.0 CI when I uploaded the PR, but didn't fail the newer one.

> Can we bump the constraint instead of hacking around the older SDK?

I can, but the change here is so innocuous and messing with the CI config so annoying that I'm inclined to just leave this as-is.

lib/src/front_end/chain_builder.dart
munificent: Oops good catch. In practice, this only works if there's only a single section (because cascades are pinned to split if there is more than one), but this code is unnecessarily complex. Fixed.

PR #1366, opened 2024-01-25
Format super expressions.

munificent:
Apparently the old formatter has *zero* tests of this. I went ahead and wrote tests of all of the ways `super` appears in the grammar. That's probably overkill given that the analyzer models `super` as a standalone expression, but that's not actually how it works in the language itself. (In Dart, `print(super)` isn't syntactically valid.)

I figured it made sense to write tests closer to the grammar.


natebosch (approved):

kallentu (approved):
Amazing.
Also, haha these tests are extensive.

munificent:
> Also, haha these tests are extensive.

Take a look for `'super'` in the grammar. It's all over the place!

PR #1368, opened 2024-01-29
Format map patterns.

kallentu:
- Map patterns and entries are now formatted.
- Migrated map pattern tests.
- Altered `createAssignment` so it can take `AstNode`s and calculate `allowInnerSplit` depending on which node.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -390,6 +390,8 @@ extension PatternExtensions on DartPattern {
   bool get canBlockSplit => switch (this) {
         ListPattern(:var elements, :var rightBracket) =>
           elements.canSplit(rightBracket),
+        MapPattern(:var elements, :var rightBracket) =>
+          elements.canSplit(rightBracket),
         _ => false,
       };
 }
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -884,8 +884,7 @@ mixin PieceFactory {
   ///
   /// If [allowInnerSplit] is `true`, then a newline inside the target or
   /// right-hand side doesn't force splitting at the operator itself.
-  Piece createAssignment(
-      AstNode target, Token operator, Expression rightHandSide,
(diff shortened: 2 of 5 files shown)


lib/src/ast_extensions.dart
```diff
@@ -390,6 +390,8 @@ extension PatternExtensions on DartPattern {
   bool get canBlockSplit => switch (this) {
         ListPattern(:var elements, :var rightBracket) =>
           elements.canSplit(rightBracket),
+        MapPattern(:var elements, :var rightBracket) =>
```
natebosch, line 393: Do both variable assignments have the same type? Consider using `||` to merge the two cases.

natebosch (approved):

lib/src/ast_extensions.dart
kallentu: They don't, sadly. `ListPatternElement`s and `MapPatternElement`s. Thanks for the reminder though.

lib/src/front_end/piece_factory.dart
```diff
@@ -896,7 +895,11 @@ mixin PieceFactory {
     //    var list = [
     //      element,
     //    ];
-    allowInnerSplit |= rightHandSide.canBlockSplit;
+    allowInnerSplit |= switch (rightHandSide) {
+      Expression() => rightHandSide.canBlockSplit,
+      DartPattern() => rightHandSide.canBlockSplit,
```
munificent, line 900: Another option is to move `canBlockSplit` to be an extension on AstNode instead of separate extensions on Expression and DartPattern and then effectively move this type switch into there. But keeping it like this is fine too if you prefer it that way.

munificent (approved):
Looks great!

lib/src/front_end/piece_factory.dart
kallentu: That's a good point. I'm okay having them as separate extensions right now. But if it gets a little unwieldy, I'll make a CL to combine the two.

PR #1369, opened 2024-01-29
Combine SequencePiece and BlockPiece.

munificent:
SequencePiece is used whenever there is a series of AST nodes separated by mandatory newlines: block bodies, members in class bodies, the top level of a compilation unit, etc.

Since the top level of a compilation unit doesn't have any enclosing curly braces or indentation, I originally designed it as two separate pieces: SequencePiece has the series of elements and it can optionally by wrapped in a BlockPiece that has the brackets and indentation.

It's a little weird, though, because BlockPiece is the one that decides whether or not the contents split even though SequencePiece contains the actual contents. And it doesn't match ListPiece which uses a single unified Piece for both the delimiters and contents.

So I changed it to be more like ListPiece which has both the brackets and the contents, and the brackets are optional in places where they aren't needed. I think this makes the design a little more consistent and simpler. (Also, I have some in-progress optimization work that *might* be easier with SequencePiece organized this way.)

At some point, we might even be able to reuse code between SequencePiece and ListPiece (and likewise Sequencebuilder and DelimitedListBuilder), but I'm not worrying about that right now.


lib/src/piece/sequence.dart
```diff
@@ -3,23 +3,41 @@
...
 
   @override
   void format(CodeWriter writer, State state) {
+    writer.setAllowNewlines(state == State.split);
+
+    if (_leftBracket case var leftBracket?) {
+      writer.format(leftBracket);
+      writer.splitIf(state == State.split,
+          space: false,
+          indent: _elements.isNotEmpty ? _elements.first.indent : 0);
```
natebosch, line 38: How about

```suggestion
          indent: _elements.firstOrNull?.indent ?? 0);
```

https://api.dart.dev/stable/3.2.6/dart-collection/IterableExtensions/firstOrNull.html

natebosch (approved):

kallentu (approved):

lib/src/piece/sequence.dart
munificent: Nice! Done.

PR #1371, opened 2024-01-30
Format record pattern.

kallentu:
- Added the helper `createRecordCollection` to handle similar trailing comma situations for record types and patterns.

Records should only have an always trailing comma with a single positional field, in every other scenario, it will have a regular trailing comma style (if split).

- Migrated record pattern tests from the temp repo.
- Allow `RecordPattern` to block split.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -715,34 +715,37 @@ mixin PieceFactory {
     Token? constKeyword,
   }) {
     var style = switch (fields) {
-      // Record patterns with named fields don't add trailing commas unless
-      // split, like:
+      // Record patterns with a single positional field will always
+      // have a trailing comma, like:
       //
-      //     if (obj case (name: value)) {
+      //     if (obj case (pattern,)) {
       //       ;
       //     }
-      [PatternField(name: _?)] => const ListStyle(commas: Commas.trailing),
-
-      // Record types with named fields don't add trailing commas unless split,
-      // like:
-      //
-      //     ({int n}) x;
-      [NamedExpression()] => const ListStyle(commas: Commas.trailing),
+      [PatternField(name: null)] =>
+        const ListStyle(commas: Commas.alwaysTrailing),
 
       // Record types with a single position field will always have a trailing
       // comma, like:
       //
       //     (int,) x;
       //
-      // Similarly, record patterns with a single positional field will always
-      // have a trailing comma, like:
+      [var field] when field is! NamedExpression && field is! PatternField =>
+        const ListStyle(commas: Commas.alwaysTrailing),
+
+      // All other record types and patterns have regular trailing commas when
+      // split.
       //
-      //     if (obj case (pattern,)) {
+      // Record types with a single named fields don't add trailing commas
+      // unless it's split, like:
+      //
+      //     ({int n}) x;
+      //
+      // Record patterns with a single named fields don't add trailing commas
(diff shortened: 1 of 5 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -707,6 +707,56 @@ mixin PieceFactory {
     );
   }
 
+  /// Creates a [ListPiece] for a record literal or pattern.
+  Piece createRecordCollection(
```
munificent, line 711: Nit: I'd probably just name this `createRecord`.

lib/src/front_end/piece_factory.dart
```diff
@@ -707,6 +707,56 @@ mixin PieceFactory {
...
+      // Record patterns with a single positional field will always
+      // have a trailing comma, like:
+      //
+      //     if (obj case (pattern,)) {
+      //       ;
+      //     }
+      [PatternField(name: null)] =>
+        const ListStyle(commas: Commas.alwaysTrailing),
+
+      // Record types with a single position field will always have a trailing
```
munificent, line 727: Nit: "position" -> "positional".

lib/src/front_end/piece_factory.dart
```diff
@@ -707,6 +707,56 @@ mixin PieceFactory {
...
+      //     ({int n}) x;
+      //
+      // Record patterns with a single named fields don't add trailing commas
+      // unless it's split, like:
+      //
+      //     if (obj case (name: value)) {
+      //       ;
+      //     }
+      _ => const ListStyle(commas: Commas.trailing)
+    };
```
munificent, line 750: Might be a little cleaner to test for the named-single-field cases first:

```dart
var style = switch (fields) {
  // Record types or patterns with a single named field don't add a trailing
  // comma unless it's split, like:
  //
  //     ({int n}) x;
  // 
  // Or:
  //
  //     if (obj case (name: value)) {
  //       ;
  //     }
  [NamedExpression field] => const ListStyle(commas: Commas.trailing),
  [PatternField(name: _?)] => const ListStyle(commas: Commas.trailing),

  // Record types or patterns with a single positional field always have a
  // trailing comma to disambiguate from parenthesized expressions or
  // patterns, like:
  //
  //     if (obj case (pattern,)) {
  //       ;
  //     }
  [_] => const ListStyle(commas: Commas.alwaysTrailing),

  // Record types or patterns with multiple fields have regular trailing
  // commas when split.
  _ => const ListStyle(commas: Commas.trailing)
};
```

Since switches stop after the first matching case, you can often eliminate guards by reordering cases. (Whether that leads to clearer code is a matter of taste. :) )

munificent (approved):
This looks great!

lib/src/front_end/piece_factory.dart
kallentu: Ah yes, that was exactly my initial switch actually. I ended up changing it to what it is now because I didn't know if writing it out that way was weirder than writing the edge cases (single positional) and leaving everything else.

Since with writing it this way, you end up reading from the trailing cases to always trailing to trailing again.

I'm okay with changing it back to how I had it originally (which is what you're suggesting).

kallentu:
@natebosch for a little readability review?

lib/src/front_end/piece_factory.dart
munificent: Yeah, either way is fine with me.

natebosch (approved):

PR #1372, opened 2024-01-31
Prune solutions derived from invalid ones.

munificent:
Prune solutions derived from invalid ones.

A solution is invalid if a child piece writes a newline at a point in time where a parent disallows newlines.

When that happens, the parent's may be explicitly bound to a state by the solution, or it may still be in an unbound state (and thus using the default). If unbound, then even though this solution is invalid, it may lead to later valid ones by binding that parent piece to a different state.

But once a solution binds a piece to a state, all solutions derived from that state keep the same binding. So if we hit an invalid solution and the piece whose newline constraint was already violated, we can discard that solution and all possible solutions derived from it.

This skips an entire potentially huge subtree from the solution space being explored.

In an example giant deeply nested function call from the benchmark, before this optimization, the solver tries 329,057 solutions before finding a winning one. With this optimization, it only takes 5,405.

I broke this PR into two commits to make it easier to review. The first commit is a refactoring that moves some code around to make the second commit easier to implement. Then after uploading that, I figured out a cleaner way to model invalid states and an idea for further optimization, so uploaded another commit with those changes.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -52,7 +52,7 @@ class CodeWriter {
   ///
   /// This is used to increase the cumulative nesting as we recurse into pieces
   /// and then unwind that as child pieces are completed.
-  final List<_PieceOptions> _pieceOptions = [_PieceOptions(0, true)];
+  final List<_PieceOptions> _options = [];
 
   /// Whether we have already found the first line where whose piece should be
   /// used to expand further solutions.
@@ -85,9 +85,6 @@ class CodeWriter {
   /// current line so that we know which piece should be expanded in the next
   /// solution if the line ends up overflowing.
   final List<Piece> _currentUnsolvedPieces = [];
-
-  /// The options for the current innermost piece being formatted.
-  _PieceOptions get _options => _pieceOptions.last;
 
   CodeWriter(this._pageWidth, this._solution);
 
@@ -134,7 +131,10 @@ class CodeWriter {
   ///
   // TODO(tall): Add another API that adds/subtracts existing indentation.
   void setIndent(int indent) {
-    _options.indent = _pieceOptions[_pieceOptions.length - 2].indent + indent;
+    var parentIndent =
+        _options.length > 1 ? _options[_options.length - 2].indent : 0;
+
+    _options.last.indent = parentIndent + indent;
   }
 
   /// Inserts a newline if [condition] is true.
@@ -188,7 +188,7 @@ class CodeWriter {
   void whitespace(Whitespace whitespace, {bool flushLeft = false}) {
     if (whitespace case Whitespace.newline || Whitespace.blankLine) {
       _handleNewline();
-      _pendingIndent = flushLeft ? 0 : _options.indent;
+      _pendingIndent = flushLeft ? 0 : _options.last.indent;
     }
 
     _pendingWhitespace = _pendingWhitespace.collapse(whitespace);
@@ -197,13 +197,14 @@ class CodeWriter {
   /// Sets whether newlines are allowed to occur from this point on for the
   /// current piece.
   void setAllowNewlines(bool allowed) {
-    _options.allowNewlines = allowed;
+    _options.last.allowNewlines = allowed;
   }
 
   /// Format [piece] and insert the result into the code being written and
   /// returned by [finish()].
   void format(Piece piece) {
-    _pieceOptions.add(_PieceOptions(_options.indent, _options.allowNewlines));
+    _options.add(_PieceOptions(piece, _options.lastOrNull?.indent ?? 0,
+        _options.lastOrNull?.allowNewlines ?? true));
 
     var isUnsolved = !_solution.isBound(piece) && piece.states.length > 1;
     if (isUnsolved) _currentUnsolvedPieces.add(piece);
@@ -215,10 +216,10 @@ class CodeWriter {
 
     if (isUnsolved) _currentUnsolvedPieces.removeLast();
 
-    var childOptions = _pieceOptions.removeLast();
+    var childOptions = _options.removeLast();
 
     // If the child [piece] contains a newline then this one transitively does.
-    if (childOptions.hasNewline) _handleNewline();
+    if (childOptions.hasNewline && _options.isNotEmpty) _handleNewline();
   }
 
   /// Format [piece] if not null.
@@ -243,11 +244,11 @@ class CodeWriter {
   /// If this occurs in a place where newlines are prohibited, then invalidates
   /// the solution.
   void _handleNewline() {
-    if (!_options.allowNewlines) _solution.invalidate();
+    if (!_options.last.allowNewlines) _solution.invalidate(_options.last.piece);
 
     // Note that this piece contains a newline so that we can propagate that
     // up to containing pieces too.
-    _options.hasNewline = true;
+    _options.last.hasNewline = true;
   }
 
   /// Write any pending whitespace.
@@ -325,6 +326,9 @@ enum Whitespace {
 
 /// The mutable state local to a single piece being formatted.
 class _PieceOptions {
+  /// The piece being formatted with these options.
+  final Piece piece;
+
   /// The absolute number of spaces of leading indentation coming from
   /// block-like structure or explicit extra indentation (aligning constructor
   /// initializers, `show` clauses, etc.).
@@ -339,5 +343,5 @@ class _PieceOptions {
   /// Whether any newlines have occurred in this piece or any of its children.
   bool hasNewline = false;
 
-  _PieceOptions(this.indent, this.allowNewlines);
+  _PieceOptions(this.piece, this.indent, this.allowNewlines);
 }
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -17,15 +17,11 @@ import 'solution.dart';
 /// an instance of this class. It has methods that the piece can call to add
 /// output text to the resulting code, recursively format child pieces, insert
 /// whitespace, etc.
-///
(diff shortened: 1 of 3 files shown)


natebosch:
> I broke this PR into two commits to make it easier to review.

Do you still plan to squash merge?

lib/src/back_end/code_writer.dart
```diff
@@ -17,15 +17,11 @@ import 'solution.dart';
...
-///
-/// This class also accumulates the score (the relative desireability of a set
-/// of formatting choices) that the resulting code has by tracking things like
-/// how many characters of code overflow the page width.
 class CodeWriter {
   final int _pageWidth;
 
-  /// The state values for the pieces being written.
-  final PieceStateSet _pieceStates;
+  /// The solution being formatted.
```
natebosch, line 23: [nit] I would expect a "solution" to already have a formatting. If the formatting hasn't been solved yet - what does "solution" refer to?

lib/src/back_end/code_writer.dart
```diff
@@ -101,31 +86,16 @@ class CodeWriter {
   /// solution if the line ends up overflowing.
   final List<Piece> _currentUnsolvedPieces = [];
 
-  /// The options for the current innermost piece being formatted.
-  _PieceOptions get _options => _pieceOptions.last;
+  CodeWriter(this._pageWidth, this._solution);
 
-  /// The offset in the formatted code where the selection starts.
+  /// Called after this writer has finished formatting the entire piece tree.
```
natebosch, line 91: [nit] Prefer to describe a method, at least in the header, by what it does instead of when it is called.

lib/src/back_end/code_writer.dart
```diff
@@ -101,31 +86,16 @@ class CodeWriter {
...
-  /// at which point it gets set. It remains `null` if there is no selection.
-  int? _selectionEnd;
-
-  CodeWriter(this._pageWidth, this._pieceStates);
-
-  /// Returns the finished code produced by formatting the tree of pieces and
-  /// the final score.
-  Solution finish() {
+  /// Returns the final formatted text and the next piece that can be expanded
+  /// from this solution, if any.
```
natebosch, line 94: [nit] What does "this solution" refer to?

lib/src/back_end/solution.dart
```diff
@@ -4,9 +4,11 @@
 import '../piece/piece.dart';
 import 'code_writer.dart';
 
-/// A possibly incomplete set of selected states for a set of pieces being
-/// solved.
-class PieceStateSet {
+/// A single possible line splitting solution.
```
natebosch, line 7: [nit] What is the word "single" referencing here? Since the field is `final` in the code writer, I expect it's mutable and the splits can change over the lifetime of this object - is that correct?

natebosch (commented):
Some initial comments - I don't think I understand the relationship between `Solution` and `CodeWriter`. The fact that they reference each other (since `Solution` instantiates a `CodeWriter` with `this` as an argument) makes this harder to read.

lib/src/back_end/code_writer.dart
```diff
@@ -134,7 +131,10 @@ class CodeWriter {
   ///
   // TODO(tall): Add another API that adds/subtracts existing indentation.
   void setIndent(int indent) {
-    _options.indent = _pieceOptions[_pieceOptions.length - 2].indent + indent;
+    var parentIndent =
+        _options.length > 1 ? _options[_options.length - 2].indent : 0;
+
+    _options.last.indent = parentIndent + indent;
```
kallentu, line 137: Nit: I know you're just renaming this, but this reminded me how hard it was to read these couple lines to figure out how the indentation worked. Could we add a comment, an example or something to make this a little more readable?

lib/src/back_end/code_writer.dart
```diff
@@ -86,9 +86,6 @@ class CodeWriter {
   /// solution if the line ends up overflowing.
   final List<Piece> _currentUnsolvedPieces = [];
 
-  /// The options for the current innermost piece being formatted.
-  _PieceOptions get _options => _pieceOptions.last;
```
kallentu, line 105: Why did we remove this getter? I'd rather have a named variable than using `_options.last` below. Maybe just rename it?

lib/src/back_end/solution.dart
```diff
@@ -130,17 +138,26 @@ class Solution implements Comparable<Solution> {
...
   ///
   /// This should only be called by [CodeWriter].
-  void invalidate() {
+  void invalidate(Piece piece) {
     _isValid = false;
+
+    // If the piece whose newline constraint was violated is already bound to
+    // one state, then every solution derived from this one will also fail in
+    // the same way, so discard the whole solution tree hanging off this one.
+    if (isBound(piece)) _canExpandSolution = false;
```
kallentu: Nice, we love to see it.

kallentu (approved):
Exciting to see! 😄

lib/src/back_end/code_writer.dart
munificent: Reworded and added some docs elsewhere.

I would characterize it like:

* Solution has all of the state required to describe a complete formatting solution. The primary state is the set of States bound to various Pieces. Then the total overflow characters, cost, and output text are derived from that state. They are mutable state, but only mutated by CodeWriter. Once CodeWriter has finished visiting the entire piece tree, Solution becomes effectively immutable.
* CodeWriter is the interface Pieces use to implement their formatting logic. It also stores any emphemeral state used while in the process of generating the formatted text from a given set of piece states but which isn't needed once we know the final text, cost, and overflow.

lib/src/back_end/code_writer.dart
munificent: Just removed the first sentence.

lib/src/back_end/code_writer.dart
munificent: The Solution the CodeWriter is bound to.

One way to think of CodeWriter is that it's the API that Piece is given access to in order to format itself for a given solution.

Reworded.

natebosch: This is cleared up by the other thread.

lib/src/back_end/solution.dart
munificent: It's single in that Solver explores a tree of Solutions. Each Solution has a fixed set of States that determines how it formats the code. (This will change *sort of* in an upcoming optimization.)

The splits don't change over the lifetime of any given Solution. But the Solver produces new Solutions derived from earlier ones and the new ones have different splits than their parents.

It's a little hard to explain in a code review comment, but at a high level, the Solver is doing [combinatorial search](https://en.wikipedia.org/wiki/State_space_search). It's trying to reach a goal state where every Piece has a chosen State. It starts the search from an initial Solution where no Piece is bound to a state. It reaches new Solutions by taking an existing Solution picking an unbound Piece from it, and making a set of new Solutions where each new Solution inherits the bound States of its parent and then also binds the chosen Piece to each of its possible states.

lib/src/back_end/code_writer.dart
munificent: Done. I don't really like how the options stack works either. I thought about reworking it here, but I felt like there was enough change already.

lib/src/back_end/code_writer.dart
munificent: Prior to this change, we initialized `_pieceOptions` with a sort of "default" option with zero indentation. That way, the stack is never empty, and there is always a "parent" even for the outermost piece. It's just that the parent might be sort of fake. That avoided a couple of places where you otherwise need to do "If _pieceOptions isn't empty then use it, otherwise here's the default."

But with this change, `_PieceOptions` has a non-nullable reference to the `Piece` it was created for. That meant I couldn't initialize the stack with a fake outermost value because there's no piece to use for it. So instead, I just check to see if it's empty in the couple of places where it might be.

After that change, it felt weird to me to have an `_options` getter because it could throw if you happened to access at a time when the stack is empty. I thought it was a little clearer to just inline the `.last` calls everywhere so it was more obvious what was going on.

lib/src/back_end/solution.dart
munificent: It took me a _long_ time to figure out why I couldn't just use `isValid` for this. So glad I was able to figure it out.

munificent:
> Do you still plan to squash merge?

I assume so, but what do you think?

natebosch:
> > Do you still plan to squash merge?
> 
> I assume so, but what do you think?

Squash merge SGTM - was just double checking whether I needed to review each commit individually if they were both going to be on the trunk branch.

lib/src/back_end/solution.dart
```diff
@@ -130,43 +82,141 @@ class Solution implements Comparable<Solution> {
...
+    var newCost = cost;
+
+    void bind(Piece thisPiece, State thisState) {
+      // If this piece is already pinned or bound to some other state, then the
+      // solution doesn't make sense.
+      var alreadyBound = thisPiece.pinnedState ?? newStates[thisPiece];
+      if (alreadyBound != null && alreadyBound != thisState) {
+        conflict = true;
+        return;
+      }
```
natebosch, line 206: Would it make sense to check for conflicts in sibling branches and stop early?

```suggestion
      if (conflict) return;
      // If this piece is already pinned or bound to some other state, then the
      // solution doesn't make sense.
      var alreadyBound = thisPiece.pinnedState ?? newStates[thisPiece];
      if (alreadyBound != null && alreadyBound != thisState) {
        conflict = true;
        return;
      }
```

natebosch (approved):

lib/src/back_end/solution.dart
munificent: I don't think it will make a measurable difference, but it doesn't hurt to early out. I'll make the change locally and add a comment.

PR #1373, opened 2024-01-31
Reorganize how Solution handles pinned states and constraints.

munificent:
In the process of turning ListElement into a Piece, I ran into a couple of bugs in how pinned states and constraints between pieces are handled.

The tests didn't catch those bugs because there are basically two bugs that cancel out. The IfPiece was incorrectly pinning an empty else block in an if statement to split. But the solver was incorrectly ignoring pinned states in some cases. This fixes both bugs.

With this change, when generating a Solution, we eagerly bind as many Pieces to States as we can when the Solution is first created before running CodeWriter. To do that, we find all of the pinned Pieces and bind them to their pinned States, then propagate any constraints from that to other child Pieces.

This fixes a couple of bugs, but also, I think, leads to a cleaner module where at Solution creation time, we fill in as many piece states as we can based on what we know statically about the Piece tree (its pinned states and their constraints). Then during solution exploring and line splitting, we don't have to worry about pinned states at all.

This will, I think, align with future optimizations where we do some other piece pinning before running the solver.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -208,7 +208,8 @@ class CodeWriter {
     _options.add(_PieceOptions(piece, _options.lastOrNull?.indent ?? 0,
         _options.lastOrNull?.allowNewlines ?? true));
 
-    var isUnsolved = !_solution.isBound(piece) && piece.states.length > 1;
+    var isUnsolved =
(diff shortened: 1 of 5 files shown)


natebosch (approved):

kallentu (approved):
Nice!

PR #1376, opened 2024-02-01
Solve subtrees separately

munificent:
This is the big optimization. It lets the formatter break out subtrees of the Piece tree format them independently, and compose the results. It also memoizes and reuses subtree solutions across different parent solutions.

I broke it into a few commits to try to make it easier to review.

This doesn't totally solve our performance problems, but it gets us much closer. I've got one more big algorithm-level optimization to clean up and send out and then after that it's (hopefully!) just constant-factor micro-optimizations.


lib/src/back_end/solution_cache.dart
```diff
@@ -0,0 +1,46 @@
...
+/// Maintains a cache of [Piece] subtrees that have been previously solved.
+///
+/// If a given [Piece] has newlines before and after it, then (in most cases,
+/// assuming there are no other constraints), then the way it is formatted
+/// really only depends on its leading indentation and state. In that case, we
+/// can format that piece using a separate Solver and insert the results in any
+/// Solution that has that piece at that leading indentation.
+///
+/// This cache stores those previously formatted subtree pieces so that
+/// [CodeWriter] can reuse them across Solution[s].
```
kallentu, line 18: Nit: The brackets for `Solution[s]` are probably not supposed to be like that.

lib/src/front_end/piece_writer.dart
```diff
@@ -242,7 +243,8 @@ class PieceWriter {
   /// Finishes writing and returns a [SourceCode] containing the final output
   /// and updated selection, if any.
   SourceCode finish(Piece rootPiece) {
-    var formatter = Solver(_formatter.pageWidth);
+    var cache = SolutionCache();
```
kallentu, line 246: Do we need to initialize SolutionCache outside of the Solver? Can the Solver initialize and just own it's own cache when created?

kallentu (approved):
Thanks for splitting up the commits, it was nice to review.

lib/src/back_end/solution_cache.dart
```diff
@@ -0,0 +1,46 @@
...
+class SolutionCache {
+  final _cache = <_Key, Solution>{};
+
+  /// Looks up a previously cached solution for formatting [root] with leading
```
natebosch, line 22: [nit] I think the fact that this method does work to solve when the solution isn't cached should be highlighted in the header.

lib/src/back_end/solution_cache.dart
```diff
@@ -0,0 +1,46 @@
...
+  ///
+  /// If found, returns the cached solution. Otherwise solves it, caches it,
+  /// and returns the result.
+  Solution find(int pageWidth, Piece root, State state, int indent) {
+    // See if we've already formatted this piece at this indentation.
+    var key = (root, state, indent: indent);
+    if (_cache[key] case var solution?) return solution;
+
+    var solver = Solver(this, pageWidth: pageWidth, leadingIndent: indent);
+    return _cache[key] = solver.format(root, state);
```
natebosch, line 33: [optional] Consider using `Map.putIfAbsent` instead of computing the hash code for this key twice.

I think some of the backends also do better with `map[key] ??=` than a lambda, so we could consider using that.

```dart
    return _cache[key] ??=
        Solver(this, pageWidth: pageWidth, leadingIndent: indent)
            .format(root, state);
```

lib/src/front_end/piece_writer.dart
munificent: Yeah, doing so lets us reuse cached subtree Solutions across different Solvers. Imagine we have a tree like:

```
Sequence
  Infix
    List
      Element
        ...
  ...
```

1. We create a Solver for the Sequence piece and start trying Solutions.
2. When we get to Infix, it decides to format its operand (List) using a separate Solver.
3. That inner Solver starts trying solutions for List.
4. When it gets to List, that piece decides to format Element using a separate Solver.
5. We find the optimal Solution for that Element subtree and stuff the result in the cache (1).
6. We pop back up to the Solver for Infix. It eventually finds a Solution which gets cached too.
7. Then we pop back to the outermost Solver for the Sequence. It keeps iterating through solutions and eventually decides to format the Infix another way.
8. We go back into Infix. Maybe Infix decides to create another Solver for its operand (List) or maybe it doesn't. Either way, it eventually gets back to Element.
9. Now, when we look up a previously solved format for Element, we find it in the cache, even though the Solver that created it (from step 3) is long gone.

It seems a little convoluted, but I did some testing and I do see that the cache is reusing Solutions created by different Solvers, so it's adding value.

lib/src/piece/list.dart
```diff
@@ -56,11 +56,50 @@ class ListPiece extends Piece {
   final ListStyle _style;
 
   ListPiece(this._before, this._elements, this._blanksAfter, this._after,
-      this._style);
+      this._style) {
+    // For most elements, we know whether or not it will have a comma based
+    // only the comma style and its position in the list, so pin those here.
```
natebosch, line 61: [nit] Missing word "on" between "only" and "the comma style"?

lib/src/back_end/solution_cache.dart
munificent: Heh, oops. :)

lib/src/back_end/solution_cache.dart
```diff
@@ -0,0 +1,46 @@
...
+  /// [indent].
+  ///
+  /// If found, returns the cached solution. Otherwise solves it, caches it,
+  /// and returns the result.
+  Solution find(int pageWidth, Piece root, State state, int indent) {
+    // See if we've already formatted this piece at this indentation.
+    var key = (root, state, indent: indent);
+    if (_cache[key] case var solution?) return solution;
+
+    var solver = Solver(this, pageWidth: pageWidth, leadingIndent: indent);
```
natebosch, line 32: [optional] These classes could be decoupled if the cache took a computation callback. Constructing an object with an argument of `this` always makes me nervous - and I think this one is relatively easy to avoid.

I also think it would be fine to inline this class to `Solver` - make the `_cache` field a `Map` and write the small `_cache[key] ??= Solver(_cache, ...).format`

This would also let us move the cache argument to a private constructor and let the root `Solver` create it's own cache to address Kallen's comment.

natebosch (approved):

lib/src/back_end/code_writer.dart
```diff
@@ -204,30 +217,60 @@ class CodeWriter {
...
+  /// optimal results together.
+  ///
+  /// It's only safe to pass [separate] when the piece's formatting depends
+  /// only on its starting indentation and state. If the piece's formatting can
+  /// be affected by the contents of the current line, the contents after the
+  /// piece's ending line, or constraints between pieces, then [separate] should
+  /// be `false`. It's up to the parent piece to only call this when it's safe
+  /// to do so. In practice, this usually means when the parent piece knows that
+  /// [piece] will have a newline before and after it.
+  void format(Piece piece, {bool separate = false}) {
```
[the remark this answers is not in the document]
munificent: I had that initially, actually. But I felt like the two halves were *just* short enough that it felt silly to split them out so I merged them. If you feel it's better split, that tips me over the fence too. Done.

lib/src/back_end/solution_cache.dart
munificent: Reworded.

lib/src/back_end/solution_cache.dart
munificent: See my reply to Kallen's comment. We specifically don't want the Solver to own its own cache because we want to share the cache across all Solvers.

PR #1378, opened 2024-02-05
Pin Pieces if we can determine they will split just by their children.

munificent:
If, for example, the elements inside a ListPiece are longer than the page width when you add up all of their characters, then the ListPiece will definitely split. But the Solver will try every possible way to *not* split that Piece and waste a ton of time doing so.

In particular, since the Solver reaches outermost Pieces first, it will fully explore the entire Solution tree where those don't split before finally giving up and letting them split. But the outmost Pieces also have the largest number of Solutions that can be derived from them, so the Solver gets stuck exploring entire huge pointless branches.

That's unnecessary because in some cases it's obvious that a Piece will split just because of how big it's contents are. Assuming an 80-character width, a ListPiece for a collection literal is *definitely* going to split if it's contents are >80 characters long. We don't need the Solver to figure that out.

This PR adds an optimization where before solving, we ask each Piece if it can be pinned to a state based just on the length of its children and whether there are any newlines in them.

Then, for ListPiece and InfixPiece, we implement that optimization.

Also, added a bunch of benchmarks that validate this optimization for various syntax using those pieces.

With this optimization, the new formatter is now only about 30% slower than the old one on the large benchmark. (It's faster than the old formatter and slower for some of the other various benchmarks.) That's at the point where we can start looking for constant-time optimizations.


lib/src/back_end/solution.dart
```diff
@@ -283,15 +283,19 @@ class Solution implements Comparable<Solution> {
...
-    var states = _pieceStates.keys
-        .map((piece) => '$piece:${_pieceStates[piece]}')
-        .join(' ');
+    var states = <String>[];
+    _pieceStates.forEach((piece, state) {
+      if (piece.additionalStates.isEmpty) return;
+      if (piece.pinnedState != null) return;
+
+      states.add('$piece${_pieceStates[piece]}');
+    });
```
natebosch, line 292: [optional]
```suggestion
    var states = [
      for (var MapEntry(key: piece, value: state) in _pieceStates.entries)
        if (piece.additionalStates.isNotEmpty && piece.pinnedState == null)
          '$piece$state'
    ];
```

lib/src/back_end/solution.dart
```diff
@@ -283,15 +283,19 @@ class Solution implements Comparable<Solution> {
...
   String toString() {
-    var states = _pieceStates.keys
-        .map((piece) => '$piece:${_pieceStates[piece]}')
-        .join(' ');
+    var states = <String>[];
+    _pieceStates.forEach((piece, state) {
+      if (piece.additionalStates.isEmpty) return;
+      if (piece.pinnedState != null) return;
+
+      states.add('$piece${_pieceStates[piece]}');
```
natebosch, line 291:
```suggestion
      states.add('$piece$state');
```

natebosch (approved):

lib/src/back_end/solution.dart
munificent: Oops, I accidentally committed your second suggestion first which collided with this one. I'll commit this locally and push it. Nice suggestion.

lib/src/piece/piece.dart
```diff
@@ -63,6 +97,30 @@ abstract class Piece {
   /// Invokes [callback] on each piece contained in this piece.
   void forEachChild(void Function(Piece piece) callback);
 
+  /// If the piece can determine that it will always end up in a certain state
+  /// given [pageWidth] and size metrics retured by calling [containsNewline]
```
kallentu, line 101: Nit: Typo on "returned"

lib/src/piece/list.dart
```diff
@@ -182,6 +182,39 @@ class ListPiece extends Piece {
...
+      if (before.containsNewline) return State.split;
+      totalLength += before.totalCharacters;
+    }
+
+    for (var element in _elements) {
+      // Elements that can be block arguments won't necessarily force the list
+      // to split.
+      if (element.allowNewlines) continue;
+
+      if (element.containsNewline) return State.split;
```
kallentu, line 201: Is it possible for the element to contain a newline if it doesn't allow newlines? Maybe I'm reading this wrong.

kallentu (approved):

lib/src/piece/list.dart
munificent: Yeah, for example:

```dart
function(arg, // comment
);
```

Here, the `arg, // comment` will have `allowNewlines` false because it isn't a kind of expression that supports block formatting (collection literal, etc.)

But it will also have `containsNewline` true because of the `// comment` hanging of the end of the element.

So this is a case where we know the list will always split because there's a newline in an argument that doesn't allow a newline when the list is unsplit.

Maybe it's confusing because of the name. The `allowNewlines` field on `ListElementPiece` means "allow newlines *when the list is unsplit*". Renamed to `allowNewlinesWhenUnsplit` to make that clearer. :)

PR #1379, opened 2024-02-09
Prefer any valid solution over an invalid one, regardless of overflow.

munificent:
While the Solver is working through Solutions, it keeps track of the best Solution it's found so far. This will get returned if there is no Solution that fits within the page width: when overflow is inevitable, the formatter still tries to do the best it can.

The Solver is careful to not consider an invalid solution (one that contains a newline where none is allowed) to be the best solution. Except in one case: If the *initial* solution is invalid, the Solver will still store it in best. Then, as long as no other solution has less overflow than the initial invalid one, it ends up returning an invalid solution.

This fixes that: any valid solution, regardless of overflow, will kick out that initial invalid solution if there is one.

### DIFF
--- a/lib/src/back_end/solver.dart
+++ b/lib/src/back_end/solver.dart
@@ -92,15 +92,18 @@ class Solver {
         debug.log('');
       }
 
-      // Since we process the solutions from lowest cost up, as soon as we find
-      // a valid one that fits, it's the best.
       if (solution.isValid) {
+        // Since we process the solutions from lowest cost up, as soon as we
+        // find a valid one that fits, it's the best.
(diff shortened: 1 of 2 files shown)


natebosch (approved):
We could also consider making solutions comparable and take validity into account in the comparison.

munificent:
> We could also consider making solutions comparable and take validity into account in the comparison.

Solutions do implement Comparable already, actually. But that deliberately ignores validity. It's used to determine the order that Solutions are explored and we want explore invalid solutions before valid ones if they have lower cost because an invalid solution may lead to a winning valid one.

kallentu (approved):

PR #1382, opened 2024-02-13
Fill in a bunch of corners of if-case and patterns.

munificent:
- Format if-case elements in collection literals.
- Format `when` clauses in if-case.
- Format wildcard patterns (which are basically the same as variables).
- Format constant patterns with a leading `const`.
- Migrated over a bunch of other patterns tests that can be run now.

To handle if-case and when clauses, I ended up creating a new IfCasePiece since it because different enough from AssignPiece to not make sense to share code any more.

This doesn't add support for guard clauses in switch cases yet. I'll do that separately.


natebosch (approved):

kallentu (approved):
Very appreciated, thank you.

PR #1383, opened 2024-02-14
Format guard clauses in switches.

munificent:
There were enough tweaks in the formatting specific to the two contexts where the guards can appear that I ultimately decided to make new Piece classes for switch expression cases and switch statement cases. I was hoping for more code reuse, but I think it would have been harder to maintain if I jammed all of this into a single Piece class.

I also slightly tweaked the indentation style of switch expression cases. When I'd first added support for switch expressions in the new formatter, I changed the style from what the old formatter had, for reasons that aren't entirely clear. Maybe I thought it made it more consistent to indent +4?

Either way, once guard clauses came into play, it became clear that the old style made more sense because it gives a clearer indentation level for the `when` clause when it splits. So this change also makes switch expression formatting more similar to the old style.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -12,6 +12,7 @@ import '../dart_formatter.dart';
 import '../piece/adjacent.dart';
 import '../piece/adjacent_strings.dart';
 import '../piece/assign.dart';
+import '../piece/case.dart';
 import '../piece/constructor.dart';
 import '../piece/for.dart';
 import '../piece/if.dart';
@@ -1005,8 +1006,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
 
   @override
   Piece visitIndexExpression(IndexExpression node) {
-    Piece? targetPiece;
-    if (node.target case var target?) targetPiece = nodePiece(target);
+    var targetPiece = optionalNodePiece(node.target);
     return createIndexExpression(targetPiece, node);
   }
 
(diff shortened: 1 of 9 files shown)


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1005,8 +1006,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
 
   @override
   Piece visitIndexExpression(IndexExpression node) {
-    Piece? targetPiece;
-    if (node.target case var target?) targetPiece = nodePiece(target);
+    var targetPiece = optionalNodePiece(node.target);
```
munificent, line 1009: Since I added `optionalNodePiece()` as a utility method in the PR, I went ahead and found every place that I could use it and swapped out the call.

test/expression/switch.stmt
```diff
@@ -28,7 +28,7 @@ e = switch (c) {
 e = switch (c) {
   first => a,
   second =>
-      veryLongExpression + thatSplits,
+    veryLongExpression + thatSplits,
```
munificent, line 31: This is the style change. By indenting bodies +2 instead of +4, it makes it more natural to indent the `when` clause +4 when it splits. If we indent the body +4, then you have to do like +8 for the `when` clause which looks pretty extreme.

lib/src/front_end/ast_node_visitor.dart
natebosch: Could it be an extension method? `node.target?.nodePiece` reads clearly and the nullability is obvious that way.

kallentu (approved):
Generally, LGTM!

lib/src/front_end/ast_node_visitor.dart
natebosch: Oh, it's an instance method... nevermind.

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: Yeah, there's some state in PieceWriter and CommentWriter that all of the front end AST -> Piece code needs to have implicit access to.

PR #1384, opened 2024-02-14
Migrate constant pattern tests to tall formatter.

kallentu:
Tried to do some check of what patterns we have left using the tests, and realized these aren't migrated (I think).
The list/map/record tests aren't the actual `ListPattern`, `MapPattern` etc, so labelled this more as an overall `constant.stmt`.


munificent (approved):
This is great, thanks!

PR #1385, opened 2024-02-15
Format object patterns.

kallentu:
- Added `ObjectPattern`s to `canBlockSplit`.
- Migrated object pattern tests and object patterns with comments tests.

There's one test case in there that's not great, but I think it's a general problem with `ListPattern`, `MapPattern` too. 
Maybe revisit it with the other patterns in another CL.

```
>>> Split in type argument.
if (obj case LongClassName<First, Second>()) {;}
<<<
### TODO(tall): It formats like this if there's no elements. Similarly with lists, maps etc.
if (obj
    case LongClassName<
        First,
        Second
        >()) {
  ;
}
```

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -393,6 +393,7 @@ extension PatternExtensions on DartPattern {
           elements.canSplit(rightBracket),
         MapPattern(:var elements, :var rightBracket) =>
           elements.canSplit(rightBracket),
+        ObjectPattern(:var fields, :var rightParenthesis) ||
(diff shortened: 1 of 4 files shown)


natebosch (approved):

munificent (approved):

munificent:
> ### TODO(tall): It formats like this if there's no elements. Similarly with lists, maps etc.
> if (obj
>     case LongClassName<
>         First,
>         Second
>         >()) {
>   ;
> }

Yeah, that looks like an ugly corner case of the collapsible indentation stuff I did. Leaving a TODO sounds good. I'll try to take a look at some point.

PR #1386, opened 2024-02-15
Format metadata annotations.

munificent:
This handles metadata everywhere except on for loop variables. I'll do that later after patterns in for loops are supported.

Since there were some metadata tests on pattern variable declarations, this also includes a trivial implementation of that. There's a TODO to revisit it and add real tests, but it's enough for the metadata tests.

Figuring out where to put the metadata-handling code was a little tricky since so many places in the language can have metadata. I ultimately decided to put it into AdjacentBuilder. Every place that was working with metadata also had one of those builders in play, so that made it easy to add a simple call to `b.metadata()` and get the metadata working.

It does mean that AdjacentBuilder isn't strictly about adjacent pieces anymore, but I think that's OK. In practice, it's used mostly as a general "build pieces out of smaller things" builder.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -117,7 +117,14 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
 
   @override
   Piece visitAnnotation(Annotation node) {
-    throw UnimplementedError();
+    return buildPiece((b) {
+      b.token(node.atSign);
+      b.visit(node.name);
+      b.visit(node.typeArguments);
+      b.token(node.period);
+      b.visit(node.constructorName);
+      b.visit(node.arguments);
+    });
   }
 
   @override
@@ -360,6 +367,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
   @override
   Piece visitConstructorDeclaration(ConstructorDeclaration node) {
     var header = buildPiece((b) {
+      b.metadata(node.metadata);
       b.modifier(node.externalKeyword);
       b.modifier(node.constKeyword);
       b.modifier(node.factoryKeyword);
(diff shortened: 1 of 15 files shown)


lib/src/piece/list.dart
```diff
@@ -487,3 +487,10 @@ class ListStyle {
       this.splitListIfBeforeSplits = false,
       this.allowBlockElement = false});
 }
+
+enum E {
```
natebosch, line 491: Remove?

natebosch (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -512,48 +521,53 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
     if (node.members.isEmpty) {
       // If there are no members, format the constants like a delimited list.
       // This keeps the enum declaration on one line if it fits.
+      // TODO(tall): The old style preserves blank lines and newlines between
+      // enum values. A newline will also force the enum to split even if it
+      // would otherwise fit. Do we want to do that with the new style too?
```
kallentu, line 526: Hmm, I think I like having the ability to split my enum values?, even if it would fit on one line. It seems more clear to me.

kallentu (approved):

lib/src/piece/list.dart
munificent: Heh, oops, yes. :)

lib/src/front_end/ast_node_visitor.dart
munificent: > Hmm, I think I like having the ability to split my enum values?

Me too. I think I didn't add support for that to the new formatter because there wasn't a clear mechanism in place to implement it (currently, we really only force newlines and blank lines around comments), but I think we probably want it here. That's what the TODO is for. :)

PR #1387, opened 2024-02-15
Format extension types.

munificent:
Nothing particularly interesting here, but it did flush out a little bug in `PieceFactory.createParameter()` where a comment before the metadata annotation would get moved to after the parameter.

### DIFF
--- a/lib/src/piece/list.dart
+++ b/lib/src/piece/list.dart
@@ -487,3 +487,10 @@ class ListStyle {
       this.splitListIfBeforeSplits = false,
       this.allowBlockElement = false});
 }
+
+enum E {
+  a,
+
+  b,
+  c,
+}
--- a/lib/src/piece/list.dart
+++ b/lib/src/piece/list.dart
@@ -487,10 +487,3 @@ class ListStyle {
       this.splitListIfBeforeSplits = false,
       this.allowBlockElement = false});
 }
-
-enum E {
-  a,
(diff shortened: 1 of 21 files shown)


natebosch (approved):

lib/src/piece/list.dart
```diff
@@ -487,3 +487,10 @@ class ListStyle {
       this.splitListIfBeforeSplits = false,
       this.allowBlockElement = false});
 }
+
```
kallentu, line 490: Heh, what's happening with this enum? Probably remove this too ahah

test/declaration/extension_type_comment.unit
```diff
@@ -0,0 +1,61 @@
+40 columns                              |
+>>> Inline block comments everywhere.
```
kallentu, line 2: Wow lol these tests are way more extensive than our typical comment tests.

kallentu (approved):

lib/src/piece/list.dart
munificent: Heh, oops! I was seeing how the current formatter behaved with some temp code. Removed.

test/declaration/extension_type_comment.unit
munificent: Yeah, Lasse wrote these and I just migrated them over.

PR #1388, opened 2024-02-15
Format cast pattern.

kallentu:
Quick pattern :)
Added some extra tests for comments and splitting, just in case. Seems simple enough.


natebosch (approved):

munificent (approved):
Looks great!

PR #1392, opened 2024-02-19
Format pattern variable declarations, including in for loops.

munificent:
This required pretty significantly revamping AssignPiece since now there can be complex syntax on the left side of the `=` too. It went through several iterations but I think I managed to simplify it down to something reasonable that still passes all of the tests.

This PR doesn't do pattern assignment, but that should hopefully be a trivial change after this.

### DIFF
--- a/lib/src/piece/assign.dart
+++ b/lib/src/piece/assign.dart
@@ -19,37 +19,75 @@ import 'piece.dart';
 ///
 ///     var x = 123;
 ///
-/// If the value is a delimited "block-like" expression, then we allow splitting
-/// inside the value but not at the `=` with no additional indentation:
+/// This state also allows splitting the right side if it can be block
+/// formatted:
 ///
 ///     var list = [
 ///       element,
 ///     ];
 ///
-/// [_atOperator] Split after the `=`:
+/// [_blockSplit] Block split in the operands that support it and expression
+/// split in the others. This state requires at least one operand to support
+/// block splitting. Examples:
 ///
-///     var name =
-///         longValueExpression;
+///     var [
+///       element,
+///     ] = value;
+///
+///     var [
+///       element,
+///     ] = longOperand +
+///         anotherOperand;
+///
+///     var (longVariable &&
+///         anotherVariable) = [
+///       element,
+///     ];
+///
+/// [_atOperator] Split at the `=` or `in` operator and allow expression
+/// splitting in either operand:
+///
+///     var (longVariable &&
+///             anotherVariable) =
+///         longOperand +
+///             anotherOperand;
 class AssignPiece extends Piece {
-  /// Split after the operator.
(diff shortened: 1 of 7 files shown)


lib/src/piece/assign.dart
```diff
@@ -19,37 +19,75 @@ import 'piece.dart';
...
+/// [_atOperator] Split at the `=` or `in` operator and allow expression
+/// splitting in either operand:
+///
+///     var (longVariable &&
+///             anotherVariable) =
+///         longOperand +
+///             anotherOperand;
 class AssignPiece extends Piece {
-  /// Split after the operator.
+  /// Split at the operator.
```
kallentu, line 55: I liked the previous "after the operator". I think it's clearer to me where it actually splits.

lib/src/piece/assign.dart
```diff
@@ -19,37 +19,75 @@ import 'piece.dart';
...
+///       element,
+///     ] = longOperand +
+///         anotherOperand;
+///
+///     var (longVariable &&
+///         anotherVariable) = [
+///       element,
+///     ];
+///
+/// [_atOperator] Split at the `=` or `in` operator and allow expression
```
kallentu, line 47: I don't know the exact ruling on this, but would it be better to have these doc comments at the site of the member? (below) I find myself scrolling up to look at an example.

kallentu (approved):
AssignPiece is getting pretty bulky. I'm interested in seeing how these block splits interact in larger/nested examples.

lib/src/piece/assign.dart
munificent: It doesn't always split after the operator, though. :) When an AssignPiece splits at the `in` keyword in a for-in loop, it splits before the `in`, not after.

lib/src/piece/assign.dart
munificent: > I don't know the exact ruling on this

Yeah, I've gone back and forth on this too. I could never decide if it makes more sense to document the different styles all in one place on the class, at the declaration of each `State` field, or at the top of the `format()` method.

I've tried to consistently document them on the Piece class itself. That's mainly because most pieces use `State.unsplit` and often `State.split` for some style, and there's no other obvious place to document how the piece styles itself when in those states.

munificent:
> AssignPiece is getting pretty bulky.

Yeah, I'm wondering if it's really three pieces in a trenchcoat and should be split into separate classes.

On the other hand, it may be complex just because it *is* used for so many different parts of the language: variable declarations, assignments, for-in loops, map entries, named arguments, etc.

> I'm interested in seeing how these block splits interact in larger/nested examples.

Me too. Pretty soon, we should have the whole language covered (only *7* UnimplementedErrors left!). Once that's down to zero, I'll start migrating the regression tests and that should kick up a lot of bugs.

PR #1393, opened 2024-02-20
Drop node bits

kevmoo:
No longer used/supported


munificent:
I think removing this is probably fine (since it's basically non-functional anyway). But adding @jacob314 since he initially wrote this and knows what the motivation was.

munificent (approved):

kevmoo:
Easy to revert if @jacob314 has...thoughts...

jacob314:
Thanks for deleting this!

PR #1395, opened 2024-02-20
Format null asserts, null checks, and relational patterns.

kallentu:
Migrated what tests were left of null-assert, null-check, and relational patterns.
Straightforward implementations for these (I think?).


munificent:
Sorry for the collision!

PR #1401, opened 2024-02-21
Add more tests for type arguments in object patterns and remove a TODO.

munificent:
I'm honestly not sure what the TODO was even trying to say. But I think the formatter behaves like we want here so there's nothing TODO as far as I can tell. But this corner did seem a little under tested, so added some more.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -153,15 +153,19 @@ class CodeWriter {
     var parentIndent = _indentStack.last.indent;
     var parentCollapse = _indentStack.last.collapsible;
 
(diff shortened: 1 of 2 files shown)


kallentu (commented):
Haha, that TODO is on me.
I wasn't sure if that's what we wanted.
The old test expected output was which I thought was much better than what's on line 172.
```
if (obj
    case LongClassName<
      First,
      Second
    >()) {
  ;
}
```

munificent:
Oh, you're right! The output is wrong and is missing some indentation.

This must be an issue with collapsible indentation thing I did. I'll look into that.

PR #1406, opened 2024-02-24
For loop metadata

munificent:
When I added support for metadata, I skipped metadata on for loop variables because I thought it would be a little annoying and wanted to do it separately. Boy was I right.

Jamming metadata in there highlighted a handful of other formatting bugs in for loops, which in turn made some general bugs in how block-like constructs to the left of assignments (which includes `=`, `=>`, and `in`) are handled. There were a lot of tests, but there are just so many combinations of kinds of syntax on the left of an assignment-like thing and expressions on the right that's hard to cover everything.

So I revamped that first. Then I added support for for loop metadata.

To make it easier to review, I tried to break it into bite-sized meaningful commits.

### DIFF
--- a/test/statement/for_in.stmt
+++ b/test/statement/for_in.stmt
@@ -61,8 +61,9 @@ for (var identifier
 >>> Split inside initializer.
 for (var identifier in someVeryLong + iterableExpression) { body; }
 <<<
-for (var identifier in someVeryLong +
-    iterableExpression) {
+for (var identifier
+    in someVeryLong +
+        iterableExpression) {
   body;
 }
 >>> Prefer block-like splitting after `in`.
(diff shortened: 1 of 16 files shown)


test/statement/for_in.stmt
```diff
@@ -61,8 +61,9 @@ for (var identifier
 >>> Split inside initializer.
 for (var identifier in someVeryLong + iterableExpression) { body; }
 <<<
-for (var identifier in someVeryLong +
-    iterableExpression) {
+for (var identifier
```
kallentu, line 72: Nice. I like this a lot better.

kallentu (approved):
There's a lot of moving pieces here. Thanks for partitioning them into smaller commits!

AssignPiece is getting quite big and I think it's almost at the max capacity for how much I can understand it haha.

munificent:
> AssignPiece is getting quite big and I think it's almost at the max capacity for how much I can understand it haha.

Same. Hard same. I'll do some thinking about how we might be able to split it up into simpler pieces.

natebosch (approved):

PR #1408, opened 2024-02-26
Tweak the heuristics on which expressions allow block formatting.

munificent:
I was trying out the new formatting on Flutter and noticed a few issues:

**1. Immediately-invoked functions in asserts are common.**

Prior to this change, these were not block format candidates, giving you:

```dart
assert(
  () {
    // Some slow computation...
  }(),
  'Message.',
);
```

With this change, they're treated like (uninvoked) function expressions:

```dart
assert(() {
  // Some slow computation...
}(), 'Message.');
```

**2. Formatting large deep widget trees looks weird.**

Prior to this change, a function call inside an argument list can be treated like a block argument even if there are other arguments. That tends to pack things in strangely in widget code, like:

```dart
SizedBox(height: 1000.0, width: double.infinity, child: Column(
  children: <Widget>[Scrollbar(key: key1, child: SizedBox(
    height: 300.0,
    width: double.infinity,
    child: SingleChildScrollView(key: innerKey, child: const SizedBox(key: Key(
      'Inner scrollable',
    ), height: 1000.0, width: double.infinity)),
  ))],
));
```

With this change, a function or method call inside an argument list is only block formatted if there are no other arguments. In other words, you can block format a function call if the outer call is a pure wrapper around the call, but not otherwise. That leads to:

```dart
SizedBox(
  height: 1000.0,
  width: double.infinity,
  child: Column(children: <Widget>[Scrollbar(
    key: key1,
    child: SizedBox(
      height: 300.0,
      width: double.infinity,
      child: SingleChildScrollView(
        key: innerKey,
        child: const SizedBox(
          key: Key('Inner scrollable'),
          height: 1000.0,
          width: double.infinity,
        ),
      ),
    ),
  )]),
);
```

I think that looks a lot nicer.

**3. Deep, complex argument lists are combinatorially slow.**

We have an optimization to eagerly force an argument list to split if its contents clearly won't fit. But that optimization ignores the size of a potential block argument since the contents of that can be wrapped across multiple lines without forcing the outer argument list to split.

A consequence of that is that big deeply nested function calls end up slow if it happens that each one only has a single block-formattable call. That turns out to be very common in Flutter code where you have a deep chain of nested widgets along with a bunch of other arguments.

The above change to disallow block formatting function calls if there are any other arguments conveniently fixes that performance issue in most cases. (Though I expect we will want to do more work here to handle pathological cases.)

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -4,6 +4,8 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 
+import 'piece/list.dart';
+
 extension AstNodeExtensions on AstNode {
   /// The first token at the beginning of this AST node, not including any
   /// tokens for leading doc comments.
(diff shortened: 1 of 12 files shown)


lib/src/ast_extensions.dart
```diff
@@ -149,58 +151,68 @@ extension ExpressionExtensions on Expression {
...
+  bool get canBlockSplit => blockFormatType != BlockFormat.none;
+
+  /// When this expression is in an argument list, what kind of block formatting
+  /// category it belongs to.
+  BlockFormat get blockFormatType {
     // Unwrap named expressions to get the real expression inside.
-    var expression = this;
-    if (expression is NamedExpression) {
-      expression = expression.expression;
+    if (this case NamedExpression named) {
```
natebosch, line 160: [nit] Can we put this in the `switch` cases?

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -51,7 +51,16 @@ class DelimitedListBuilder {
 
   /// Creates the final [ListPiece] out of the added brackets, delimiters,
   /// elements, and style.
-  ListPiece build() {
+  Piece build() {
+    // To simplify the piece tree, if there are no elements, just return the
+    // brackets concatenated together.
```
natebosch, line 56: Double checking - is there any issue with empty lists containing comments?

lib/src/piece/list.dart
```diff
@@ -55,8 +55,9 @@ class ListPiece extends Piece {
   /// The details of how this particular list should be formatted.
   final ListStyle _style;
 
-  ListPiece(this._before, this._elements, this._blanksAfter, this._after,
-      this._style) {
+  ListPiece(
+      this._before, this._elements, this._blanksAfter, this._after, this._style)
+      : assert(_elements.isNotEmpty) {
```
natebosch, line 60: [nit] This change probably deserves a doc change somewhere in this class. Consider adding a doc on the ctor, or (maybe better?) adding a note to the class level doc.

natebosch (approved):

lib/src/ast_extensions.dart
munificent: Good catch, yes.

lib/src/front_end/delimited_list_builder.dart
munificent: No, we don't. DelimitedListBuilder hoists comments up into separate elements, so `_elements` wouldn't be empty if there were any. Added a sentence here to that effect.

lib/src/piece/list.dart
munificent: Added a word on the class doc and a sentence on the constructor.

PR #1410, opened 2024-02-28
Fix a crash and some formatting bugs in cascades.

munificent:
I did a test run of the new formatter on the Flutter repo and it hit some crashes in cascades where a cascade section contains a method chain like `foo..a.b`.

While investigating that, I realized that the existing code for formatting cascades tries to reuse the regular method chain formatting code entirely unnecessarily.

This fixes the crash, and the formatting, and is simpler for handling cascades.

### DIFF
--- a/lib/src/front_end/chain_builder.dart
+++ b/lib/src/front_end/chain_builder.dart
@@ -72,8 +72,19 @@ class ChainBuilder {
     if (_root case CascadeExpression cascade) {
       _visitTarget(cascade.target);
 
+      // When [_root] is a cascade, the chain is the series of cascade sections.
       for (var section in cascade.cascadeSections) {
-        _unwrapCall(section);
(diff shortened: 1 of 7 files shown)


lib/src/front_end/chain_builder.dart
```diff
@@ -176,10 +187,8 @@ class ChainBuilder {
   ///     .baz[0][1]
   ///     .bang()
   void _unwrapCall(Expression expression) {
-    var isCascade = _root is CascadeExpression;
```
kallentu, line 179: Remove the `CascadeSection` comment above?

kallentu (approved):

lib/src/front_end/chain_builder.dart
munificent: Good catch, thanks!

natebosch (approved):

PR #1418, opened 2024-03-05
Force list, set, and maps to split if they are nested.

munificent:
If a list, set, or map literal contains (directly or indirectly) another list, set, or map literal, then force the outer one to split, even if it would otherwise fit.

We don't do this for record expressions, or any kind of pattern because those tend to be smaller and look better when kept closely packed.

The behavior here is exactly the same as in the current formatter.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -136,7 +136,7 @@ mixin PieceFactory {
   /// [splitOnNestedCollection] is also `true`, even if the collection would
   /// otherwise not need to split. This is `true` for list, map, and set
   /// expressions because they are often used for composite data structures and
-  /// they're easier to read if they don't get packed to densely:
+  /// they're easier to read if they don't get packed too densely:
   ///
   ///     // Prefer:
   ///     data = {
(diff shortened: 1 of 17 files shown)


natebosch (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -118,13 +130,43 @@ mixin PieceFactory {
   }
 
   /// Creates a [ListPiece] for a collection literal or pattern.
+  ///
+  /// If [splitOnNestedCollection] is `true`, then this collection is forced to
+  /// split if it contains any non-empty collections where
+  /// [splitOnNestedCollection] is also `true`, even if the collection would
+  /// otherwise not need to split. This is `true` for list, map, and set
+  /// expressions because they are often used for composite data structures and
+  /// they're easier to read if they don't get packed to densely:
```
kallentu, line 139: Typo on "too".

kallentu (approved):

lib/src/front_end/piece_factory.dart
munificent: Nice catch, thanks. :)

PR #1419, opened 2024-03-05
Test that benchmarks produce their expected output.

munificent:
The benchmark expectations weren't being run as part of the tests which meant that it was easy to break the benchmarks without realizing it. A recent fix to not block format collection literal elements broke a couple of the benchmarks.

This fixes them. It also ensures they won't break again by having the format tests also test all of the benchmarks.

And it adds support to the benchmark runner to run all the benchmarks in one go.

### DIFF
--- a/test/utils.dart
+++ b/test/utils.dart
@@ -6,6 +6,7 @@ import 'dart:io';
 
 import 'package:dart_style/dart_style.dart';
 import 'package:dart_style/src/constants.dart';
+import 'package:dart_style/src/testing/benchmark.dart';
 import 'package:dart_style/src/testing/test_file.dart';
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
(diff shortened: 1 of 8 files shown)


test/utils.dart
```diff
@@ -141,6 +142,41 @@ Future<void> testFile(String path,
...
+        // The test files always put a newline at the end of the expectation.
+        // Statements from the formatter (correctly) don't have that, so add
+        // one to line up with the expected result.
+        var actualText = actual.text;
+        if (!benchmark.isCompilationUnit) actualText += '\n';
+
+        var expected =
+            useTallStyle ? benchmark.tallOutput : benchmark.shortOutput;
+
+        // Fail with an explicit message because it's easier to read than
```
natebosch, line 169: The matcher output gives more information about where the difference is doesn't it?

natebosch (approved):

kallentu (approved):

test/utils.dart
munificent: It does show the location of the difference, which is nice. But it also quotes/escapes the string which makes it a little harder to read.

More importantly, it makes my test iteration workflow harder. One of the things I tend to do when writing tests is just put random text in the expected output, let the test fail, then copy the actual output back to the test after verifying it looks like I expect. Doing the custom printing in the test here makes that easier. For example, here's the output you get if you do `expect(actual, expected)`:

```
  Expected: 'variableName = [\n'
              '  element,\n'
              '  element, oops\n'
              '  element,\n'
              '];\n'
              ''
    Actual: 'variableName = [\n'
              '  element,\n'
              '  element,\n'
              '  element,\n'
              '];\n'
              ''
     Which: is different.
            Expected: ...   element, oops\n  e ...
              Actual: ...   element,\n  elemen ...
                                    ^
             Differ at offset 40

  package:matcher        expect
  test/utils.dart 203:9  _testFile.<fn>.<fn>
```

And here's the output you get with the custom `fail()` call I use:

```
  Formatting did not match expectation. Expected:
  variableName = [
    element,
    element, oops
    element,
  ];

  Actual:
  variableName = [
    element,
    element,
    element,
  ];
```

I lose highlighting the location of the difference, but overall I find it a lot easier to read and copy from.

natebosch: Side tangent - in `package:checks` this is the current output;

```
  Expected: a String that:
    equals '              variableName = [
        element,
        element,
        element,
      ];'
  Actual: '              variableName = [
      element,
      element, oops
      element,
    ];'
  Which: differs at offset 58:
  ...   element,\n    elem ...
  ...   element, oops\n    ...
                ^
```

We avoid the extra `'\n` noise, but do have single quotes and somewhat awkward initial indentation. I wonder if we should consider treating multiline strings specially for pretty printing.

What do you think of that output? Would you find it acceptable here, or still use a custom approach?

PR #1420, opened 2024-03-07
Preserve newlines (or lack of) in collections with line comments.

munificent:
Large collection literals often have some internal structure that the author wants to highlight by having multiple elements on one line, sort of like a matrix.

The old style has a sort of hacky but in practice useful heuristic to enable that: if a collection literal contains a line comment, then the newlines (or lack of) between pairs of elements are preserved:

https://github.com/dart-lang/dart_style/wiki/FAQ#why-does-the-formatter-mess-up-my-collection-literals

This implements that same heuristic in the new style. The behavior is slightly different in the new style. In the old style, an argument list can be split in a variety of ways:

```dart
// 1. All on one line:
f(argument1, argument2, argument3, argument4)

// 2. Split into two lines at after any particular argument:
f(argument1, argument2, argument3,
    argument4)
f(argument1, argument2,
    argument3, argument4)
f(argument1,
    argument2, argument3, argument4)

// 3. Split every argument into its own line:
f(
    argument1,
    argument2,
    argument3,
    argument4)
```

It uses that same logic for the elements on a single line in a collection formatted with this rule. That kicks in if the authored code has all of the elements on one line but they don't actually fit. In that case, the formatter will still split them to try to fit. With the old style, it may use any of the above approaches to try to fit.

In the new style, argument lists don't support splitting between an arbitrary pair of elements like (2) above. So you just get:

```dart
// 1. All on one line:
f(argument1, argument2, argument3, argument4)

// 3. Split every argument into its own line:
f(
  argument1,
  argument2,
  argument3,
  argument4,
)
```

Likewise for collection literals containing a line comment. If all of the elements on a single line don't fit, it will split all of them onto their own lines instead of trying to split just between one pair of them.

Longer-term, I intend to support opting regions of code out of formatting completely. I expect users will use that for large matrix-like collections where they want to control both line splitting and aligning things into columns. So the specific behavior of this hack won't matter as much.

But it *is* a widely relied on rule today, so this part preserves it.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -125,6 +125,25 @@ extension AstIterableExtensions on Iterable<AstNode> {
   /// but allow a split if there are elements or comments inside.
   bool canSplit(Token rightBracket) =>
       isNotEmpty || rightBracket.precedingComments != null;
+
+  /// Returns `true` if the collection containing these elements and terminated
+  /// by [rightBracket] contains any line comments before, between, or after
+  /// any elements.
+  ///
+  /// Comments within an element are ignored.
+  bool containsLineComments([Token? rightBracket]) {
+    // Look before each element.
+    for (var element in this) {
+      if (element.beginToken.hasLineCommentBefore) return true;
+    }
+
+    // Look before the closing bracket.
+    if (rightBracket != null) {
+      if (rightBracket.hasLineCommentBefore) return true;
+    }
+
+    return false;
+  }
 }
 
 extension ExpressionExtensions on Expression {
@@ -408,4 +427,17 @@ extension PatternExtensions on DartPattern {
           fields.canSplit(rightParenthesis),
         _ => false,
       };
+}
+
+extension TokenExtensions on Token {
+  /// Whether this token has a preceding comment that is a line comment.
+  bool get hasLineCommentBefore {
+    for (Token? comment = precedingComments;
+        comment != null;
+        comment = comment.next) {
+      if (comment.type == TokenType.SINGLE_LINE_COMMENT) return true;
+    }
+
+    return false;
+  }
 }
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -131,19 +131,9 @@ extension AstIterableExtensions on Iterable<AstNode> {
   /// any elements.
   ///
   /// Comments within an element are ignored.
-  bool containsLineComments([Token? rightBracket]) {
-    // Look before each element.
-    for (var element in this) {
-      if (element.beginToken.hasLineCommentBefore) return true;
-    }
-
-    // Look before the closing bracket.
-    if (rightBracket != null) {
-      if (rightBracket.hasLineCommentBefore) return true;
-    }
-
-    return false;
-  }
+  bool containsLineComments([Token? rightBracket]) =>
+      any((element) => element.beginToken.hasLineCommentBefore) ||
+      (rightBracket?.hasLineCommentBefore ?? false);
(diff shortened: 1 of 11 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -828,17 +830,96 @@ mixin PieceFactory {
...
   /// Creates a [ListPiece] for the given bracket-delimited set of elements.
-  Piece createList(Iterable<AstNode> elements,
+  ///
+  /// If [preserveNewlines] is `true`, then any newlines or lack of newlines
+  /// between pairs of elements in the input are preserved in the output. This
+  /// is used for collection literals that contain line comments to preserve
+  /// the author's deliberate structuring, as in:
+  ///
+  ///     matrix = [
+  ///       1, 2, 3,
```
natebosch, line 840: [nit] Show this "contain line comments" detail in the example.

```suggestion
  /// is used for collection literals that contain line comments to preserve
  /// the author's deliberate structuring, as in:
  ///
  ///     matrix = [
  ///       1, 2, 3, //
```

lib/src/ast_extensions.dart
```diff
@@ -125,6 +125,25 @@ extension AstIterableExtensions on Iterable<AstNode> {
...
   bool canSplit(Token rightBracket) =>
       isNotEmpty || rightBracket.precedingComments != null;
+
+  /// Returns `true` if the collection containing these elements and terminated
+  /// by [rightBracket] contains any line comments before, between, or after
+  /// any elements.
+  ///
+  /// Comments within an element are ignored.
+  bool containsLineComments([Token? rightBracket]) {
+    // Look before each element.
```
natebosch, line 135: [nit] Line comments are leaning towards redundant against the code here - consider if any should be removed.

lib/src/ast_extensions.dart
```diff
@@ -125,6 +125,25 @@ extension AstIterableExtensions on Iterable<AstNode> {
...
+      if (element.beginToken.hasLineCommentBefore) return true;
+    }
+
+    // Look before the closing bracket.
+    if (rightBracket != null) {
+      if (rightBracket.hasLineCommentBefore) return true;
+    }
+
+    return false;
+  }
```
natebosch, line 146: [optional] I think the code-golfed version is pretty readable because it looks declarative.

```suggestion
  bool containsLineComments([Token? rightBracket]) =>
      any((element) => element.beginToken.hasLineCommentBefore) ||
      (rightBracket?.hasLineCommentBefore ?? false);
```

natebosch (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -828,17 +831,96 @@ mixin PieceFactory {
...
+          comments.hasNewlineBetween(
+              elements[i - 1].endToken, element.beginToken)) {
+        // This element begins a new line. Add the elements on the previous
+        // line to the list builder and start a new line.
+        builder.add(lineBuilder.build());
+        lineBuilder = DelimitedListBuilder(this, lineStyle);
+        atLineStart = true;
+      } else if (!atLineStart) {
+        // This element follows another on the same line.
+        // lineBuilder.space();
```
kallentu, line 908: Do we need this else if and the comment?

kallentu (approved):

lib/src/front_end/piece_factory.dart
munificent: Oops, you're right. That was a remnant of a different approach where I was using AdjacentPieceBuilder for the line and not DelimitedListBuilder. Removed. Thanks!

PR #1422, opened 2024-03-07
Tweak the block formatting rules to disallow `=>` function expressions.

munificent:
When I first added support for block formatting arguments in the new style, I allowed function expressions with `=>` bodies if they had parameters with the idea that the parameter list itself would get the block-like formatting, as in:

```dart
function((
  parameter1,
  parameter2,
) => body);
```

I'm honestly not sure why I thought that was a good idea. The old formatter doesn't allow that. I don't think it looks particularly good. And, worst of all, the formatter doesn't actually *require* the parameter list to be the place where the split occurs, so the current implementation allows unwanted output like:

```dart
function((parameter) => operand +
    anotherOperand));
```

This is definitely not intended.

I've started migrating regressions tests and in the examples I've seen where this rule comes into play, I think it looks worse to allow block formatting here. So this PR changes it to not allow block formatting any `=>` body function expression.

### DIFF
--- a/test/invocation/block_argument_kind.stmt
+++ b/test/invocation/block_argument_kind.stmt
@@ -25,46 +25,18 @@ function((parameter) { body; });
 function((parameter) {
   body;
 });
->>> Empty block-bodied function expression with parameters.
+>>> Empty block-bodied function expression with parameters is not a block argument.
 function((parameter, anotherParameter) {});
 <<<
-function((
-  parameter,
-  anotherParameter,
-) {});
->>> Expression-bodied function expression with parameters.
+function(
+  (parameter, anotherParameter) {},
+);
+>>> Expression-bodied function expression with parameters is not a block argument.
 function((parameter, anotherParameter) => body);
 <<<
-function((
-  parameter,
-  anotherParameter,
-) => body);
->>> Block-bodied function expression with block comment in the parameters.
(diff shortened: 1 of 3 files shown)


natebosch (approved):

test/invocation/block_argument_kind.stmt
```diff
@@ -25,46 +25,18 @@ function((parameter) { body; });
...
-) => body);
->>> Expression-bodied function expression with line comment in the parameters.
-function((// comment
-) => body);
-<<<
-function((
-  // comment
-) => body);
+function(
+  (parameter, anotherParameter) => body,
```
kallentu, line 38: Can you add a testing example (if it's not already here, I didn't check too hard!) with a parameter list that's > 40 characters? Would we expect it to block split then?

kallentu (approved):

test/invocation/block_argument_kind.stmt
munificent: Ah, testing that's a good idea, thanks!

It doesn't block split then either. You get:

```dart
function(
  (
    parameter,
    anotherParameter,
    thirdParameter,
  ) => body,
);
```

I think I prefer that over block splitting it (and in practice, it's rare for a lambda to have so many parameters that they split).

PR #1431, opened 2024-03-23
Don't indent split binary operands in an assignment-like context.

munificent:
Binary operators indent their subsequent operands in most places:

```dart
function(
  operand1 +
      operand2 +
      operand3,
);
```

But it's gratuitous on contexts where a split between the operands will also lead to a split before the operand. In that case, it looks nicer to align the operands:

```dart
variable =
    operand1 +
    operand2 +
    operand3;
```

The places where this can occur are "assignment-like": the right-hand side of `=`, `:`, or `=>`.

This is another example where the relationship between a parent and child AST node affects how they are formatted. We don't have a great systematic way of modeling those relationships in the codebase right now. Sometimes, we have the code building the parent Piece peek at the child's AST node type and do something different with the child based on that. Sometimes, the child calls `.parent` on itself and peeks at the parent type. We don't generally look at the type of the Piece of the child because we don't consider that to be reliably encoding what kind of original syntax it came from (though we could).

This takes yet another approach where we allow a parent AST node to pass in a "context" when visiting one of its children. A child can then look at that context to tweak its formatting. I'm not sure if I love the approach, but I thought I'd give it a try and see what you think.


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1159,11 +1177,24 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
...
+    // format the operands in parallel, like:
+    //
+    //     var {
+    //       key:
+    //         operand1 &&
+    //         operand2,
+    //     } = value;
+    //
+    // Also, if it's the outermost pattern in a switch expression case, we
+    // flatten the operands like parallel cases:
```
natebosch, line 1190: Why don't we do the same with `&&`?

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -42,6 +42,10 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
   @override
   final CommentWriter comments;
 
+  /// The context set by the surrounding AstNode when visiting a child, or
+  /// [NodeContext.none] if the parent node doesn't set a context.
+  NodeContext _parentContext = NodeContext.none;
```
natebosch, line 47: This has a general name and description, but the current usage is pretty specific. Should we be worried that it might not generalize in practice?

lib/src/front_end/ast_node_visitor.dart
munificent: Good question. We definitely *could.* I added special formatting for `||` because switch expressions don't allow multiple cases to share a body, so you end up using `||` instead and I wanted the resulting formatted code to look like parallel cases. That same logic doesn't really apply to `&&`.

Do you think it would be more readable to format `&&` the same way? For what it's worth, the current formatting style has this same special rule for `||` but not `&&`.

lib/src/front_end/ast_node_visitor.dart
munificent: I'm somewhat worried.

There are an increasing number of places where we need to look at both parent and child to decide how to format something, and there are a number of different strategies we take to do that. I would really like something more consistent and unified, but I haven't figure out what yet. I have some aspirations that this could be used for more things, but haven't taken the time to refactor other code to using this yet.

Or maybe there's some other better solution out there that I haven't found yet. One idea in the back of my head is that maybe Pieces should encode a little more about the original syntax nodes they are based on? Right now, if you want to choose how to format a parent based on its children, you need to make that choice *before* the children are converted to Pieces, because Pieces tend to drop too much information on the floor. (Dropping stuff on the floor is sort of the *point* because it lets different kinds of syntax share the same formatting and means we need far fewer Piece classes than AST node classes. But it makes situations like this harder.)

I'm not sure.

lib/src/front_end/ast_node_visitor.dart
natebosch: I think I might be surprised/confused if I saw a chain of `&&` next to a chain of `||` and they were formatted differently. I haven't noticed that the current formatting differs between them. That's probably because `&&` is rare in that position - and I don't think I've _ever_ seen a chain with more than one `&&` in a pattern in real code.

So maybe it doesn't matter then. I was surprised by the difference as I read it here, but I likely won't ever be surprised by the difference in real code.

lib/src/front_end/ast_node_visitor.dart
natebosch: It feels brittle to me - you can't know if `_parentContext` is an _immediate_ parent, or if there is some chain of ancestry that didn't care to set that field.

If being able to preserve context for a non-immediate ancestor is a feature - then we need to worry about any ways where nodes using this strategy could interleave in the ancestry chain with the nodes they are trying to communicate context to.

munificent: It will always be the context of your immediate parent. More concretely, it will be whatever context that the parent node passes in when it calls `nodePiece()` (which every way of visiting a node ultimately bottoms out on) or it will be `NodeContext.none`. If you don't pass in a context when calling `nodePiece()`, it defaults to `NodeContext.none`. It doesn't inherit the surrounding one.

lib/src/front_end/ast_node_visitor.dart
munificent: > I likely won't ever be surprised by the difference in real code.

I think that's true. I'll leave it as it is for now, and if users request the same formatting for `&&`, we can consider it then.

natebosch (approved):
I think this approach should be OK for now - we can revisit if it turns out not to work for the next use case that comes up.

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1940,8 +1973,12 @@ class AstNodeVisitor extends ThrowingAstVisitor<Piece> with PieceFactory {
   /// If [commaAfter] is `true`, looks for a comma token after [node] and
   /// writes it to the piece as well.
   @override
-  Piece nodePiece(AstNode node, {bool commaAfter = false}) {
+  Piece nodePiece(AstNode node,
+      {bool commaAfter = false, NodeContext context = NodeContext.none}) {
+    var previousContext = _parentContext;
+    _parentContext = context;
     var result = node.accept(this)!;
+    _parentContext = previousContext;
```
kallentu, line 1981: Will this replace `.parent` calls? Like in `visitConditionalExpression`? It would be nice to have one consistent way of finding out the parent context.

I also was wondering if we wanted each piece to have a parent context (or a parent) they could look at. It seems like parent context should exist on each Piece rather than in the visitor.

kallentu (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: > Will this replace `.parent` calls? Like in `visitConditionalExpression`? It would be nice to have one consistent way of finding out the parent context.

That's my aspiration, yes. I'll write a long TODO comment that effect.

> I also was wondering if we wanted each piece to have a parent context (or a parent) they could look at. It seems like parent context should exist on each Piece rather than in the visitor.

Maybe, I but I try to avoid storing state in Pieces unless we actually need it to persist all the way to line splitting. So far, I think we only really need the parent context during visiting and piece construction and it can be discarded afterwards, which is why I made it an ephemeral bit of state in the visitor.

PR #1433, opened 2024-03-28
Allow comments before infix expressions to not split the operator.

munificent:
Currently, unless captured by some SequenceBuilder or DelimitedListBuilder, comments (that aren't hanging of the previous token) get attached to the token immediately following them. For example:

```dart
value =
    // comment
    a + b;
```

Here, since there is a newline before `// comment`, it's not attached to the `=` and is instead attached as a leading comment to the `a` token owned by the identifier expression as the first operand to `+`.

That means that as far as the `+` expression is concerned, there is a newline in one of its operands, so it thinks it needs to split, leading to:

```dart
value =
    // comment
    a +
        b;
```

But since the comment is, to a user, outside of the infix expression, that split looks pointless.

This fixes that. When leading comments appear right before any infix-like expression (binary operator, conditional operator, binary pattern, etc.), we hoist them out of the innermost operand that would otherwise claim them and make them precede the infix piece itself.


kallentu (approved):

kallentu (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -853,14 +858,20 @@ mixin PieceFactory {
...
+
       // Otherwise, just write the node itself.
       builder.visit(e);
     }
 
     traverse(node);
     operands.add(builder.build());
 
-    return InfixPiece(operands, indent: indent);
+    return InfixPiece(leadingComments, operands, indent: indent);
```
natebosch, line 874: How do we know that `leadingComments` has been definitely been assigned by this point?

munificent: It's a little unclear since the code is recursive, but it's traversing a tree of binary expressions in order to flatten a chain of the same binary operator into a single list of operands (since we split them all together or not).

As it traverses, it will definitely hit leaf expressions at some point for all operands. And there will definitely be a first leaf expressions. So `leadingComments` will always get initialized. It's just hard for the compiler to see it because of the recursion.

But actually, now that I think about it, I don't need to wait until we've recursed into the left-most leaf operand to initialize `leadingComments`. The left-most operand's left-most token will always be the same as the left-most token of the entire `node`, so I can just initialize `leadingComments` from that. Fixed.

natebosch (approved):

PR #1437, opened 2024-04-01
Remove some unneeded code and update a doc comment.

munificent:


lib/src/front_end/chain_builder.dart
```diff
@@ -230,11 +230,6 @@ class ChainBuilder {
         _visitTarget(expression);
 
       // Selectors.
-      case AssignmentExpression():
-        var piece = _visitor.createAssignment(expression.leftHandSide,
-            expression.operator, expression.rightHandSide);
-        _calls.add(ChainCall(piece, CallType.property));
-
```
munificent, line 237: I think this was used back when cascades went through this code path, but they don't any longer, so it's just dead code.

lib/src/piece/piece.dart
```diff
@@ -174,8 +174,8 @@ sealed class TextPiece extends Piece {
 
   /// Append [text] to the end of this piece.
   ///
-  /// If [text] internally contains a newline, then [containsNewline] should
-  /// be `true`.
+  /// If [text] may contain any newline characters, then [multiline] must be
+  /// `true`.
```
munificent, line 178: Looks like I didn't update the doc comment when I reworked comment handling.

kallentu (approved):

PR #1438, opened 2024-04-03
Remove pushAllowNewlines() and popAllowNewlines() from CodeWriter.

munificent:
One of the main ways that formatting rules are expressed is by a parent piece controlling whether a child piece may contain newlines when the parent is in a certain state. The API for this was stateful: Pieces could push or pop onto a stack to say whether they allow or disallow newlines. When child pieces are formatted, we use the top of the stack to determine whether the solution should be invalidated if the child ended up containing a newline.

I *thought* there was a compelling need for this API to be stateful because if a parent piece doesn't push anything onto the stack, then it will inherit the constraint of its parent. In that way, newlines would percolate up until reaching a piece that cared. It turns out that isn't necessary: Each parent piece will either directly invalidate based on its immediate children or not.

So I got rid of the stack and the stateful API completely. When calling CodeWriter.format() to format a child piece, the parent passes in right then whether or not it allows a newline in the child. At least according to the current tests, that seems to be sufficient.

I think this simplifies reasoning about the formatting code in each piece, because you don't have to mentally track the allow newline stack.

Also, I'm working on some larger changes to be able to express "a newline is allowed here but only from block splitting, not expression splitting", which we need to prevent some annoying incorrect format. I believe this will get the codebase in a better state to be able to express that.

### DIFF
--- a/lib/src/piece/case.dart
+++ b/lib/src/piece/case.dart
@@ -11,11 +11,8 @@ class CaseExpressionPiece extends Piece {
   /// Split after the `=>` before the body.
   static const State _beforeBody = State(1);
 
-  /// Split before the `when` guard clause.
-  static const State _beforeWhen = State(2);
-
   /// Split before the `when` guard clause and after the `=>`.
-  static const State _beforeWhenAndBody = State(3);
+  static const State _beforeWhenAndBody = State(2);
 
   /// The pattern the value is matched against along with the leading `case`.
   final Piece _pattern;
@@ -58,7 +55,7 @@ class CaseExpressionPiece extends Piece {
   @override
   List<State> get additionalStates => [
         _beforeBody,
-        if (_guard != null) ...[_beforeWhen, _beforeWhenAndBody],
+        if (_guard != null) ...[_beforeWhenAndBody],
       ];
 
   @override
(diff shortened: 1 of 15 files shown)


lib/src/piece/case.dart
```diff
@@ -11,11 +11,8 @@ class CaseExpressionPiece extends Piece {
   /// Split after the `=>` before the body.
   static const State _beforeBody = State(1);
 
-  /// Split before the `when` guard clause.
-  static const State _beforeWhen = State(2);
```
munificent, line 15: It turns out this state was never actually usable...

lib/src/piece/case.dart
```diff
@@ -89,22 +81,18 @@ class CaseExpressionPiece extends Piece {
...
-    writer.pushAllowNewlines(allowNewlineInPattern);
-    writer.format(_pattern);
-    writer.popAllowNewlines();
+    writer.format(_pattern, allowNewlines: allowNewlineInPattern);
 
     if (indentPatternForGuard) writer.popIndent();
 
     if (_guard case var guard?) {
       writer.pushIndent(Indent.expression);
-      writer.pushAllowNewlines(allowNewlineInGuard);
```
munificent, line 104: ...on this line `allowNewlineInGuard` will be `false` when `state` is `_beforeWhen`. So this solution will be invalidated if a newline is written...

lib/src/piece/case.dart
```diff
@@ -89,22 +81,18 @@ class CaseExpressionPiece extends Piece {
...
-      writer.splitIf(state == _beforeWhen || state == _beforeWhenAndBody);
```
munificent, line 105: ...and then the piece writes a newline *itself* when `state` is `_beforeWhen`, invalidating its own state and preventing `_beforeWhen` from ever being a valid solution. Oops.

natebosch (approved):

kallentu (approved):

PR #1444, opened 2024-04-24
Don't allow function calls to be block arguments.

munificent:
I've been running the new formatter on the Flutter repo looking for ugly output and performance problems. (Mostly the latter.) One source of pathologically bad performance is nested function call block arguments. It's not a problem with block arguments in general because when a collection is a block argument, the collection itself doesn't support block arguments. But consider:

```
f1(f2(f3(f4(f5(f6(f7(f8(f9(f10(arg))))))))))
```

There are 2^10 ways that can be formatted, and no current optimizations kick in to avoid that rathole.

The slowest file to format in the Flutter repo is popup_menu_test.dart and the main culprit is pathological performance from this. I added a new benchmark for the slowest statement in that file.

With the old formatter, it finishes in 0.088ms. With the new formatter before this change, it's about 211ms, so over 2,000 times slower. This PR gets it down to 4.827ms, about 43 times faster.

My machine can format the entire Flutter repo packages directory in 35.17s. This change gets it down to 31.872. Not bad for essentially a one line change.

I also generally think code looks better when we don't allow function calls to be block arguments. In simple cases, it can look nice, but it gets weird quickly. I scanned a diff of this PR on the Flutter repo and in almost every case, I think it leads to easier to read code.

### DIFF
--- a/lib/src/front_end/delimited_list_builder.dart
+++ b/lib/src/front_end/delimited_list_builder.dart
@@ -421,7 +421,6 @@ class DelimitedListBuilder {
     // TODO(tall): These heuristics will probably need some iteration.
     var functions = <int>[];
     var collections = <int>[];
-    var invocations = <int>[];
     var adjacentStrings = <int>[];
 
     for (var i = 0; i < _elements.length; i++) {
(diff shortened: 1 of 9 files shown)


lib/src/front_end/delimited_list_builder.dart
```diff
@@ -431,7 +430,30 @@ class DelimitedListBuilder {
         case BlockFormat.collection:
           collections.add(i);
         case BlockFormat.invocation:
```
natebosch, line 432: [optional] Consider reordering the cases by behavior - cases where the element is collected vs ignored.

natebosch (approved):

lib/src/front_end/delimited_list_builder.dart
munificent: I thought about that, but I (for perhaps not justifiable reasons) tend to keep cases in the same order that the enum defines the values. That way every switch on that same type has the same order. I'm not sure if that's actually better or not, but it made sense to me at the time.

PR #1448, opened 2024-04-26
Don't allow splitting in interpolation expressions.

munificent:
The old formatter doesn't allow splitting inside interpolation because it almost never looks good. It's better to just let the line overflow, especially if it's a multi-line string. If the user wants it to fit, they're better off splitting the string themselves instead of letting the formatter split inside the interpolation.

Also, a single string can have an unbounded number of arbitrarily large interpolation expressions. Since the expressions can't be separately formatted (because they appear in the middle of a line), this leads to combinatorial performance. It's fairly rare, but it's quite bad when it happens. When I format the Flutter repo, one of the slowest files is packages/flutter/lib/src/material/toggle_buttons.dart. It's spending almost all of its time in:

```
  @override
  String toString() {
    return '''
    {
      selected:
        hovered: $hoverColor, otherwise: ${colorScheme?.primary.withOpacity(0.04)},
        focused: $focusColor, otherwise: ${colorScheme?.primary.withOpacity(0.12)},
        pressed: $splashColor, otherwise: ${colorScheme?.primary.withOpacity(0.16)},
      unselected:
        hovered: $hoverColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.04)},
        focused: $focusColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.12)},
        pressed: $splashColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.16)},
      otherwise: null,
    }
    ''';
  }
```

I added a benchmark containing three of the worst examples I found in the Flutter repo. Before this change:

```
interpolation                  1225.785 1241.865 1242.435  1236.695    (none)
```

With this PR, the benchmark is over 10,000x faster:

```
interpolation                     0.115    0.122    0.303    0.132    (none)
```

Since large string interpolations are rare in practice, this doesn't have a huge impact on overall format times of a large corpus. (On my machine, it takes formatting the Flutter repo from 25.43s to 25.01s.) But it does address a significant hot spot and I also think the resulting output looks better.

### DIFF
--- a/benchmark/case/interpolation.expect
+++ b/benchmark/case/interpolation.expect
@@ -1,0 +1,22 @@
+main() {
+  var a = '''
+      {
+        selected:
+          hovered: $hoverColor, otherwise: ${colorScheme?.primary.withOpacity(0.04)},
+          focused: $focusColor, otherwise: ${colorScheme?.primary.withOpacity(0.12)},
+          pressed: $splashColor, otherwise: ${colorScheme?.primary.withOpacity(0.16)},
+        unselected:
+          hovered: $hoverColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.04)},
+          focused: $focusColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.12)},
+          pressed: $splashColor, otherwise: ${colorScheme?.onSurface.withOpacity(0.16)},
+        otherwise: null,
+      }
+      ''';
+
+  log.writeln(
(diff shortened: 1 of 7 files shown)


kallentu (approved):
I agree with your comment of the pinning _each_ child of the interpolation expression is maybe not ideal? The overall improvement looks good though.

natebosch (approved):

munificent:
> I agree with your comment of the pinning _each_ child of the interpolation expression is maybe not ideal?

Yeah, I don't love it. But if we just pin the outermost node, it often won't accomplish anything. To fully prevent splitting, we really do have to walk the whole tree (or have an entirely different mechanism).

PR #1458, opened 2024-05-03
Migrate regression/0000/

munificent:
Migrate the `regression/0000` directory to the new tall style.

Some of the resulting output isn't exactly what I'd hoped for. For now, I've just added TODOs so that we can land this and pin down the behavior but I hope I can iterate on the style rules some and get the output better.

Aside from those cases, though, overall I like the output.

I split this into two commits for ease of reviewing. The first commit just copies the tests over as-is. The second commit applies the new style to them. The diff there shows the difference between the old style and the new style. If you have feedback on the style itself, now is a good time to comment on it.


test/tall/regression/0000/0037.stmt
```diff
@@ -30,7 +30,7 @@ remote
...
     .then((sendPort) {
-  sendReceive(sendPort, 'foo').then((msg) {
-    print('remote isolate handler received msg: $msg');
-  });
-});
\ No newline at end of file
+      sendReceive(sendPort, 'foo').then((msg) {
+        print('remote isolate handler received msg: $msg');
+      });
+    });
```
munificent, line 36: Here's an example of a style the old formatter simply couldn't do but which I think looks much better.

test/tall/regression/0000/0083.unit
```diff
@@ -12,10 +12,14 @@ main() {
...
-        .then((_) {
+      toASecondLine,
+      wrapWap,
+      onThe,
+      argsHere,
+    ).then((_) {
       someLongFunctionHereToo(
-          'value value value value value value valvalue value value value');
+        'value value value value value value valvalue value value value',
+      );
```
munificent, line 22: Here's another one where the old style was weird and busted but the new style looks good.

kallentu (approved):

natebosch (approved):

PR #1459, opened 2024-05-03
Migrate regression/0100/

munificent:
Here's the next batch.


test/tall/regression/0100/0108.unit
```diff
@@ -134,49 +142,112 @@ class ResolutionCopier {
...
-                                            _isEqualTokens(node.abstractKeyword, toNode.abstractKeyword)),
-                                        _isEqualTokens(node.classKeyword, toNode.classKeyword)),
-                                    _isEqualNodes(node.name, toNode.name)),
-                                _isEqualNodes(node.typeParameters, toNode.typeParameters)),
-                            _isEqualNodes(node.extendsClause, toNode.extendsClause)),
-                        _isEqualNodes(node.withClause, toNode.withClause)),
-                    _isEqualNodes(node.implementsClause, toNode.implementsClause)),
-                _isEqualTokens(node.leftBracket, toNode.leftBracket)),
-            _isEqualNodeLists(node.members, toNode.members)),
-        _isEqualTokens(node.rightBracket, toNode.rightBracket));
```
munificent: The formatting gets weird here because the old formatter goes pathological and ends up having to give up because it can't find a solution in time. The new formatter actually gets to a solution.

kallentu (approved):

natebosch (approved):

PR #1468, opened 2024-05-06
Migration regression/0500/

munificent:
Here's the next batch.


test/tall/regression/0500/0584.unit
```diff
@@ -2,5 +2,7 @@
 bool
     listEquals<T>(List<T> a, List<T> b) => const ListEquality<T>().equals(a, b);
 <<<
-bool listEquals<T>(List<T> a, List<T> b) =>
-    const ListEquality<T>().equals(a, b);
\ No newline at end of file
+bool listEquals<T>(List<T> a, List<T> b) => const ListEquality<T>().equals(
```
kallentu, line 5: Kinda wish this split at the `=>` instead. Especially for the small `a, b` parameters.

kallentu (approved):

test/tall/regression/0500/0584.unit
munificent: Yeah, me too. This is a really tricky corner that I've seen come up in a few places. There are many examples where (to me at least) it's obviously better to split the argument list:

```dart
widget = SomeWidget(
  someParam: value,
  anotherParam: anotherValue,
);
```

But there do sometimes seem to be places like this example where that looks weird and it would be better to split at the `=`/`:`/`=>`. But I've yet to intuit any sort of rule to determine when to split the arguments and when not to.

Maybe it has something to do with the size of the argument list compared to the size of the stuff before the `(`? I'm not sure.

test/tall/regression/0500/0506.unit
```diff
@@ -65,14 +65,15 @@ class Foo {
...
-          null); // async modifier
+        makeIdentifier(param.name),
+        makeParameters(param.parameters),
+        null, // body
+        param.type == null ? null : makeType(param.type),
+        makeEmptyModifiers(), // TODO: Function parameter modifiers?
+        null, // initializers
+        null, // get/set
+        null,
+      ); // async modifier
```
natebosch, line 76: I think it's probably fine to expect authors to manually move some edge case comments like this.


Should we move it here too, or leave it as is and let the regression test remain non-representative for code using the new format?

test/tall/regression/0500/0584.unit
natebosch: +1

How difficult would it be to have it prefer to split at the `=>` when the expression fits on a single line at that indentation level? Would that heuristic be enough, or would it lead to weird editing experiences?

natebosch (approved):

test/tall/regression/0500/0506.unit
munificent: My rule for regression tests is to always leave them as-is. (That's why, for example, many of them still contain `new` even though it's no longer idiomatic to use it.)

PR #1472, opened 2024-05-06
Migrate regression/0700/

munificent:
Here's the next batch.


test/tall/regression/0700/0704.unit
```diff
@@ -22,15 +22,17 @@ void main() {
...
       var source = new File(sourcePath).readAsStringSync();
       var target = new File(targetPath).readAsStringSync();
 
       var hash = sha1.convert(utf8.encode(source));
       if (!target.contains("Checksum: $hash")) {
-        fail("$targetPath is out-of-date.\n"
-            "Run pub run grinder to update it.");
+        fail(
+          "$targetPath is out-of-date.\n"
+          "Run pub run grinder to update it.",
```
natebosch, line 34: Do we only indent string literal continuations in collections? Or is that not yet implemented for the tall style?

test/tall/regression/0700/0707.stmt
```diff
@@ -1,4 +1,4 @@
 >>> (indent 4) (fix optional-new)
     sink.write('FILE ACCESSED ${new DateTime.now()}\n');
 <<<
-    sink.write('FILE ACCESSED ${DateTime.now()}\n');
\ No newline at end of file
+    sink.write('FILE ACCESSED ${new DateTime.now()}\n');
```
natebosch, line 4: Did this test run with different arguments before?
Or have we not implemented fixes in the tall style?

natebosch (approved):

test/tall/regression/0700/0704.unit
munificent: It uses the same logic as the old formatter where it doesn't indent them so that the strings are all nicely lined up when doing so is unlikely to be confusing but will indent them otherwise. The logic is [here](https://github.com/dart-lang/dart_style/blob/main/lib/src/ast_extensions.dart#L371).

In this case, since there's only one argument that's a string, it doesn't indent the subsequent lines. (I also think the new style helps here because now the argument list splits before the first line too so that the strings actually line up.)

test/tall/regression/0700/0707.stmt
munificent: Oh, good catch! I didn't notice this when I was looking at the diff.

The new style doesn't implement fixes and won't implement them. We'll be removing `dart format --fix` when the new style ships (since `dart fix` can now do all the same fixes and plenty more). It looks like there are only two regression tests (both coincidentally in the 0700s) that use fixes, so I'll just remove them from the tall tests since they aren't relevant.

kallentu (approved):

Issue #1466, opened 2024-05-06
Allow an unsplit method target on the same line as an assignment


munificent:
This issue is somewhat similar to #1465. Where that one relates to conditional expressions on the right-hand side of "assignments" (`=`, `:`, `=>`), this is about method chains. A method chain can split in a few ways:

```dart
// Unsplit:
target.method(arg).another(arg);

// Block split:
target.method(arg).another(
  arg,
);

// Chain split:
target
    .method(arg)
    .another(arg);
```

Currently, if the right-hand side of an assignment is a method chain that chain splits, it also splits the assignment, like:

```dart
thing = 
    target
        .method(arg)
        .another(arg);
```

It would be nice to allow the target to stay on the same line as the assignment:

```dart
thing = target
    .method(arg)
    .another(arg);
```

This would only apply if the target itself doesn't split. A split in the target would still force the assignment to split. So:

```dart
// Disallowed:
thing = (veryLong +
        targetExpression)
    .method(arg)
    .another(arg);

// Allowed:
thing =
    (veryLong +
            targetExpression)
        .method(arg)
        .another(arg);
```

munificent:
Fixed by https://github.com/dart-lang/dart_style/pull/1687.

PR #1477, opened 2024-05-06
Eagerly bind AssignPiece if it can't block split and operands don't fit.

munificent:
If neither of the two operands to an AssignPiece can be block formatted, and they contain newlines or the whole thing doesn't fit the page width, then the AssignPiece will definitely split at the operator.

This optimization helps a few rare cases (like deeply nested curried-style functions) from going pathological. I noticed it when migrating one of the regression tests over.

It doesn't have a noticeable affect on formatting the whole Flutter repo, but it doesn't hurt either.


benchmark/case/curry.expect
```diff
@@ -0,0 +1,56 @@
...
+        (x1) =>
+            (x2) =>
+                (x3) =>
+                    (x4) =>
+                        (x5) =>
+                            (x6) =>
+                                (x7) =>
+                                    (
+                                      x8,
+                                    ) =>
```
natebosch, line 13: What is happening here?

munificent: Oops! The short answer is that in the benchmark case file, I accidentally copy/pasted:

```
40 columns                              |
```

Which is trying to fit this into 40 columns, unlike the original regression test which is a full 80 columns. That's where there's a weird split here. I fixed that.

The longer answer is that it seems like splitting multiple of the parameter lists would help it pack things into the page width in this weird case, but the solver doesn't seem to figure that out. I believe that's at least in part because the AssignPiece used for `=>` doesn't have the parameter list as the left operand of the AssignPiece itself and instead makes that part of the surrounding FunctionPiece.

I spent a little time poking around at that this morning but didn't make anything better. I think it probably makes sense to worry about that in the context of #1466.

natebosch (approved):

benchmark/case/curry.expect
natebosch: It's probably not a big deal if we have a weird behavior right at the edge of the intended width. It does stand out, though, seeing the 1 level in the middle of the nesting formatted differently from the identical syntax before and after it.

kallentu (approved):

Issue #558, opened 2016-11-24
Use a long annotation w/ parameters with trailing commas constructor


matanlurey:
Apologies for the long issue name. The code will make it more clear, me thinks.

Nice, just using `@required`:

```dart
const factory RouteDefinition.fork(
  String path,
  List<RouteDefinition> children, {
  @required String defaultTo,
}) = ForkedRouteDefinition._;
```

Also works fine with something like `@Required('I told you so, Bob')`.

Less nice, using `@Required(... with a long enough string here...)`:

```dart
const factory RouteDefinition.fork(
  String path,
  List<RouteDefinition> children, {
  @Required('When `path` is navigated to, what path within `children` to use')
      String defaultTo,
//^^<-- What is this doing here?
}) = ForkedRouteDefinition._;
```

munificent:
Ah, yes. In this case, the annotation and the parameter don't all fit on one line.

I had to figure out what to do in this case and there wasn't a lot of hand-formatted prior art in the corpus to go on, so I kind of just picked something that was reasonable to implement.

srawlins:
Are you just expecting fewer spaces, Matan?

matanlurey:
I mean yeah, I'm expecting it to line up nicely :-/

srawlins:
This has been labelled trailing-commas, but it applies without commas as well. The formatter currently outputs code like:

```dart
void reallyLongFunctionName0(
    @Deprecated('This has been deprecated for many interesting reasons...')
        arg1) {
  return;
}

void fn1(
    {@Deprecated('This has been deprecated for many interesting reasons...')
        String argument1}) {
  return;
}
```

munificent:
Thanks for pointing that out, Sam!

munificent:
In the [forthcoming](https://github.com/dart-lang/dart_style/issues/1403) [tall style](https://github.com/dart-lang/dart_style/issues/1253), these examples get formatted as:

```dart
class RouteDefinition {
  const factory RouteDefinition.fork(
    String path,
    List<RouteDefinition> children, {
    @Required('When `path` is navigated to, what path within `children` to use')
    String defaultTo,
  }) = ForkedRouteDefinition._;
}
```

And:

```dart
void reallyLongFunctionName0(
  @Deprecated('This has been deprecated for many interesting reasons...')
  arg1,
) {
  return;
}

void fn1({
  @Deprecated('This has been deprecated for many interesting reasons...')
  String argument1,
}) {
  return;
}
```

[...]

PR #1478, opened 2024-05-09
Correct indentation on parameters with both metadata and default values.

### LINKED ISSUE #1461: [tall] Incorrect indentation with multiple annotations on a parameter
I'm migrating regression tests and the tall style produces:

```dart
greet(
  @Rest(valueHelp: 'who', help: 'Name(s) to greet.') List<String> who, {
  @Group.start(title: 'Output')
      @Option(help: 'How many !\'s to append.')
      int enthusiasm:
      0,
  @Flag(abbr: 'l', help: 'Put names on separate lines.') bool lineMode: false,
  @Option(name: 'greeting', help: 'Alternate word to greet with e.g. "Hi".')
      String salutation:
      'Hello',
}) {}
```

There are several things wrong here:

* The subsequent `@Option` annotations shouldn't be indented.
* The parameter shouldn't be indented.
* There's no reason to split at the `:`.

munificent:
The AST node for a parameter with both metadata and a default value looks like:

    DefaultValueFormalParameter
      FormalParameter

The inner node has the metadata and the outer one has the default value. Prior to this PR, that meant that we'd create an AssignPiece for the default value whose left-hand side piece was the function parameter and its metadata. That meant that when the metadata split, the outer AssignPiece would be forced to split.

This slightly awkward change makes the piece structure reflect how the user (and, alas, not Analyzer) thinks of the syntax where the metadata is part of the entire parameter and the default value is inside the parameter.

Fix #1461.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -623,6 +623,7 @@ mixin PieceFactory {
       pieces.add(parameter);
       return;
     }
+
     var (separator, value) = defaultValue;
     var operatorPiece = pieces.build(() {
       if (separator.type == TokenType.EQ) pieces.space();
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -619,24 +619,24 @@ mixin PieceFactory {
   /// Otherwise, just writes [parameter].
   void writeDefaultValue(
       Piece parameter, (Token separator, Expression value)? defaultValue) {
-    if (defaultValue case (var separator, var value)) {
-      var operatorPiece = pieces.build(() {
-        if (separator.type == TokenType.EQ) pieces.space();
-        pieces.token(separator);
-        if (separator.type != TokenType.EQ) pieces.space();
-      });
-
-      var valuePiece = nodePiece(value, context: NodeContext.assignment);
-
-      pieces.add(AssignPiece(
-          left: parameter,
-          operatorPiece,
-          valuePiece,
-          canBlockSplitRight: value.canBlockSplit));
-    } else {
-      // No default value.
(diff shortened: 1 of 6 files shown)


srawlins:
Woohoo! Just for paper-trail purposes, I believe this will fix https://github.com/dart-lang/dart_style/issues/558 when Tall Style is The Style.

lib/src/front_end/piece_factory.dart
```diff
@@ -604,10 +613,37 @@ mixin PieceFactory {
...
+
+      pieces.add(AssignPiece(
+          left: parameter,
+          operatorPiece,
+          valuePiece,
+          canBlockSplitRight: value.canBlockSplit));
+    } else {
+      // No default value.
+      pieces.add(parameter);
+    }
```
natebosch, line 639: Consider refactoring to a guard clause and reducing indentation on the interesting path.

```suggestion
    if (defaultValue == null) {
      pieces.add(parameter);
      return;
    }
    var (separator, value) = defaultValue;
    var operatorPiece = pieces.build(() {
      if (separator.type == TokenType.EQ) pieces.space();
      pieces.token(separator);
      if (separator.type != TokenType.EQ) pieces.space();
    });

    var valuePiece = nodePiece(value, context: NodeContext.assignment);

    pieces.add(AssignPiece(
        left: parameter,
        operatorPiece,
        valuePiece,
        canBlockSplitRight: value.canBlockSplit));
```

natebosch (approved):

kallentu (approved):

lib/src/front_end/piece_factory.dart
munificent: Great idea, thanks!

PR #1485, opened 2024-05-14
Don't use AssignPiece for for-in loop "in" clauses.

munificent:
One of the reasons that AssignPiece is so hairy is because it handles both splitting after the operator ("=", ":", etc.) and before "in".

Splitting out for-in loops to their own Piece class lets us get rid of a little of that complexity.

Also, I thought that was the only reason that the operator is a separate Piece. That's actually not true because we also need the LHS to be its own piece for applyConstraints().

Even so, I still think it's a little easier to understand with for-in loops being their own piece.

I also slightly tweaked the formatting of for-in loops. In the rare case where the left side block splits, we now indent it and split at "in". This is consistent with the old style. I don't have a strong preference, but I think it looks a little neater this way.

### DIFF
--- a/lib/src/piece/assign.dart
+++ b/lib/src/piece/assign.dart
@@ -81,7 +81,7 @@ class AssignPiece extends Piece {
   /// The left-hand side of the operation.
   final Piece? _left;
 
-  // TODO(rnystrom): If it wasn't for the need to constraint [_left] to split
+  // TODO(rnystrom): If it wasn't for the need to constrain [_left] to split
   // in [applyConstraints()], we could write the operator into the same piece
   // as [_left]. In the common case where the AssignPiece is for a named
   // argument, the name and `:` would then end up in a single atomic
--- a/lib/src/piece/assign.dart
+++ b/lib/src/piece/assign.dart
@@ -5,13 +5,15 @@ import '../back_end/code_writer.dart';
 import '../constants.dart';
 import 'piece.dart';
 
-/// A piece for any construct where `=` is followed by an expression: variable
-/// initializer, assignment, constructor initializer, etc.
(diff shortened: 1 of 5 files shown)


test/tall/statement/for_in.stmt
```diff
@@ -113,9 +113,10 @@ for (var (longIdentifier &&
 for (var [longIdentifier, anotherLongOne] in obj) {;}
 <<<
 for (var [
-  longIdentifier,
-  anotherLongOne,
-] in obj) {
+      longIdentifier,
+      anotherLongOne,
+    ]
+    in obj) {
```
kallentu, line 119: I think this is easier to read for me.

kallentu (approved):

lib/src/piece/assign.dart
```diff
@@ -76,25 +78,21 @@ class AssignPiece extends Piece {
...
   final Piece? _left;
 
-  // TODO(perf): Most AssignPieces don't allow splitting between [_left] and
-  // [_operator]. Also, in the common case where the AssignPiece is for a named
-  // argument, then both [_left] and [_operator] will be simple CodePieces. If
-  // we used a single piece for both, they can often be concatenated into a
-  // single [CodePiece]. We only store [_operator] separately for for-in loops.
-  // Consider handling those with a separate Piece class and merging [_left]
-  // and [_operator] in this one.
+  // TODO(rnystrom): If it wasn't for the need to constraint [_left] to split
```
natebosch, line 84: [nit] typo

```suggestion
  // TODO(rnystrom): If it wasn't for the need to constrain [_left] to split
```

natebosch (approved):

PR #1486, opened 2024-05-14
Support running benchmarks in AOT mode.

munificent:
Many of the optimizations I'm trying have a relatively small improvement. That can make it hard to tell if they are actually an improvement if there is enough variance in the runs that they are lost in the noise.

AOT builds don't need warm-up time and seem to have generally lower variance than JIT runs. Also, they are how the shipped formatter is run in `dart format`. So this adds support to the two benchmark scripts for running themselves in AOT snapshot mode.

I added support for this directly to the scripts instead of just manually compiling and running them on the command line (which works fine) because this way it's less error-prone. I don't have to remember to re-compile the AOT snapshot and risk using an old build which is an easy mistake to make when the only behavioral difference is (possibly) performance.

Also, tweaked the profiler output a little.

### DIFF
--- a/benchmark/directory.dart
+++ b/benchmark/directory.dart
@@ -4,23 +4,27 @@
 
 import 'dart:io';
 
+import 'package:args/args.dart';
 import 'package:dart_style/dart_style.dart';
 import 'package:dart_style/src/constants.dart';
 import 'package:dart_style/src/profile.dart';
+import 'package:dart_style/src/testing/benchmark.dart';
 
 /// Reads a given directory of Dart files and repeatedly formats their contents.
 ///
 /// This allows getting profile information on a large real-world codebase
 /// while also factoring out the file IO time from the process.
-void main(List<String> args) {
-  if (args.length != 1) {
-    stderr.writeln('Usage: dart run directory.dart <directory to format>');
-    exit(65);
-  }
+void main(List<String> arguments) async {
+  var directory = await _parseArguments(arguments);
 
-  var directory = args[0];
   print('Listing entries in $directory...');
   var entries = Directory(directory).listSync(recursive: true);
+
+  // Make sure the benchmark is deterministic. The order shouldn't really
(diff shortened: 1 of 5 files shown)


benchmark/directory.dart
```diff
@@ -4,24 +4,28 @@
...
 
-  var directory = args[0];
   print('Listing entries in $directory...');
   var entries = Directory(directory).listSync(recursive: true);
 
+  // Make sure the benchmark is deterministic. The order shouldn't really
+  // matter for performance, but since the JIT is warming up as it goes through
+  // the files, different orders could potentially affect how it chooses to
+  // optimize.
+  entries.sort((a, b) => a.path.compareTo(b.path));
```
natebosch, line 27: [optional] If you add an import to `package:collection`

```suggestion
  entries.sortBy((e) => e.path);
```

natebosch (approved):

benchmark/directory.dart
munificent: TIL. Done!

benchmark/run.dart
```diff
@@ -175,10 +180,15 @@ double _runTrial(DartFormatter formatter, ParseStringResult parseResult,
...
   argParser.addFlag('help', negatable: false, help: 'Show usage information.');
+  argParser.addFlag('aot',
+      negatable: false,
+      help: 'Whether the benchmark should run in AOT mode versus JIT.');
   argParser.addFlag('short',
       abbr: 's',
       negatable: false,
       help: 'Whether the formatter should use short or tall style.');
+  argParser.addFlag('no-warmup',
+      negatable: false, help: 'Skip the JIT warmum runs.');
```
kallentu, line 191: Nit: Typo in the help.

kallentu (approved):

benchmark/run.dart
munificent: Oops!

PR #1503, opened 2024-07-24
Fix firstNonCommentToken in anticipation of analyzer fix.

stereotype441:
This commit fixes a subtle bug in the formatter which is currently covered up by an analyzer bug. The formatter bug is this: `AstNodeExtensions.firstNonCommentToken` assumes that if an AST node doesn't implement `AnnotatedNode`, then it can't possibly begin with a documentation comment, so it is safe to simply defer to `beginToken`. But that isn't true `DefaultFormalParameter` doesn't implement `AnnotatedNode`, but its first child is a `NormalFormalParameter`, which *can* begin with a documentation comment.

Currently, analyzer bug https://github.com/dart-lang/sdk/issues/56313 prevents `NormalFormalParameter.beginToken` from properly accounting for the presence of the documentation comment, so as a result, `AstNodeExtensions.firstNonCommentToken` works correctly on `DefaultFormalParameter`. In order to avoid breaking the formatter when I fix the analyzer bug, I need to first fix the formatter.

The formatter fix is for `AstNodeExtensions.firstNonCommentToken` to recognize that `DefaultFormalParameter` is unusual, and to compute its first non-comment token directly (by looking at the first non-cmoment token of its first child, `parameter`). The formatter fix works regardless of whether the analyzer bug is fixed, so it is safe to land the formatter fix first.

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -18,6 +18,9 @@ extension AstNodeExtensions on AstNode {
     return switch (this) {
       AnnotatedNode(metadata: [var annotation, ...]) => annotation.beginToken,
       AnnotatedNode(firstTokenAfterCommentAndMetadata: var token) => token,
+      // DefaultFormalParameter is not an AnnotatedNode, but its first child
+      // (parameter) *is* an AnnotatedNode, so we can't just use beginToken.
+      DefaultFormalParameter(:var parameter) => parameter.firstNonCommentToken,
       _ => beginToken
     };
   }


munificent (approved):
LGTM.

Honestly, I wish I didn't need this extension at all. Is there a reason that the doc comment is included as part of `beginToken`? Conceptually, I don't think of the doc comment as being part of the AST node.

stereotype441:
> LGTM.
> 
> Honestly, I wish I didn't need this extension at all. Is there a reason that the doc comment is included as part of `beginToken`? Conceptually, I don't think of the doc comment as being part of the AST node.

Haha, yeah, just this morning I had a conversation with @bwilkerson about exactly this topic. He pointed out that the `visitChildren` methods *do* consider the doc comment to be part of the AST node, so for consistency it makes sense for `beginToken` to consider them part of the AST node too.

Now, you could argue that maybe `visitChildren` shouldn't consider the doc comment to be part of the AST node either, but  if we made that change, then a lot of nice analysis server functionality would break (like, for example, the fact that renaming an identifier propertly updates comments that reference that identifier). 🤷 

I suppose another approach would be to add `firstNonCommentToken` as an explicit part of the analyzer API. Then it would be the analyzer team's responsibility to make sure it works correctly 😃

Issue #1402, opened 2024-02-21
Make the formatter aware of the language version of what it's formatting


munificent:
Dart format is moving to a new "tall" style designed to be more readable for the kind of deeply nested declarative code that is common in Flutter (#1253). When users install a new Dart SDK, they don't generally expect the formatter to radically change the style of their already-formatted code, and we don't want that to happen when the new style ships either.

We've discussed a few ways to let users control when they adopt the new tall style, and the best approach we've been able to come up with is to tie it to language version. Let's say that `dart format` support for the tall style ships in Dart SDK 3.*x*.0. The idea is:

- Any Dart file whose language version is 3.x or greater will automatically be formatted using the new tall style.

- Any Dart file whose language version is older than 3.x will continue to use the current "short" style.

This means that by default, when you update to a new Flutter or Dart SDK, your formatting won't change. It's only when you change the Dart SDK constraint in your pubspec to opt into the new language version that you'll also get a formatting change. (We also intend to support some explicit way to opt in to the new style for older language versioned files.)

To implement that, dart_style needs to know the language version of everything it's formatting. Currently, it doesn't. This issue is to track adding support for that. It may need some elaboration, but the basic idea is:

- [x] The `FormatCommand` class which drives the `dart format` command-line tool and accesses the file system directly should determine the language version of each file it processes. This means looking for either a pubspec or package configuration file in surrounding directories to determine the default language version. I'm not sure what the exact logic should be here, but it should probably follow how `dart analyze` behaves. Ideally, there is some code we can reuse.

- [x] Extend the `DartFormatter` library API to support passing in a language version for each file. If omitted, default to something reasonable (without hitting the file system). Probably just the most recent language version.

- [x] If a compilation unit being formatted (from the command line or through the `DartFormatter` library API) has a `@dart=` language version comment, honor that to determine the style used for that file.

munificent:
I'm working on this now. One key question is how the formatter should behave when it can't determine the language version of what it's formatting. Cases where that can occur:

* You run `dart format` on the command line and the file paths you pass in aren't inside a package with a `package_config.json` file that specifies a default language version.

* You run `dart format`, pass in the code to format on stdin, and don't specify some sort of `--language-version=` option.

* You use the `DartFormatter` library API and don't pass in a language version.

[...]

PR #1506, opened 2024-07-30
Allow passing the language version in to the DartFormatter API.

munificent:
For users of the dart_style library API, this lets them control which language version formatted files are parsed as. Currently, the constructor parameter is optional and defaults to the latest supported version. In a future major release, it will become required.

Right now, the language version affects how the formatted code is parsed, but does not affect the applied style. When the tall-style experiment flag ships, any language version later than the Dart SDK version that the tall style ships in which get the tall style applied.

cc @jakemac53 @kallentu


munificent:
This is part of #1402.

natebosch (approved):

PR #1508, opened 2024-07-31
Use surrounding packages to infer language version in the CLI.

munificent:
When formatting file paths using the formatter CLI, look for a surrounding package (identified by a package config) to infer the default (i.e. not `// @dart=`) language version for each formatted file.

Currently, this behavior only happens when the tall-style experiment is enabled. When that experiment ships, it will be the default behavior. If the user passes an explicit language version using "--language-version", then no directory searching happens.

This only affects the CLI behavior. The library API never searches the file system. Also, when passing input to the formatter CLI over stdin, it doesn't search the file system.

### DIFF
--- a/bin/format.dart
+++ b/bin/format.dart
@@ -138,7 +138,7 @@ void main(List<String> args) async {
   if (argResults.rest.isEmpty) {
     await formatStdin(options, selection, argResults['stdin-name'] as String);
   } else {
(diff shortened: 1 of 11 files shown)


munificent:
This is the last big piece of engineering needed for #1402. Work after this is mostly coordination with other teams.

natebosch (approved):

PR #1509, opened 2024-08-01
Force a cascade to split if an indirect method call section splits.

munificent:
We always split a cascade if it has multiple sections. We don't require it to split if it has only one:

```dart
cascadeTarget..method(argument);
```

We also don't require it to split if the section is a method call whose argument list splits:

```dart
cascadeTarget..method(
  argument,
);
```

But the section might not be an immediate method call like `method()` here. It could be some longer call chain as in:

```dart
cascadeTarget..some.property.chain.method(argument);
```

In that case, if the ultimate trailing argument list splits, I think it's better to force the cascade section to split too:

```dart
cascadeTarget..some.property.chain.method(
  argument,
);

// Better:
cascadeTarget
  ..some.property.chain.method(
    argument,
  );
```

Otherwise, I think the preceding call chain gets sort of buried in the line with the cascade target.

This makes that change.

### DIFF
--- a/lib/src/front_end/chain_builder.dart
+++ b/lib/src/front_end/chain_builder.dart
@@ -78,11 +78,25 @@ class ChainBuilder {
         var piece = _visitor.nodePiece(section);
 
         var callType = switch (section) {
+          // If the section is itself a method chain, then force the cascade to
+          // split if the method does, as in:
+          //
+          //     cascadeTarget
+          //       ..methodTarget.method(
+          //         argument,
+          //       );
+          MethodInvocation(target: _?) => CallType.unsplittableCall,
+
+          // Otherwise, allow a direct method call in the cascade to not split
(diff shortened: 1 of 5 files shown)


natebosch (approved):

natebosch:
Seems like an improvement. I think it would be OK to force `..method(` to split too, but either way is fine since that situation isn't going to come up enough to have much impact.

munificent:
> I think it would be OK to force `..method(` to split too, but either way is fine since that situation isn't going to come up enough to have much impact.

I've thought about that too. I'm *right* on the fence with it. In simple cases (which are basically the only cases left after this PR), I think it looks pretty much OK. For example:

```dart
      var s =
          new Serialization()..addRuleFor(
            Various,
            constructor: "Foo",
            constructorFields: ["d", "e"],
          );
```

As you say, it doesn't come up much.

PR #1513, opened 2024-08-03
Add regression tests for all issues fixed by the new formatter.

munificent:
I went through the entire issue tracker. For every bug whose output now looks like we want using the new formatter, I added a regression test and closed the issue. These are those regression tests.

There are no code changes. It's just tracking issues where the new formatter already does the right thing.


test/tall/regression/other/misc.unit
```diff
@@ -19,21 +19,4 @@ SomeVeryLongReturnType someFunctionName(
...
-  // ...
-}
-<<<
-abstract class StreamNotifierProviderBase<
-  NotifierT extends AsyncNotifierBase<T>,
-  T
->
-    extends ProviderWithNotifier<AsyncValue<T>, NotifierT>
-    with FutureModifier<T> {
-  // ...
```
munificent, line 38: I removed this from here because it's covered by 1348.unit. I found this snippet in a text file of mine and didn't realize it was already in an issue.

natebosch (approved):
RSLGTM

PR #1522, opened 2024-08-13
Allow disabling formatting for a region of code.

### LINKED ISSUE #361: dartfmt should recognize off/on delimiter comment pairs to temporarily disable reformatting
Since no formatter is perfect (e.g., some const/final nested collection constructs try to represent actual readable structure meaning in the code via their indentation that the formatter cannot be expected to understand and preserve), formatter off/on comment pairs should be recognized that turn off reformatting for the specified lines in between these delimiters.

```
// dartfmt off
Foo barBaz() => new Quux({
    'alpha.beta.gamma': [new Zuul({
      'yarrr': Abracadabra.OPEN_SESAME,
    }, {
    'toil': {
      'trouble': {
        'fireBurn': {
            'cauldronBubble': EYE_OF_NEWT,
        }
      }
    }
  })]
});
// dartfmt on
```

Yes, use of this construct should be discouraged and limited. Perhaps emit a warning as dartfmt is run indicating that it encountered such blocks of unformatted code?

Also, it is only reasonable to require complete constructs be present between the delimiters, and not expect the formatter to do anything reasonable when the delimiters arbitrarily break up the middle of some block level construct. For example, the formatter could produce an error, or just not be expected to do anything sane, for misuses like:

```
  if (someTest) {
    someThenCode();
// dartfmt off
               unformattedThenCode();
} else {
          unformattedElseCode();
// dartfmt on
    moreElseCode();
  }
```

At worst, such unformatted segments that span block boundaries should produced undefined results for the rest of the source file. At best, an error should be printed and formatting should stop.

munificent:
Allow disabling formatting for a region of code.

There are two special line comments:

```
// dart format off
// dart format on
```

Any code between those keeps its original formatting completely, including indentation. It just splats the original unformatted code into the output.

Users can have as many of those pairs of comments as they want and formatting is disabled in them. The comments do not nest.

I didn't update the docs or CHANGELOG because this feature only works in the new tall style which is still behind an experiment flag. When it ships, I'll update the docs.

Fix #361.


srawlins:
Drive-by suggestion: the analyzer offers a few "warnings" for other machine-meaningful comments (mostly `// @dart=x.y`), for example maybe when it _looks_ like the user is trying to write a Dart Language Version comment, but didn't get it quite right. We could offer similar "warnings" or lint rules for "dart-format-toggling" comments. E.g. misspelled "dart format on/off", "dart format yes/no", "dart format true/false", `/* dart format on/off */` "off" without "on", "on" without "off", etc. In order to support this feature, we'd take any static analysis suggestions in the dart-lang/linter or dart-lang/sdk repo issue trackers. Or if it is expected to be used exceedingly rarely, these ideas may be just low priority.

kallentu (approved):

lib/src/back_end/code.dart
```diff
@@ -132,49 +149,230 @@ final class _MarkerCode extends Code {
...
+          //
+          // Here, `bef  +  ore` is some amount of code appearing before
+          // formatting is disabled, `off` is the `// dart format off` comment,
+          // `code` is some code inside the unformatted region, `|` is the
+          // selection marker, `inside` is more code in the unformatted region,
+          // `on` turns formatting back on, and `more` is formatted code at the
+          // end.
+          //
+          // We know the beginning of the unformatted region is at offset 10
+          // (just after the comment) in the original ource. We know the
```
natebosch, line 310:
```suggestion
          // (just after the comment) in the original source. We know the
```

lib/src/back_end/code.dart
```diff
@@ -132,49 +149,230 @@ final class _MarkerCode extends Code {
...
+          //     bef  +  ore off code | inside on more
+          //
+          // Here, `bef  +  ore` is some amount of code appearing before
+          // formatting is disabled, `off` is the `// dart format off` comment,
+          // `code` is some code inside the unformatted region, `|` is the
+          // selection marker, `inside` is more code in the unformatted region,
+          // `on` turns formatting back on, and `more` is formatted code at the
+          // end.
+          //
+          // We know the beginning of the unformatted region is at offset 10
```
natebosch, line 309: I don't understand how "just after the comment" is "offset 10". How are you counting offset?

natebosch (approved):

lib/src/back_end/code.dart
munificent: Oops! Sorry for the confusion. I tweaked the sample line of code after writing the paragraph explaining it and then forgot to go back and fix the offsets. Fixed.

PR #1537, opened 2024-08-17
Always split switch expressions

### LINKED ISSUE #1529: Tall style: Never write expression switches in a single line
Consider a codebase using pattern matching.  
I think it is very important for readability to have every "=>" be on a separate newline.

In my codebase, I had:

```dart
    return switch (that) {
      Family() => that,
      $FamilyOverride() => that.from,
    };
```

But it changed to:

```dart
    return switch (that) { Family() => that, $FamilyOverride() => that.from };
```

This change makes it much harder to view all the cases that are matched against.

It's difficult to differentiate `Foo() => really * Long() * expression` from `Foo() => short, Bar() => short2`.
I think that pattern matching is a special case and should always use trailing commas.

munificent:
I figured that since other comma-delimited expression forms don't split
if they don't have to, switch expressions should be the same. But I
went and checked on a huge corpus and every single place where a switch
expression wasn't split was clearly much worse than splitting it would
be.

Fix #1529.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -1718,7 +1718,7 @@ class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
     }
 
     list.rightBracket(node.rightBracket);
-    pieces.add(list.build());
+    pieces.add(list.build(forceSplit: node.cases.isNotEmpty));
   }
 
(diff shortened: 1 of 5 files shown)


kallentu (approved):
I'm in favor of this change also.
Thanks for doing it!

natebosch (approved):

PR #1547, opened 2024-08-23
Delete unreachable `default` clause.

stereotype441:
The Dart analyzer will soon be changed so that if the `default` clause of a `switch` statement is determined to be unreachable by the exhaustiveness checker, a new warning of type
`unreachable_switch_default` will be issued. This parallels the behavior of the existing `unreachable_switch_case` warning, which is issued whenever a `case` clause of a `switch` statement is determined to be unreachable. For details see https://github.com/dart-lang/sdk/issues/54575.

This PR deletes an unreachable `default` clause from `dart_style` now, to avoid a spurious warning when the analyzer change lands.

---

<details>

</details>

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -837,8 +837,6 @@ mixin PieceFactory {
               tokenPiece(combinatorNode.keyword),
               for (var name in names) tokenPiece(name.token, commaAfter: true),
             ]));
-          default:
-            throw StateError('Unknown combinator type $combinatorNode.');
         }
       }
 


munificent (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -837,8 +837,6 @@ mixin PieceFactory {
               tokenPiece(combinatorNode.keyword),
               for (var name in names) tokenPiece(name.token, commaAfter: true),
             ]));
-          default:
-            throw StateError('Unknown combinator type $combinatorNode.');
```
munificent, line 841: Will removing this cause a non-exhaustive switch compile error for versions of Dart that *don't* have your planned fix?

stereotype441: For this specific case, no, for two reasons:
- The type of the scrutinee is `Combinator`, which is sealed, so it's an "always exhaustive" type. Hence, flow analysis will assume the switch statement is exhaustive even if it can't prove it (because it knows that the exhaustiveness checker will come around later and generate a compile-time error if it's not).
- And besides, even if flow analysis _didn't_ assume the switch statement was exhaustive, there would be no error because there's nothing that would go wrong if the the switch statement _weren't_ exhaustive (there no locals that would fail to be initialized, no missing `return`s, no failed promotions, etc.)

However, you're right to be concerned about this in the general case. My initial implementation of the new `unreachable_switch_default` warning didn't account for whether the scrutinee was an always exhaustive type, and so there were some rare cases (<1%) where it wasn't, and that meant that sometimes the user would get an `unreachable_switch_default` warning that they couldn't safely "fix" by removing the `default` clause. That was a bad situation.

Fortunately I got some pushback when I tried to modify Google3 to account for this, and that was enough to convince me to change the implementation. Assuming https://dart-review.googlesource.com/c/sdk/+/382120 passes code review, the `unreachable_switch_default` warning will only fire for always-exhaustive types, which means that it will _always_ be safe to drop a `default` clause that's flagged as unreachable.

PR #1548, opened 2024-08-28
Tweak block argument heuristics

### LINKED ISSUE #1527: Tall style: Missing trailling comma when passing a closure with a block body to a function
Hello again!
I encountered the following when playing with the new style:

```dart
  Result<Res> chain<Res>(Res Function(T value) cb) {
    return when(
      data: (value) {
        try {
          return Result.data(cb(value));
        } catch (err, stack) {
          return Result.error(err, stack);
        }
      },
      error: Result.error,
    );
  }

  T get dataOrThrow {
    return when(
      data: (value) => value,
      error: (err, stack) {
        // ignore: only_throw_errors
        throw err;
      },
    );
  }
```

Became:

```dart
  Result<Res> chain<Res>(Res Function(T value) cb) {
    return when(data: (value) {
      try {
        return Result.data(cb(value));
      } catch (err, stack) {
        return Result.error(err, stack);
      }
    }, error: Result.error);
  }

  T get dataOrThrow {
    return when(data: (value) => value, error: (err, stack) {
      // ignore: only_throw_errors
      throw err;
    });
  }
```

I don't mind so much that the formatter removes the trailing comma when a "block" is the last parameter of function invocation. So I'm not against the `dataOrThrow` changes.

But I think that for `chain`, that's a big downgrade.

It is very difficult to stop that `error: Result.error` named parameter with the new change. It's easy for the eye to gloss over that parameter, by thinking that it's part of the logic of the block above.

IMO, if a closure with a block body is used and isn't the last function parameter, we should have a trailing comma.
This would make it easier to view all named parameters.

### LINKED ISSUE #1528: Tall style: Multiline array literals in a parameter list should induce a trailing commas
Apologies for the spam. I hope this isn't inappropriate!  
I came across an issue related to multiline array literals inside ConstructorInvocations.

The following:

```dart
          result = ProviderScope(
            overrides: [
              selectedCharacterId.overrideWithValue(split.last),
            ],
            child: const CharacterView(),
          );
```

Now becomes:

```dart
          result = ProviderScope(overrides: [
            selectedCharacterId.overrideWithValue(split.last),
          ], child: const CharacterView());
```

That feels incorrect. I assume this is a bug?

### LINKED ISSUE #1543: Tall style: Add trailing comma if a named parameter's value is splitted
This:

```dart
void main() async {
  fn(
    param1: true,
    paraaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaam2: [
      ['value'],
    ],
  );
}
```

becomes:

```dart
void main() async {
  fn(param1: true, paraaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaam2: [
    ['value'],
  ]);
}
```

This is similar to
- #1528, but I think named parameters should gravitate to a trailing comma more than positional ones, and some suggestions there will be bad for named parameters, like this one:

```dart
fn([a], [
  b
  c,
]);
```

The problem is, with named parameters, a name is very important, and name-value pair should stay coupled and ideally separate from other pairs. If a distinction of a line break is introduced between `param2` and its value, at least the same distinction should be introduced between the pairs.

I take the rule of pair distinction further, and my personal preference is to have a trailing comma if we have at least 2 named parameters:

```dart
void main() async {
  fn(
    param1: true,
    param2: true,
  );
}
```

munificent:
This produces output that better matches what users expected in the associated issues. I think it also looks better in the affected regression tests. And I ran this on a large corpus and looking at the diffs, these rules seem to be much better across the board.

The new rules are basically:

* A block formatted argument list can't end up with named arguments on multiple lines.

* Only one trailing argument is allowed after the block argument.

The existing rules are also kept:

* Prefer a function block argument over a collection one.

* Only allow a single block argument.

* Allow the first argument to be adjacent strings with a later function block argument (to handle calls like `test()` and `group()`).

Fix #1527.
Fix #1528.
Fix #1543.

### DIFF
--- a/lib/src/ast_extensions.dart
+++ b/lib/src/ast_extensions.dart
@@ -7,6 +7,16 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'piece/list.dart';
 
 extension AstNodeExtensions on AstNode {
+  /// When this node is in an argument list, what kind of block formatting
+  /// category it belongs to.
+  BlockFormat get blockFormatType => switch (this) {
+        AdjacentStrings(indentStrings: true) =>
+          BlockFormat.indentedAdjacentStrings,
+        AdjacentStrings() => BlockFormat.unindentedAdjacentStrings,
+        Expression(:var blockFormatType) => blockFormatType,
+        _ => BlockFormat.none,
+      };
+
   /// The first token at the beginning of this AST node, not including any
   /// tokens for leading doc comments.
   ///
--- a/lib/src/front_end/delimited_list_builder.dart
+++ b/lib/src/front_end/delimited_list_builder.dart
@@ -62,8 +62,6 @@ class DelimitedListBuilder {
       });
     }
 
-    if (_style.allowBlockElement) _setBlockElementFormatting();
-
     var piece =
         ListPiece(_leftBracket, _elements, _blanksAfter, _rightBracket, _style);
     if (_mustSplit || forceSplit) piece.pin(State.split);
@@ -134,8 +132,8 @@ class DelimitedListBuilder {
   /// [addCommentsBefore()] for the first token in the [piece].
   ///
   /// Assumes there is no comma after this piece.
-  void add(Piece piece, [BlockFormat format = BlockFormat.none]) {
-    _elements.add(ListElementPiece(_leadingComments, piece, format));
+  void add(Piece piece) {
+    _elements.add(ListElementPiece(_leadingComments, piece));
     _leadingComments.clear();
     _commentsBeforeComma = CommentSequence.empty;
   }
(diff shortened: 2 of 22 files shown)


natebosch (approved):
FWIW I'd be on board to recommend using named arguments anywhere to move named `test` and `group` arguments before the callback argument for the best formatting.

munificent:
> FWIW I'd be on board to recommend using named arguments anywhere to move named `test` and `group` arguments before the callback argument for the best formatting.

I should have mentioned this initially but I investigated this. I agree that it would be better for those named arguments to be moved before the lambda, but it turns out that it's still pretty common in existing code for those arguments to be trailing. I looked at all calls to functions named `test()`, `group()`, and `testWidgets()` in a corpus of pub packages:

-- Unnamed (4126 total) --
   3312 ( 80.271%): ("...", () {...})
    459 ( 11.125%): ("...", () {...}, name: arg)
    173 (  4.193%): (arg)
     48 (  1.163%): (arg, arg)
     44 (  1.066%): ("...", arg)
     42 (  1.018%): (arg, () {...})
      8 (  0.194%): ()
      8 (  0.194%): ("...", name: arg, () {...})
      6 (  0.145%): (arg, () {...}, name: arg)
      5 (  0.121%): (arg, arg, arg)
      4 (  0.097%): (arg, arg, name: arg, name: arg, name: arg, name: arg, name: arg, name: arg)
      3 (  0.073%): ("...", () {...}, name: "...")
      3 (  0.073%): ("...", () {...}, name: arg, name: arg)
      3 (  0.073%): ("..\n..", () {...})
      1 (  0.024%): (arg, () {...}, name: arg, name: arg, name: arg, name: arg, name: arg)
      1 (  0.024%): (arg, arg, name: arg)
      1 (  0.024%): (arg, arg, name: arg, name: arg, name: arg, name: arg, name: arg)
      1 (  0.024%): ("...", arg, name: "...")
      1 (  0.024%): (arg, name: arg, () {...})
      1 (  0.024%): (arg, () {...}, name: arg, name: arg, name: arg, name: arg)
      1 (  0.024%): ("...", arg, name: arg, name: arg, name: arg, name: arg, name: arg, name: arg)
      1 (  0.024%): ("...", arg, name: arg, name: arg, name: arg, name: arg)
```

Over 10% of calls have a trailing named argument, so I figured it was worth allowing that directly to avoid disrupting code more than necessary.

natebosch:
> it turns out that it's still pretty common in existing code for those arguments to be trailing.

Yeah, there are a whole bunch from before named-arguments-anywhere was an option, and I suspect they get copy/paste often enough that many authors aren't even aware of the option. Making the formatter work well for the existing corpus sounds good.

PR #1555, opened 2024-09-04
Handle imports with "as" and "if" clauses in either order.

munificent:
PR #1550 fixed the bug where if you had both clauses, it would output them in the *wrong* order. But it turns out that the parser today allows them in *either* order even though the language spec doesn't.

Since I've seen a handful of cases like this in the wild, instead of just failing to format, simply preserve the clause order that the user wrote.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -812,8 +812,26 @@ mixin PieceFactory {
         pieces.visit(directive.uri);
       });
 
-      // Include the `as` clause.
+      // Add all of the clauses and combinators.
       var clauses = <Piece>[];
+
+      // The language specifies that configurations must appear after any `as`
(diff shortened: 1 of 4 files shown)


natebosch:
> instead of just failing to format

Does it fail to format, or does it change the order to the correct one in the formatted code?

munificent:
Prior to this PR, it would change the order to the correct one in the formatter output. But that would then cause the formatter's own internal "I only make whitespace changes" sanity check to fail since it would see that as a non-whitespace change, which would then cause the formatter to abort writing that file.

With this PR, it just keeps the order that the user wrote, which also then means it is only making whitespace changes.

natebosch:
Oh I see from the other thread it trips the detection that something other than whitespace changed...

natebosch (approved):

PR #1556, opened 2024-09-04
Don't crash on an index cascade section followed by a method chain.

munificent:
I found this weird corner case in the wild when formatting a large corpus.


natebosch (approved):
Consider expanding the commit message with an explanation of why that code snippet results in an IndexExpression with a null `target`.

munificent:
Good idea, done.

PR #1560, opened 2024-09-05
Fix test that fails on Windows.

munificent:
I guess Dart on Windows uses `/` as a path separator in stack traces?

I'm not sure if it's safe to rely on that, so just tweaking the expectation to not care about the separator at all.


natebosch (approved):
I think we do guarantee URI format, so it would be safe to use `p.url.join` here if you want, but this is fine too.

kallentu (approved):

PR #1561, opened 2024-09-05
Don't allow a block argument to be named.

### LINKED ISSUE #1552: New style: Minor change causes huge whitespace diff
The code below was formatted before I changed a minor detail and then formatted again. Can you spot what the actual non-whitespace change is?

Screenshot 2024-08-29 at 10 49 42 AM

(The change is that a key param was added to the Stack widget)

These kind of dramatic diffs for a minor code change are going to make code reviews harder.

### LINKED ISSUE #1559: [tall-style] Trailing comma in argument list dropped
I'm not sure why the trailing comma after the single named argument is dropped here: 

https://dart-review.googlesource.com/c/sdk/+/383722/1/pkg/compiler/lib/src/js_backend/backend_impact.dart#434

The net effect for this file is that it is less readable because _sometimes_ the name of the argument is tucked out of sight on the previous line.  The file as a whole is less consistent (compare with line 421).

munificent:
There are three ways an argument list can be formatted:

```dart
// 1. All inline:
function(argument, another);

// 2. Block formatted:
function(argument, [
  element,
  element,
]);

// 3. Tall style:
function(
  argument,
  another,
);
```

One of the trickiest parts of the formatter is deciding the heuristics to choose between 2 and 3.

Prior to this PR, there is a fairly subtle rule: Arguments before the block argument can be named or not, the block argument can be named or not, and the argument after the block argument can be named or not, but only one of those three sections can have a named argument.

The intent of that rule is to allow named block arguments and named non-block arguments, while avoiding multiple argument names on the same line when block formatted. Also, it allows an argument list containing only a named argument to be block formatted:

```dart
function(name: [
  element,
]);
```

From looking at how contributors to Flutter have hand-formatted their code, it doesn't look like this level of complexity is well motivated. And allowing a single named argument to be block formatted but not if there are other named arguments means that adding a second named argument to a function can cause the entire argument list to be reformatted and indented differently. That can be a lot of churn if the block argument is itself a large nested widget tree.

This PR tweaks that rule to something simpler:

1. The block argument must be positional.

2. Other non-block arguments have no restrictions on whether they can be named or not.

This reduces diff churn and means that widget trees (which almost always use named arguments) have a more consistent (but taller) style across the entire tree.

From talking to Michael Goderbauer, this seems like a reasonable trade-off to me.

Fix #1552.
Fix #1559.

### DIFF
--- a/lib/src/front_end/delimited_list_builder.dart
+++ b/lib/src/front_end/delimited_list_builder.dart
@@ -444,32 +444,14 @@ class DelimitedListBuilder {
     var candidateIndex = _candidateBlockArgument(arguments);
     if (candidateIndex == -1) return;
 
+    // The block argument must be positional.
+    if (arguments[candidateIndex] is NamedExpression) return;
+
     // Only allow up to one trailing argument after the block argument. This
     // handles the common `tags` and `timeout` named arguments in `test()` and
     // `group()` while still mostly having the block argument be at the end of
     // the argument list.
     if (candidateIndex < arguments.length - 2) return;
-
-    // If there are multiple named arguments, they should never end up on
-    // separate lines (unless the whole argument list fully splits). Otherwise,
-    // it's too easy for an argument name to get buried in the middle of a line.
-    // So we look for named arguments before, on, and after the candidate
-    // argument. If more than one of those sections of arguments has a named
-    // argument, then we don't allow the block argument.
-    var namedSections = 0;
-    bool hasNamedArgument(int from, int to) {
-      for (var i = from; i < to; i++) {
-        if (arguments[i] is NamedExpression) return true;
-      }
-
-      return false;
-    }
-
-    if (hasNamedArgument(0, candidateIndex)) namedSections++;
-    if (hasNamedArgument(candidateIndex, candidateIndex + 1)) namedSections++;
-    if (hasNamedArgument(candidateIndex + 1, arguments.length)) namedSections++;
-
-    if (namedSections > 1) return;
 
     // Edge case: If the first argument is adjacent strings and the second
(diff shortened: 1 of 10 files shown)


munificent:
cc @goderbauer

kallentu (approved):

lib/src/front_end/delimited_list_builder.dart
```diff
@@ -444,33 +444,15 @@ class DelimitedListBuilder {
     var candidateIndex = _candidateBlockArgument(arguments);
     if (candidateIndex == -1) return;
 
+    // The block argument must be positional.
+    if (arguments[candidateIndex] is NamedExpression) return;
```
natebosch, line 448: Do you have an example of the undesired formatting that would result if we omit this line?

natebosch (approved):

lib/src/front_end/delimited_list_builder.dart
munificent: Before this PR, you can have output like:

```dart
Stack(children: [
  widget,
  widget,
]);
```

With this change, you get:

```dart
Stack(
  children: [
    widget,
    widget,
  ],
);
```

If the argument is positional, it can still be block formatted, so this is still fine:

```dart
setState(() {
  ...
});
```

In practice, it's pretty rare for a functional call to have only one named argument that could be block formatted, so it doesn't make a big difference in practice. What it does avoid is diff churn if a function call goes from one to multiple named arguments and then is no longer able to block format what used to be the only argument (#1552).

Another way to think about it is that it using named arguments means you are getting tall style formatting (or all inline if they fit).

PR #1563, opened 2024-09-10
Make the language version parameter to DartFormatter mandatory.

munificent:
This is a breaking change, so bumping the major version of dart_style, the first time I've done that since the null safety migration (!).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.0.0-wip
+
+* Make the language version parameter to `DartFormatter()` mandatory.
+
 ## 2.3.7
 
 * Allow passing a language version to `DartFomatter()`. Formatted code will be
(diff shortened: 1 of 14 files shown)


kallentu (approved):

natebosch (approved):
We will need to make sure google3 is ready for this before we roll it to the SDK.

Mechanically we can either patch this in to a CL and run a global test suite before we land it, or we can land this and roll to google3 before we roll to the SDK to run the tests.

munificent:
Yeah, I'm expecting that this means dart_style can't be rolled into google3 for a while until we're ready for it. If you'd like, I can set up a "3.0" branch and land the changes there so that the main branch remains in a state where we can land bug fixes and roll into google3 if you think it's worth it.

My thinking was to just go ahead and land directly on main. Then if it turns out we do need to land some bug fixes without the 3.0 changes, I can always just branch off a previous commit at that point in time.

natebosch:
> My thinking was to just go ahead and land directly on main. Then if it turns out we do need to land some bug fixes without the 3.0 changes, I can always just branch off a previous commit at that point in time.

Landing and delaying the roll until we're ready SGTM.

PR #1566, opened 2024-09-13
Stdin name search

### LINKED ISSUE #1530: Add a `--stdin-path` argument
The vim plugin formats using stdin so that we can format without writing the buffer. We added ` --stdin-name` argument so that error messages can include the file name and makes it easier for us to associate the errors back to the buffer. Now that formatting depends on language version can we add a `--stdin-path` argument to ask the formatter to treat the content as if it had been written to that location, and find the language version accordingly.

https://github.com/dart-lang/dart_style/issues/1402#issuecomment-2292271853

munificent:
Use --stdin-name as a file path for inferring the language version.

If --stdin-name is given, then look for a surrounding package config to
infer the default language version from for code read from stdin. To
avoid that behavior, you can pass in a language version explicitly
using --language-version.

Fix #1530.

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -41,6 +41,8 @@ Future<void> formatStdin(
   // If they didn't specify a version or a path, default to the latest.
   languageVersion ??= DartFormatter.latestLanguageVersion;
 
+  var name = path ?? 'stdin';
+
   var completer = Completer<void>();
   var input = StringBuffer();
   stdin.transform(const Utf8Decoder()).listen(input.write, onDone: () {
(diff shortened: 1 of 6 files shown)


munificent:
I split this into two commits because I also needed to fix the grouping of some tests. You might want to review the commits separately to get an easier diff.

lib/src/io.dart
```diff
@@ -26,23 +26,37 @@ Future<void> formatStdin(
...
-      options.beforeFile(null, name);
+      options.beforeFile(null, path ?? 'stdin');
       var source = SourceCode(input.toString(),
-          uri: name,
+          uri: path,
           selectionStart: selectionStart,
           selectionLength: selectionLength);
       var output = formatter.formatSource(source);
-      options.afterFile(null, name, output,
+      options.afterFile(null, path ?? 'stdin', output,
```
natebosch, line 59: [optional] Consider adding a `name` variable to do this defaulting once.

natebosch (approved):

PR #1571, opened 2024-10-04
Allow configuring project-wide page width using a surrounding analysis_options.yaml file

### LINKED ISSUE #833: Rethink the 80 characters line limit
I am not sure if this is the right repository to discuss  this, but I would like to propose a investigation if 80 characters per line should still be the recommendation and dartfmt default.

The Dart style guide says:

>Readability studies show that long lines of text are harder to read because your eye has to travel farther when moving to the beginning of the next line. This is why newspapers and magazines use multiple columns of text.

I think this is perfectly valid for regular text for newspapers and magazines, that usually come in the form of:

>Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem
Ipsum has been the industry's standard dummy text ever since the 1500s, when an
unknown printer took a galley of type and scrambled it to make a type specimen book. 

But recently with the soaring success of Flutter, we are finding deeply nested code, which leads to lots of space being wasted as pure indentation. This picture shows how one of my Flutter widgets has just about half its line wasted in indentation, forcing me to break lines all the time:

image

Bear in mind, this is just a 9 Widget deep tree: 
Provider>Scaffold>SafeArea>WillPopScope>Column>Consumer>Padding>Row>Text.

And I didn't have any classes there with long names. If I did, it would have been a multi-lined mess.

IMHO, we can't use studies made for newspapers for coding styles, specially after the usage has become so different in the Flutter age.

I have seen people in the past defending 80 character lines saying it fits better all monitors, but nowadays with everyone using 24" monitors 1080+ I don't think that's longer the case.

I don't want to propose a new number, like 120 characters, but instead I would like to suggest that Google should launch a research on this, taking in consideration Flutter Apps and what line character count fits better their code.

My hope is that we don't turn out in the end with just 40 characters per line as real usable space due to indentations and enforced standards.

munificent:
When using the `dart format` CLI (and *not* the library API), if the user doesn't specify a page width using `--line-length`, then the formatter will walk the directories surrounding each formatted file looking for an `analysis_options.yaml` file. If one is found, then it looks for a configured page width like:

```yaml
formatter:
  page_width: 123
```

If found, then the file is formatted at that page width. If any sort of failure occurs, the default page width is used instead.

This is hidden behind the "tall-style" experiment flag and the intent is to ship this when the rest of the new tall style ships.

This is a fairly large change. To try to make it easier to review, I broke it into a series of hopefully more digestible commits. You might want to review those separately.

Fix #833.

cc @mit-mit


lib/src/analysis_options/merge_options.dart
```diff
@@ -0,0 +1,65 @@
...
+/// *   Lists are concatenated without duplicates.
+/// *   A list of strings is promoted to a map of strings to `true` when merged
+///     with another map of strings to booleans. For example `['opt1', 'opt2']`
+///     is promoted to `{'opt1': true, 'opt2': true}`.
+/// *   Maps unioned. When both have the same key, the corresponding values are
+///     merged, recursively.
+/// *   Otherwise, a non-`null` override replaces a default value.
+Object? merge(Object? defaults, Object? overrides) {
+  return switch ((defaults, overrides)) {
+    (List(isAllStrings: true) && var list, Map(isToBools: true)) =>
```
lrhn, line 21: Wouldn't it be nice if you could just write:
```dart
  (List(every(isA<String>): true) && var list, Map(values.every(isA<bool>))) =>
```
and not need extension getters, just a simple `bool isA<T>(Object? value) => value is T;`. :wink: 
and not *need* a getter? :wink:

lib/src/cli/format_command.dart
```diff
@@ -189,9 +189,12 @@ class FormatCommand extends Command<int> {
...
-            '"${argResults['line-length']}".');
+    int? pageWidth;
+    if (argResults.wasParsed('line-length')) {
+      pageWidth = int.tryParse(argResults['line-length'] as String) ??
+          usageException('--line-length must be an integer, was '
+              '"${argResults['line-length']}".');
+    }
 
     var indent = int.tryParse(argResults['indent'] as String) ??
         usageException('--indent must be an integer, was '
```
lrhn, line 200: Should the be an `if (line-length < 1)` test. (Or `< _minimumSaneWidth`? I'd be fine with refusing to work for widths less than twice the longest keyword in the language, or something.)

lib/src/short/line_splitting/solve_state.dart
```diff
@@ -525,7 +525,7 @@ class SolveState {
     var unboundConstraints = <Rule, Set<int>>{};
     for (var unbound in _unboundRules) {
       // Lazily create and add the set to the constraints only if needed.
-      late final disallowedValues = unboundConstraints[unbound] = {};
+      late var disallowedValues = unboundConstraints[unbound] = {};
```
lrhn, line 528: Tricky. I like.
(But ti's only used in one place, so I'd avoid the complexity of understanding `late` and when its expression is evaluated, and just  do:
```dart
  Set<int>? disallowedValues;
  ....
     (disallowedValues ??= unboundConstraints[unbound] = {}).add(value)
```
It's the same thing, just more direct.
Maybe I just don't like `late`. Yep, that's it.)

lib/src/analysis_options/merge_options.dart
munificent: Yes! I would love some kind of way to embed predicates inside patterns.

lib/src/cli/format_command.dart
munificent: I don't know. I don't think it's much of a problem in practice.

lib/src/short/line_splitting/solve_state.dart
munificent: This is in the old formatter code, so I'm trying not to touch it too much since it will go away at some point when we no longer support the old style. The only reason there's a change here is because I turned on the "unnecessary_file" lint to be more consistent about using `var` for locals.

But, I agree, using `late` for a local is always a little funny feeling. But in this case, laziness is exactly what I want and the language does support it, so I figured I may as well use it.

lib/src/analysis_options/analysis_options_file.dart
```diff
@@ -0,0 +1,116 @@
...
+/// If there any "package:" includes, then they are resolved to file paths
+/// using [resolvePackageUri]. If [resolvePackageUri] is omitted, an exception
+/// is thrown if any "package:" includes are found.
+Future<AnalysisOptions> readAnalysisOptions(
+        FileSystem fileSystem, FileSystemPath optionsPath,
+        {ResolvePackageUri? resolvePackageUri}) async =>
+    _readAnalysisOptions(fileSystem, resolvePackageUri, optionsPath);
+
+Future<AnalysisOptions> _readAnalysisOptions(FileSystem fileSystem,
+    ResolvePackageUri? resolvePackageUri, FileSystemPath optionsPath) async {
```
natebosch, line 65: [nit] I don't think the private name is adding much value, it only saves one argument name at one callsite... Consider moving this to the body of the public definition.

lib/src/analysis_options/analysis_options_file.dart
```diff
@@ -0,0 +1,116 @@
...
+  var yaml = loadYamlNode(await fileSystem.readFile(optionsPath));
+
+  // Lower the YAML to a regular map.
+  if (yaml is! YamlMap) return const {};
```
natebosch, line 69: [nit] Returning an empty map doesn't correspond to the comment above. Can you move the comment below this line, and consider adding a comment explaining why reading a non-map defaults to empty options.

lib/src/analysis_options/io_file_system.dart
```diff
@@ -0,0 +1,52 @@
...
+/// To create an instance of this, use [IOFileSystem.makePath()].
+class IOFileSystemPath implements FileSystemPath {
+  /// The underlying physical file system path.
+  final String path;
+
+  IOFileSystemPath._(this.path);
+}
+
+extension on FileSystemPath {
+  String get ioPath => (this as IOFileSystemPath).path;
```
natebosch, line 51: Would it make sense to add a generic to `FileSystem` so that we don't need to cast `FileSystemPath` instances?

Or can we use `covariant` to make the expectation visible in the signature?

natebosch (approved):

lib/src/analysis_options/merge_options.dart
lrhn: Well, do go vote on https://github.com/dart-lang/language/issues/2433 then! :)

lib/src/cli/format_command.dart
lrhn: I guess someone actually saying `--line-length=-4` isn't going to get much worse results than someone saying `--line-length=1`. And if it crashes the formatter, I guess they asked for it! Ship it!

lib/src/analysis_options/io_file_system.dart
munificent: I did consider that. I believe this is a somewhat famous design problem in programming languages. The problem is if you do that, all code that works with `FileSystem` has to itself become generic. You end up having to propagate the type parameter through entire regions of the program.

That's not what you want because most other code doesn't care about the concrete `FileSystemPath` type. The feature we would actually need to model this without using casts is abstract types, like Rust, ML and some other languages have. Something like:

```dart
abstract class FileSystem {
  abstract type Path;

  ...
}

class IOFileSystem implements FileSystem {
  type Path = IOFileSystemPath;

  ...
}

class IOFileSystemPath implements FileSystemPath { ... }
```

But Dart doesn't have that. OOP works really well for modeling a *single* hierarchy of types with encapsulation, but doesn't handle families of related types like this. "Family polymorphism" is the name for another feature that I think tackles these kinds of problems.

In the absence of that, I just went with downcasting. I like your suggestion to mark the method parameters with `covariant`, though. I'll do that.

lib/src/analysis_options/analysis_options_file.dart
munificent: Oh, good call. I think there used to be more difference between the public and private one before the code evolved away from it. Done.

PR #1572, opened 2024-10-04
Don't write trailing commas in split catch clauses.

### LINKED ISSUE #1570: Split catch clause incorrectly has trailing comma
In the rare case where the formatter decides to split the parameter-list-like part of a `catch` clause, it appends a trailing comma. This is consistent with other places in the language.

However, it appears that catch clauses don't actually allow a trailing comma there, so the result is invalid syntax. As with type parameter and type argument lists, it shouldn't put a trailing comma here.

munificent:
Apparently the language doesn't allow them. Who knew?

Fix #1570.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -775,7 +775,8 @@ mixin PieceFactory {
         pieces.token(catchKeyword);
         pieces.space();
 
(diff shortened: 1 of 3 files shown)


kallentu (approved):
Huh. I never knew this?

munificent:
Same. <shrug>

PR #1574, opened 2024-10-04
Allow a comment in code to explicitly configure the page width.

munificent:
If the code being formatted contains a comment anywhere in it like:

```
// dart format width=123
```

Then the code is formatted at that width instead of the default width.

The intent is that tools that generate code and then format it will put this comment in the generated code. That way, the tool doesn't have to handle the complex logic to find a surrounding analysis_options.yaml file and get the user's project-wide page width configuration.

End users may also want to use this comment in rare cases. For example, maybe they have a library that contains a large value of data that looks better as a big wide table but they don't want their entire project to have a wider page width.

### DIFF
--- a/lib/src/front_end/comment_writer.dart
+++ b/lib/src/front_end/comment_writer.dart
@@ -42,11 +42,28 @@ import '../comment_type.dart';
 /// written. When that [TextPiece] is output later, it will include the comments
 /// as well.
 class CommentWriter {
+  /// Regular expression that matches a format width comment like:
+  ///
+  ///     // dart format width=123
+  static final RegExp _widthCommentPattern =
+      RegExp(r'^// dart format width=(\d+)$');
+
   final LineInfo _lineInfo;
 
   /// The tokens whose preceding comments have already been taken by calls to
   /// [takeCommentsBefore()].
   final Set<Token> _takenTokens = {};
+
+  /// If we have encountered a comment that sets an explicit page width, that
+  /// width, otherwise `null`.
+  ///
+  /// The comment looks like:
+  ///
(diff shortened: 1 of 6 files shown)


lib/src/front_end/comment_writer.dart
```diff
@@ -106,6 +123,14 @@ class CommentWriter {
         if (comment == token.precedingComments) linesBefore = 2;
       }
 
+      // If this comment configures the page width, remember it.
+      if (_widthCommentPattern.firstMatch(text) case var match?
+          when _pageWidthFromComment == null) {
```
natebosch, line 128: [optional] Consider nesting the conditionals with the null check first to avoid doing regex work when it won't be used.

test/tall/other/format_width.unit
```diff
@@ -0,0 +1,118 @@
...
+  // dart format width=30
+  fitsUnsplitAt40   +   butNotAt30;
+}
+<<<
+main() {
+  // dart format width=30
+  fitsUnsplitAt40 +
+      butNotAt30;
+}
+>>> Comment anywhere affects all code.
```
natebosch, line 14: I don't think this is a useful behavior. Would it be difficult to restrict this to comments before directives?

natebosch (approved):

test/tall/other/format_width.unit
munificent: I thought it would, but it turned out to be easy.

I was thinking it might be hard because the formatter supports formatting individual statements too, but there was an easy place to slot it in.

The rule is that the comment must appear before the first non-comment token (basically the first token in the AST being formatted).

PR #1576, opened 2024-10-08
Better formatting for long type parameter bounds.

### LINKED ISSUE #1568: Line longer than 80 characters in class type parameters
https://dart-review.googlesource.com/c/sdk/+/386736 introduces several declarations that look similar to this:

```dart
abstract interface class TypeAnalyzerOperations<
        TypeStructure extends SharedTypeStructure<TypeStructure>,
        Variable extends Object,
        TypeParameterStructure extends SharedTypeParameterStructure<TypeStructure>,
        TypeDeclarationType extends Object,
        TypeDeclaration extends Object>
    implements FlowAnalysisOperations<Variable, SharedTypeView<TypeStructure>> {
  ...
}
```

Note that the line `TypeParameterStructure extends SharedTypeParameterStructure<TypeStructure>,` is longer than 80 characters. The formatter insists on placing this text all on the same line, causing failures in the `lines_longer_than_80_chars` lint.

To see more examples, search for the text `// ignore: lines_longer_than_80_chars` in https://dart-review.googlesource.com/c/sdk/+/386736.

munificent:
The tall style was already doing a better job on #1568 than the old short style does. (I don't know why it leaves an overly long line there. It seems like a bug but perhaps a nasty subtle one in the solver.)

But it still didn't allow splitting before `extends` in a bound which could leave long lines in cases where the type parameter and bound name are both quite long. Rare, but we may as well do a better job.

So this introduces a new piece that allows splitting before `extends` if needed.

Fix #1568.

cc @stereotype441 Take a look at the regression test if you want to see how the new style handles your example.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -10,7 +10,6 @@ import '../ast_extensions.dart';
 import '../constants.dart';
 import '../dart_formatter.dart';
 import '../piece/assign.dart';
-import '../piece/bound.dart';
 import '../piece/case.dart';
 import '../piece/constructor.dart';
 import '../piece/control_flow.dart';
@@ -18,6 +17,7 @@ import '../piece/infix.dart';
 import '../piece/list.dart';
 import '../piece/piece.dart';
 import '../piece/type.dart';
+import '../piece/type_parameter_bound.dart';
 import '../piece/variable.dart';
 import '../profile.dart';
 import '../source_code.dart';
@@ -1852,7 +1852,7 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
           pieces.visit(bound);
         });
 
-        pieces.add(BoundPiece(typeParameterPiece, boundPiece));
+        pieces.add(TypeParameterBoundPiece(typeParameterPiece, boundPiece));
       } else {
         // No bound.
         pieces.token(node.name);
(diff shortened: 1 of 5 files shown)


kallentu (approved):

lib/src/front_end/ast_node_visitor.dart
```diff
@@ -1840,12 +1841,21 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
   @override
   void visitTypeParameter(TypeParameter node) {
     pieces.withMetadata(node.metadata, inlineMetadata: true, () {
-      pieces.token(node.name);
       if (node.bound case var bound?) {
```
natebosch, line 1844: If only we had https://github.com/dart-lang/language/issues/2537 we could reduce indentation here.

lib/src/piece/bound.dart
```diff
@@ -0,0 +1,65 @@
...
+import '../back_end/code_writer.dart';
+import '../constants.dart';
+import 'piece.dart';
+
+/// A piece for a type parameter and its bound.
+///
+/// Handles not splitting before `extends` if we can split inside the bound's
+/// own type arguments, or splitting before `extends` if that isn't enough to
+/// get it to fit.
+final class BoundPiece extends Piece {
```
natebosch, line 13: [nit] since "bind" is a verb we can to to pieces this name might be confusing. Consider `TypeBoundPiece` or similar to make it more clearn.

natebosch (approved):

lib/src/piece/bound.dart
munificent: Oh, good point.

PR #1577, opened 2024-10-10
Rename "--line-length" option to "--page-width".

munificent:
The formatter has always been inconsistent about how to refer to the limit that it tries to fit code in. The option name (which is older than when I started working on the formatter) is "--line-length". But the public API and all of the internal code uses "page width".

This renames that option to "--page-width". The old name is still supported (but hidden in help) for backwards compatibility.

In practice, this option isn't used often, so this likely won't affect many users. We've already made a major version bump, so now seems like a good time to clean things up as much as possible.

cc @johnmccutchan since the inconsistency was bugging you. :)

### DIFF
--- a/test/cli/page_width_test.dart
+++ b/test/cli/page_width_test.dart
@@ -12,6 +12,54 @@ import '../utils.dart';
 void main() {
   compileFormatter();
 
+  group('--page-width', () {
+    test('must be a number', () async {
+      var process = await runFormatter(['--page-width=12.34']);
+      await process.shouldExit(64);
+    });
+
+    test('must be an integer', () async {
+      var process = await runFormatter(['--page-width=notint']);
+      await process.shouldExit(64);
+    });
+
+    test('must be positive', () async {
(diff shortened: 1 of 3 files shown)


test/cli/page_width_test.dart
```diff
@@ -12,6 +12,54 @@ import '../utils.dart';
...
+  group('--page-width', () {
+    test('must be a number', () async {
+      var process = await runFormatter(['--page-width=12.34']);
+      await process.shouldExit(64);
+    });
+
+    test('must be an integer', () async {
+      var process = await runFormatter(['--page-width=notint']);
+      await process.shouldExit(64);
+    });
```
natebosch, line 24: [nit] The test titles and conditions look swapped

```suggestion
    test('must be a number', () async {
      var process = await runFormatter(['--page-width=notint']);
      await process.shouldExit(64);
    });

    test('must be an integer', () async {
      var process = await runFormatter(['--page-width=12.34']);
      await process.shouldExit(64);
    });
```

natebosch (approved):

PR #1578, opened 2024-10-14
Ship the "tall-style" experiment.

munificent:
This PR does a few things:

- Remove the "tall-style" experiment flag and treat it as always enabled.

- Use language version instead of the experiment flag to determine which style you get. If the version is 3.7 or higher, you get the tall style. Otherwise, you get the short style.

- Clean up the CHANGELOG.md and README.md a bit. I'll probably do some more copy editing on them before this gets published, but I wanted to get things roughly in order.

cc @leafpetersen @mit-mit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ are included, some of them breaking:
   surrounding `analysis_options.yaml` file.
 
   This feature is mainly for code generators that generate and immediately
-  format code but don't know what about any surrounding `analysis_options.yaml`
+  format code but don't know about any surrounding `analysis_options.yaml`
   that might be configuring the page width. By inserting this comment in the
   generated code before formatting, it ensures that the code generator's
   behavior matches the behavior of `dart format`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,87 @@
 ## 3.0.0-wip
 
+This is a large change. Under the hood, the formatter was almost completely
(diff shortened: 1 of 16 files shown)


test/cli/page_width_test.dart
```diff
@@ -50,40 +47,20 @@ void main() {
...
-  });
-
   test('no options search if page width is specified on the CLI', () async {
     await d.dir('foo', [
       analysisOptionsFile(pageWidth: 20),
       d.file('main.dart', _unformatted),
     ]).create();
 
-    var process = await runFormatterOnDir([
-      '--language-version=latest', // Error to not have language version.
```
munificent, line 83: These `--language-version=latest` lines were no longer necessary as of #1573, but I forgot to clean them up in that PR.

CHANGELOG.md
```diff
@@ -1,5 +1,87 @@
...
+        thatSplitsAt30;
+  }
+  ```
+
+  This comment must appear before any code in the file and must match that
+  format exactly. The width set by the comment overrides the width set by any
+  surrounding `analysis_options.yaml` file.
+
+  This feature is mainly for code generators that generate and immediately
+  format code but don't know what about any surrounding `analysis_options.yaml`
```
natebosch, line 55: [nit] extra word

```suggestion
  format code but don't know about any surrounding `analysis_options.yaml`
```

natebosch (approved):

PR #1588, opened 2024-10-24
Don't move a comment between a modifier and return type.

### LINKED ISSUE #1585: Formatter moves inline comment in external declaration
Given this (sanitized) input:

```dart
class AaaaAaaaa {
  external Aaaaaa get aa;
  external Aaaaaa get aaAaa;
  external Aaaaaa get aa;
  external aaaa get aaaa;
  external /*AaaaaaAaaa*/ aaa get aa;
  external Aaaaaa get aaa;
}
```

The formatter outputs:

```dart
class AaaaAaaaa {
  external Aaaaaa get aa;
  external Aaaaaa get aaAaa;
  external Aaaaaa get aa;
  external aaaa get aaaa; /*AaaaaaAaaa*/
  external aaa get aa;
  external Aaaaaa get aaa;
}
```

munificent:
When formatting a function, method, getter, or function type, the formatter hoists any leading comments out so that they don't force a split between the return type and body.

However, it failed to take into account modifiers that may occur before the return type but after the comment. If that happened, the comment would get moved before the modifiers.

Fix #1585.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
+import 'package:collection/collection.dart';
 
 import '../ast_extensions.dart';
 import '../piece/adjacent.dart';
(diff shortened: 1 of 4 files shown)


kallentu (approved):

lib/src/front_end/piece_factory.dart
```diff
@@ -626,8 +627,9 @@ mixin PieceFactory {
     //     @meta
     //     // Weird place for comment.
     //     int f() {}
-    var leadingComments =
-        pieces.takeCommentsBefore(returnType.firstNonCommentToken);
+    var firstToken = modifiers.firstWhereOrNull((token) => token != null) ??
```
natebosch, line 630: [optional] You could use a non-null filter instead of writing it as a callback.

```suggestion
    var firstToken = modifiers.nonNulls.firstOrNull ??
```

natebosch (approved):

lib/src/front_end/piece_factory.dart
munificent: Well crap. I landed this before I saw your suggestion. Today I learned about `nonNulls`.

PR #1592, opened 2024-10-31
Clean up README and TODO comments.

munificent:
Updated the example in the README to one that uses the new style. I also took the opportunity to find a more interesting example and generally rewrote things based on how the formatter is used today.

When we were in the middle of the rewrite, I used "TODO(tall)" and "TODO(perf)" comments to track things I wanted to look into before I considered the rewrite "done". Now that the rewrite is basically good enough, I went through those. If I think the TODO is no longer worth doing at all, I just deleted it. Otherwise, I turned it into a general "TODO(rnystrom)" for us to keep in mind at some unspecified time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
-The dart_style package defines an automatic, opinionated formatter for Dart
-code. It replaces the whitespace in your program with what it deems to be the
-best formatting for it. Resulting code should follow the [Dart style guide][]
-but, moreso, should look nice to most human readers, most of the time.
+The dart_style package defines an opinionated automated formatter for Dart code.
+It replaces the whitespace in your program with what it deems to be the
+best formatting for it. It also makes minor changes around non-semantic
+punctuation like trailing commas and brackets in parameter lists.
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -14,80 +17,89 @@ The formatter turns code like this:
...
-choose **Reformat with Dart Style**.
+formatter. For example, in Visual Studio Code, formatting Dart code will use
+the dart_style formatter by default. Most users have it set to reformat every
+time they save a file.
 
 Here's a simple example of using the formatter on the command line:
 
-    $ dart format test.dart
+```sh
+$ dart format test.dart
```
natebosch, line 70: [very nitpicky] WDYT about `dart format foo.dart`? `test.dart` is not a file anyone should write since the convention is `foo_test.dart`.

natebosch (approved):

README.md
munificent: Oh, good point. Done (more or less).

kallentu (approved):

PR #1609, opened 2024-12-03
Bump to publish 3.0.0.

munificent:
Also fixed the format-self call in the grinder script which was still using the old CLI format.


srawlins (approved):
Ship it!

PR #1610, opened 2024-12-05
Ensure comment formatting is idempotent.

### LINKED ISSUE #1606: Two runs needed to format with awkward comment
This code

```
class AaaaAaaaAaaaAaaaaaaaAaaaaaaa {
  @aaaaaaaa
  AaaaaAaaaAaaaAaaaaaaa get aaaaAaaaAaaaaaaa => aaaaaaaaAaaaaaaaaAaaaaa
      ? (AaaaaaaaaAaaAaaaaaaa()
        // AAA/Aaaaa aaaaaaaaa aaaaaaaa aaaa AAAAAAAA aaaaaaaaa aaaaaa aaaaaaaa.
        ..aaaaAaaaaaaaaAaaaaaAaagetaaa =
            AaaaAaaaaaaaaAaaaaaAaagetaaa.aaaaAaaaaa(aaaaaaaa: [
          AaaaaaaaAaaaaaa()..aaaaaaaa = AaaaaaaaAA_Aaaaaaaa.AAAAA,
        ], aaaaaaAaaaaa: [
          AaaaaaAaaaaaAaaaaaa()
            ..aaaaaaAaaaaa = AaaaaaAaaaaa.AAAAAA_AAAAAAAAAA_AAAAAA,
          AaaaaaAaaaaaAaaaaaa()..aaaaaaAaaaaa = AaaaaaAaaaaa.AAAAAA_AAAA_AAAAAAA
        ]))
      : AaaaAaaaaaaAaaaaaaAaaaaaaa();
}
```

is only fully formatted after two formatter runs:

```
dartfmt t.dart > t1.dart; dartfmt t1.dart > t2.dart; diff t1.dart t2.dart
5c5
<           ? (AaaaaaaaaAaaAaaaaaaa()// AAA/Aaaaa aaaaaaaaa aaaaaaaa aaaa AAAAAAAA aaaaaaaaa aaaaaa aaaaaaaa.
---
>           ? (AaaaaaaaaAaaAaaaaaaa() // AAA/Aaaaa aaaaaaaaa aaaaaaaa aaaa AAAAAAAA aaaaaaaaa aaaaaa aaaaaaaa.

```

munificent:
In some cases, a line comment can appear between two tokens that otherwise never split, like after "if (" and before the condition. That leads to some tricky edge case behavior. If you formatted:

```dart
if (
  // Comment
  condition) {
  ;
}
```

It would see the newline before the comment and decide the comment was a "leading comment" which means it gets attached to the condition expression. Then the formatter would output it like:

```dart
if (// Comment
  condition) {
  ;
}
```

That's because leading comments don't write a newline before themselves. Then if you format that again, there's no newline before the `//`, so now it's a hanging comment. Hanging comments get a space before them, so you get:

```dart
if ( // Comment
  condition) {
  ;
}
```

Really, leading comments (as the name implies) are intended to always begin a line. So this PR makes sure they do that.

While I was at it, I modified the test runner to run the formatter twice on *every* test to ensure that everything is idempotent. That doesn't *prove* that the formatter will always produce idempotent output, but it at least gives us pretty good test coverage that it *does* behave idempotent-ly.

In practice, most normal looking code would never hit this edge case. You have to put a comment in an unusual spot where a split doesn't occur.

This still feels like a fairly brittle part of the formatter to me. Comments appearing between tokens that never split otherwise is handled on a pretty ad hoc basis (which is why some of the tests in this PR have weird indentation). I'd like a cleaner more systematic solution, but I'm not sure what that would look like.

Fix #1606.

### DIFF
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -372,15 +372,14 @@ final class CodeWriter {
     // If we found a problematic line, and there is are pieces on the line that
     // we can try to split, then remember them so that the solution will expand
     // them next.
-    if (!_foundExpandLine) {
-      if (_currentLinePieces.isNotEmpty &&
-          (_column > _pageWidth || !_solution.isValid)) {
-        _expandPieces.addAll(_currentLinePieces);
-        _foundExpandLine = true;
-      } else {
-        // This line was OK, so we don't need to expand the pieces on it.
-        _currentLinePieces.clear();
-      }
+    if (_foundExpandLine) return;
+    if (_currentLinePieces.isNotEmpty &&
+        (_column > _pageWidth || !_solution.isValid)) {
+      _expandPieces.addAll(_currentLinePieces);
+      _foundExpandLine = true;
+    } else {
+      // This line was OK, so we don't need to expand the pieces on it.
+      _currentLinePieces.clear();
     }
   }
 }
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -372,13 +372,15 @@ final class CodeWriter {
     // If we found a problematic line, and there is are pieces on the line that
     // we can try to split, then remember them so that the solution will expand
     // them next.
-    if (!_foundExpandLine && (_column > _pageWidth || !_solution.isValid)) {
-      // We found a problematic line, so remember the pieces on it.
-      _foundExpandLine = true;
-      _expandPieces.addAll(_currentLinePieces);
-    } else if (!_foundExpandLine) {
-      // This line was OK, so we don't need to expand the piece on it.
-      _currentLinePieces.clear();
+    if (!_foundExpandLine) {
+      if (_currentLinePieces.isNotEmpty &&
+          (_column > _pageWidth || !_solution.isValid)) {
+        _expandPieces.addAll(_currentLinePieces);
+        _foundExpandLine = true;
+      } else {
+        // This line was OK, so we don't need to expand the pieces on it.
+        _currentLinePieces.clear();
+      }
     }
   }
 }
--- a/test/utils.dart
+++ b/test/utils.dart
@@ -143,8 +143,8 @@ void _testFile(TestFile testFile) {
 
         // Make sure that formatting is idempotent. Format the output and make
         // sure we get the same result.
-        _validateFormat(formatter, actual, formatTest.output,
-            'was not idempotent', testFile.isCompilationUnit);
+        _validateFormat(formatter, actual, actual, 'was not idempotent',
+            testFile.isCompilationUnit);
       });
     }
   });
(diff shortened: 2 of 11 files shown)


lib/src/back_end/code_writer.dart
```diff
@@ -372,13 +372,15 @@ final class CodeWriter {
...
+    if (!_foundExpandLine) {
+      if (_currentLinePieces.isNotEmpty &&
+          (_column > _pageWidth || !_solution.isValid)) {
+        _expandPieces.addAll(_currentLinePieces);
+        _foundExpandLine = true;
+      } else {
+        // This line was OK, so we don't need to expand the pieces on it.
+        _currentLinePieces.clear();
+      }
     }
```
natebosch, line 384: [optional] Consider reducing nesting with a guard clause

```suggestion
    if (_foundExpandLine) return;
    if (_currentLinePieces.isNotEmpty &&
        (_column > _pageWidth || !_solution.isValid)) {
      _expandPieces.addAll(_currentLinePieces);
      _foundExpandLine = true;
    } else {
      // This line was OK, so we don't need to expand the pieces on it.
      _currentLinePieces.clear();
    }
```

test/utils.dart
```diff
@@ -134,34 +134,52 @@ void _testFile(TestFile testFile) {
...
+        var actual = _validateFormat(
+            formatter,
+            formatTest.input,
+            formatTest.output,
+            'did not match expectation',
+            testFile.isCompilationUnit);
+
+        // Make sure that formatting is idempotent. Format the output and make
+        // sure we get the same result.
+        _validateFormat(formatter, actual, formatTest.output,
```
natebosch, line 146: [optional] We know that `actual` and `formatTest.output` are equivalent given the above line passed, but my initial expectation reading the line performing a test for idempotency was to see `actual, actual` as the arguments.

```suggestion
        _validateFormat(formatter, actual, actual,
```

test/utils.dart
```diff
@@ -134,34 +134,52 @@ void _testFile(TestFile testFile) {
...
+  // the matcher output.
+  if (actualText != expected.text) {
+    fail('Formatting $reason. Expected:\n'
+        '${expected.text}\nActual:\n$actualText');
+  } else if (actual.selectionStart != expected.selectionStart ||
+      actual.selectionLength != expected.selectionLength) {
+    fail('Selection $reason. Expected:\n'
+        '${expected.textWithSelectionMarkers}\n'
+        'Actual:\n${actual.textWithSelectionMarkers}');
+  }
```
natebosch, line 178: I think using `expect` would give better output in some cases since it will point to the specific character of the first difference for the string equality test. What cases does this output have an advantage over `expect` with a `reason:` argument?

```suggestion
  expect(reason: 'Formatting $reason.', actualText, expected.text);
  expect(reason: 'Selection start $reason.', actual.selectionStart, expected.selectionStart);
  expect(reason: 'Selection length $reason.', actual.selectionLength, expected.selectionLength);
```

natebosch (approved):

test/utils.dart
munificent: Done, but I had to move some stuff around. The problem is that the actual doesn't *exactly* match the expectation when formatting a single statement because the actual output doesn't have a trailing newline (correct) while the expectation from the test file does (because each expectation ends on its own line).

Moved the code to handle this into TestFile so that the expectation is always exactly what we want.

test/utils.dart
munificent: It's really common for me to update a test expectation by simply copying the actual output from the failed test. Using `fail()` makes that easy: I just unindent the whole thing two spaces and it's done. The default output for strings turns them into string literals like:

```
    Actual: '  Widget build(BuildContext context) => CompositionRoot(\n'
              '    configureOverrides: configureOverrides,\n'
              '    compose:\n'
              '        (builder) =>\n'
              '            builder\n'
              '              //Adds a singleton CounterController to the container\n'
              '              ..addSingleton((container) => CounterController()),\n'
              '    child:\n'
              '    //We need the BuildContext from the Builder here so the children\n'
              '    //can access the container in the CompositionRoot\n'
              '    Builder(elided),\n'
              '    sdf,\n'
              '  );'
```

Which is much more annoying. Highlighting the first character difference is kind of nice, I admit. But in practice, it's usually pretty quick for me to spot the difference and being able to copy paste into tests is really handy.

PR #1612, opened 2024-12-06
Handle comments and metadata before variables more gracefully.

### LINKED ISSUE #1604: Tall style: field formatting
When the field declaration is long, so that the name has to be wrapped, it used to be wrapped with an indent, like this:

```
class C {
  final LoremIpsumDolorSitAmetFactory
      _loremIpsumDolorSitAmetFactory;
}
```

Now there's no indent, and it is harder to read, especially when there are many fields like that.

```
class C {
  final LoremIpsumDolorSitAmetFactory
  _loremIpsumDolorSitAmetFactory;
}
```

Also when there's a comment between an annotation and a type, field name is put on a new line:

Before:

```
  @override
  // ignore: hash_and_equals
  final int hashCode;
```

Now:

```
  @override
  // ignore: hash_and_equals
  final int
  hashCode;
```

munificent:
Comments and metadata before a declaration are always tricky because the formatter by default attaches comments to the innermost piece following the comment. So in:

```dart
// Comment
int field;
```

We want the comment attached to the Piece for the entire field declaration. But the formatter won't see the comment until it hits the `int` token and will end up attaching it to the piece for the type annotation which is embedded inside the field piece. That in turn means that the formatter will think there is a newline inside the type (because there is one between the comment and `int`) which then forces a split after the type annotation too:

```dart
// Comment
int
field;
```

In most places, this is handled by having the surrounding SequenceBuilder grab the comment before visiting the subsequent declaration. But that doesn't work when you have a comment after metadata:

```dart
@meta
// Comment
int field;
```

Now the comment isn't before the field declaration. It's stuck inside it. But we still want to hoist it out.

This PR fixes that for every place I could find: fields, functions, variables, and for-loop variables. The latter is particularly hard because for-in loops have some weird formatting already and it's also just a weird place for splits to occur.

I wish I had a cleaner more systematic way of handling these but despite trying for most of today, I haven't been able to come up with a cleaner approach. This at least gets the formatter producing better output.

Fix #1604.


lib/src/front_end/piece_factory.dart
```diff
@@ -1399,21 +1416,65 @@ mixin PieceFactory {
         canBlockSplitRight: canBlockSplitRight));
   }
 
-  /// Writes a [Piece] for the `<variable> in <expression>` part of a for-in
-  /// loop.
-  void writeForIn(AstNode leftHandSide, Token inKeyword, Expression sequence) {
+  /// Writes the `<variable> in <expression>` part of a for-in loop.
+  void _writeForIn(AstNode leftHandSide, Token inKeyword, Expression sequence) {
```
natebosch, line 1420: [nit] With one method `_writeDeclaredForIn` having more specificity in the name than `_writeForIn` I anticipated some layering or difference in specificity that doesn't exist in the implementation. WDYT about something like  `_writeIdentifierForIn` ?

natebosch (approved):

PR #1618, opened 2024-12-07
Add regression test for #624.

### LINKED ISSUE #624: Inconsistent indentation of list literal between variable and getter declarations.
```dart
var x = [                                       
  "long value that won't fit everywhere",       
  "long value that won't fit everywhere",       
  "long value that won't fit everywhere"        
];                                              
                                                
get y => [                                      
      "long value that won't fit everywhere",   
      "long value that won't fit everywhere",   
      "long value that won't fit everywhere"    
    ];                                          
```
In the var declaration, the list is indented by two, in the getter, it's indented by six. 
I'd prefer if they both used the var indentation.

munificent:
This is doing what users want in the new tall style. \o/

Close #624.

### DIFF
--- a/test/tall/regression/0600/0624.unit
+++ b/test/tall/regression/0600/0624.unit
@@ -1,0 +1,52 @@
+>>>
+var x = [
+  "long value that won't fit everywhere",
+  "long value that won't fit everywhere",
+  "long value that won't fit everywhere"
+];
+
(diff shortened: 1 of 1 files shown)


kallentu (approved):
We love less indentation.

natebosch (approved):

PR #1620, opened 2024-12-10
Format `||` patterns like fallthrough cases in switch expressions.

### LINKED ISSUE #1602: Prefer to split in `||` patterns in a switch expression instead of the value
Currently, the formatter formats the following switch expression case like so:

```dart
    return switch (_advance().type) {
      TokenType.float || TokenType.int || TokenType.string => LiteralExpression(
        _previous.value!,
      ),
    };
```

We already have special formatting for `||` patterns in a switch expression to make them look sort of like a series of fallthrough cases in a switch statement. Given that, I think it makes sense to also tweak the splitting costs so that it prefers to split on the `||` instead of in the case expression. That way, you get:

```dart
    return switch (_advance().type) {
      TokenType.float ||
      TokenType.int ||
      TokenType.string => LiteralExpression(_previous.value!),
    };
```

munificent:
Switch statements allow multiple cases to share a body like:

```dart
switch (obj) {
  case pattern1:
  case pattern2:
    body;
}
```

Switch expressions don't support that, but `||` patterns are the idiomatic way to accomplish the same thing. Because of that, the formatter has some special formatting when the outermost pattern in a switch expression case is `||`:

```dart
x = switch (obj) {
  pattern1 ||
  pattern2 =>
    body,
};
```

Note how the `pattern2` operand isn't indented.

This PR extends that special handling to allow the `=>` on the same line as the `=>` even if the pattern is a split `||` pattern, like:

```dart
x = switch (obj) {
  pattern1 ||
  pattern2 => body,
};
```

And it prefers to split the `||` over the body when the body is block formatted:

```dart
// Prefer:
x = switch (obj) {
  pattern1 ||
  pattern2 => function(argument),
};

// Over:
x = switch (obj) {
  pattern1 || pattern2 => function(
    argument,
  ),
};
```

This is one of those rules that's mostly a matter of taste, but I ran this on a large corpus and most of the diffs look better to me. Here are a few examples:

```dart
// Before:
      typeName = switch (targetType) {
        DriftSqlType.int || DriftSqlType.bigInt || DriftSqlType.bool =>
          'INTEGER',
        DriftSqlType.string => 'CHAR',
        DriftSqlType.double => 'DOUBLE',
        DriftSqlType.blob => 'BINARY',
        DriftSqlType.dateTime => 'DATETIME',
        DriftSqlType.any => '',
        CustomSqlType() || DialectAwareSqlType() => targetType.sqlTypeName(
          context,
        ),
      };

// After:
      typeName = switch (targetType) {
        DriftSqlType.int ||
        DriftSqlType.bigInt ||
        DriftSqlType.bool => 'INTEGER',
        DriftSqlType.string => 'CHAR',
        DriftSqlType.double => 'DOUBLE',
        DriftSqlType.blob => 'BINARY',
        DriftSqlType.dateTime => 'DATETIME',
        DriftSqlType.any => '',
        CustomSqlType() ||
        DialectAwareSqlType() => targetType.sqlTypeName(context),
      };

// Before:
    return switch (side) {
      AxisSide.right || AxisSide.left =>
        titlesPadding.vertical + borderPadding.vertical,
      AxisSide.top || AxisSide.bottom =>
        titlesPadding.horizontal + borderPadding.horizontal,
    };

// After:
    return switch (side) {
      AxisSide.right ||
      AxisSide.left => titlesPadding.vertical + borderPadding.vertical,
      AxisSide.top ||
      AxisSide.bottom => titlesPadding.horizontal + borderPadding.horizontal,
    };

// Before:
    final defaultConstraints = switch (side) {
      ShadSheetSide.top || ShadSheetSide.bottom => BoxConstraints(
        minWidth: mSize.width,
      ),
      ShadSheetSide.left || ShadSheetSide.right => BoxConstraints(
        minHeight: mSize.height,
      ),
    };

    final defaultConstraints = switch (side) {
      ShadSheetSide.top ||
      ShadSheetSide.bottom => BoxConstraints(minWidth: mSize.width),
      ShadSheetSide.left ||
      ShadSheetSide.right => BoxConstraints(minHeight: mSize.height),
    };
```

Fix #1602.

- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/lib/src/piece/case.dart
+++ b/lib/src/piece/case.dart
@@ -8,13 +8,20 @@ import 'piece.dart';
 
 /// Piece for a case pattern, guard, and body in a switch expression.
 final class CaseExpressionPiece extends Piece {
+  /// Split inside the body, which must be block formattable, like:
+  ///
+  ///     pattern => function(
+  ///       argument,
+  ///     ),
+  static const State _blockSplitBody = State(1, cost: 0);
+
   /// Split after the `=>` before the body.
-  static const State _beforeBody = State(1);
+  static const State _beforeBody = State(2);
 
   /// Split before the `when` guard clause and after the `=>`.
-  static const State _beforeWhenAndBody = State(2);
+  static const State _beforeWhenAndBody = State(3);
 
-  /// The pattern the value is matched against along with the leading `case`.
+  /// The pattern the value is matched against.
   final Piece _pattern;
 
   /// If there is a `when` clause, that clause.
@@ -54,6 +61,7 @@ final class CaseExpressionPiece extends Piece {
 
   @override
   List<State> get additionalStates => [
+        if (_canBlockSplitBody) _blockSplitBody,
         _beforeBody,
         if (_guard != null) ...[_beforeWhenAndBody],
       ];
@@ -61,10 +69,15 @@ final class CaseExpressionPiece extends Piece {
   @override
   bool allowNewlineInChild(State state, Piece child) {
     return switch (state) {
+      // If the outermost pattern is `||`, then always let it split even while
+      // allowing the body on the same line as `=>`.
+      _ when child == _pattern && _patternIsLogicalOr => true,
+
+      // There are almost never splits in the arrow piece. It requires a comment
+      // in a funny location, but if it happens, allow it.
       _ when child == _arrow => true,
-      State.unsplit when child == _body && _canBlockSplitBody => true,
-      _beforeBody when child == _pattern =>
-        _guard == null || _patternIsLogicalOr,
+      _blockSplitBody when child == _body => true,
+      _beforeBody when child == _pattern => _guard == null,
       _beforeBody when child == _body => true,
       _beforeWhenAndBody => true,
       _ => false,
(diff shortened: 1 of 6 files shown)


kallentu (approved):

lib/src/piece/case.dart
```diff
@@ -54,17 +61,23 @@ final class CaseExpressionPiece extends Piece {
...
         _beforeBody,
         if (_guard != null) ...[_beforeWhenAndBody],
       ];
 
   @override
   bool allowNewlineInChild(State state, Piece child) {
     return switch (state) {
+      // If the outermost pattern is `||`, then always let it split even while
+      // allowing the body on the same line as `=>`.
+      _ when child == _pattern && _patternIsLogicalOr => true,
```
natebosch, line 74: Why do we do this for `||` but not `&&`?

natebosch (approved):

lib/src/piece/case.dart
munificent: The intent of this formatting is to treat:

```dart
switch (obj) {
  constant ||
  another => body,
};
```

Sort of like:

```dart
switch (obj) {
  case constant:
  case another:
    body;
};
```

Since switch expressions don't allow multiple cases to share a body, using `||` as the top level pattern emulates that and yields similar formatting where each arm is as the beginning of the line (assuming they need to split).

In other words, readers seeing a `||` pattern at the top level of an expression switch case read it like a series of independent cases with a single body, and we format it like that.

There's no similar analogue for `&&`. In that pattern, you need to reason about all of the arms together to understand when the pattern matches.

PR #1627, opened 2024-12-17
Update grinder script.

munificent:
The workflow for shipping dart_style used to be that we published a new version on pub and then rolled it into the Dart SDK. That meant that the dart_style version number in the SDK accurately described what was in dart_style.

A few years ago, we changed to rolling into the SDK first. That lets us validate dart_style inside google3 before irrevocably publishing, which is a good thing. But it has an unfortunate side effect. If you run `dart format --version`, it shows the *previous* version number of dart_style because the version number only gets bumped when we publish. This is really confusing if you're trying to debug the formatter in the SDK and it appears to be an older version than it actually is.

To fix that, I'm tweaking the workflow slightly. When we begin working on a new version of the formatter, we bump the version number in the pubspec to, say, 3.4.5-wip. At that point, we immediately update the version number constant shown by `--version` to "3.4.5" (the version it *will* be). That way, when it gets rolled into the SDK, it prints the upcoming version number.

When a version of dart_style is about to be published, we remove `-wip` from the pubspec but the constant printed by `--version` is already done.

To make this workflow a little less error-prone, I updated the grinder script to add tasks for starting a new patch, minor, and major version.

I also updated the docs for the old task now that we use the workflow automation for publishing.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -26,48 +26,57 @@ Future<void> validate() async {
   Dart.run('bin/format.dart', arguments: ['.']);
 }
 
+/// Increments the patch version of the current version.
+@Task()
+Future<void> patch() async {
+  var version = _readVersion();
+  _updateVersion(
+      Version(version.major, version.minor, version.patch + 1, pre: 'wip'));
+}
+
+/// Increments the minor version of the current version.
+@Task()
+Future<void> minor() async {
+  var version = _readVersion();
+  _updateVersion(Version(version.major, version.minor + 1, 0, pre: 'wip'));
+}
+
+/// Increments the major version of the current version.
+@Task()
+Future<void> major() async {
+  var version = _readVersion();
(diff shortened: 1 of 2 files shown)


tool/grind.dart
```diff
@@ -26,48 +26,57 @@ Future<void> validate() async {
...
+/// 1.  Make sure the version in the pubspec is a "-wip" number. This should
+///     already be the case since you've already landed patches that change
+///     the formatter and bumped to that as a consequence.
 ///
-///   5. After the review is complete, land it:
+/// 2.  Run this task:
 ///
-///         git cl land
+///     ```
+///     pub run grinder bump
```
kevmoo, line 62:
```suggestion
///     dart run grinder bump
```

I think?

kevmoo (approved):

tool/grind.dart
munificent: Good catch, thanks.

PR #1629, opened 2024-12-17
Add regression test for #1628.

### LINKED ISSUE #1628: Inserting a block-comment before a property call chain wraps each chained element
I was surprised by this in the new formatting. Here is an input text:

```dart
bool f(bool b, int i) {
  return // Check b.
  b == true || i.isEven;
}
```

The formatter likes all that code on one line. If I want to add a comment for the `i.isEven` expression, then it insists on splitting:

```dart
bool f(bool b, int i) {
  return // Check b.
  b == true ||
          // Check i.
          i
          .isEven;
}
```

🤕 Ouch. If this is intentional, I'll trust your research, and you can close.

munificent:
Looks like this was already fixed by #1615.

Close #1628.

### DIFF
--- a/test/tall/regression/1600/1628.unit
+++ b/test/tall/regression/1600/1628.unit
@@ -8,3 +8,18 @@ bool f(bool b, int i) {
   return // Check b.
   b == true || i.isEven;
 }
+>>>
+bool f(bool b, int i) {
+  return // Check b.
+  b == true ||
+          // Check i.
+          i
+          .isEven;
+}
+<<<
+bool f(bool b, int i) {
+  return // Check b.
+  b == true ||
+      // Check i.
+      i.isEven;
+}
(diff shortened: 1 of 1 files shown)


test/tall/regression/1600/1628.unit
```diff
@@ -0,0 +1,10 @@
+>>>
+bool f(bool b, int i) {
+  return // Check b.
+  b == true || i.isEven;
```
natebosch, line 4: This reproduction is missing the `// Check i.` comment from the failure case in the filed issue

munificent: Oh dear, good catch and poor reading comprehension on my part. Added. (Fortunately, that case is passing too.)

natebosch (approved):

PR #1669, opened 2025-03-12
Update the SDK constraint and move onto the new formatter.

munificent:
Finally get to dogfood it in the dart_style repo!

I tweaked a couple of very minor things to play nice with the new style:

- A couple of test() calls had their test body indented because the description string didn't fit. Reworded or split the strings.

- Hoisted out one very large closure because it got indented farther.

- Rewrapped a couple of comments.

Otherwise, no real surprises.

I also took the opportunity to ensure every formatter test file had a consistent newline at the end.


natebosch (approved):
RSLGTM

PR #1672, opened 2025-03-14
Add an optional configuration to preserve trailing commas.

### LINKED ISSUE #1652: Preserve trailing commas
**TL;DR:** The new Dart formatter (3.7.0) automatically removes trailing commas, which changes code formatting in an undesirable way. This contradicts the **`require_trailing_commas`** linter rule and removes the ability to force line breaks. A workaround exists (adding `//` comments), but it's inconvenient. A setting to restore the old behavior or prevent automatic comma removal would be greatly appreciated.

---

Could there be an option to preserve the trailing commas, so they are not automatically removed? That way, developers who rely on them for readability or structure would still be able to use them without the formatter interfering. This would give us more control over the formatting style while still maintaining some level of automation.

I really liked the ability to force a line wrap by placing a comma at the end of a line. However, with the new formatter, this trailing comma is automatically removed.

This feels awkward and changes the code style in a way that I really don’t like.

It would be great if there were an option to prevent the automatic removal of trailing commas.

I'm not the only one raising concerns about this:

[Reddit discussion](https://www.reddit.com/r/FlutterDev/comments/1h1d9fd/the_new_formatter_of_dart_37/?rdt=64360)
[GitHub issue comment 1](https://github.com/dart-lang/dart_style/issues/1253#issuecomment-2661120645)
[GitHub issue comment 2](https://github.com/dart-lang/dart_style/issues/1253#issuecomment-2662235094)

---

As [jamesderlin](https://github.com/dart-lang/dart_style/issues/1253#issuecomment-2663907930) pointed out, there’s an annoying workaround:

I think I'm going to have to resign myself to adding a superfluous empty // comment at every site where I don't want the formatter to remove an explicit trailing comma.

Sure, this keeps the comma, but having to do this for every line? That’s just monkey business... 🐵🍌

---

As [intraector](https://github.com/dart-lang/dart_style/issues/1253#issuecomment-2662235094) mentioned:

The formatter should not remove trailing commas, but it should add them.

Consider the following example:  

### **Old style:**  
```dart
Material(
  color: Colors.transparent,
  child: InkWell(
    customBorder: const CircleBorder(),
    child: Padding(
      padding: const EdgeInsets.all(2),
      child: child,
    ),
    onTap: () {
      context.read<ScGalleryVm>().crop();
    },
  ),
)
```
### **New style:**  
```dart
Material(
  color: Colors.transparent,
  child: InkWell(
    customBorder: const CircleBorder(),
    child: Padding(padding: const EdgeInsets.all(2), child: child),
    onTap: () {
      context.read<ScGalleryVm>().crop();
    },
  ),
)
```  

In some cases, the new formatting even contradicts the **`require_trailing_commas`** linter rule.  

---

I completely agree **`require_trailing_commas`** should take precedence over this new change. It would be great if this behavior could be reconsidered.

---

Oneother example:

_Old way:_

Image

New way (BAD):
Image

My IDE even shows errors because it says a trailing comma is required. 🤕  
Sure, the new approach removes seven lines, but the formatting is much harder to read.

---

👍 Yes, I like the proposal.

👎 No, I don't like the proposal.

munificent:
In the old short formatter, when a delimited construct had a trailing comma, it had two effects:

1.  The construct would be formatted "Flutter style" where the closing bracket would be moved to the next line and the contents indented +2, as opposed to the earlier style where the bracket stays on the same line as the last element and the elements are indented +4).

2.  The construct is forced to split even if it otherwise wouldn't.

The new formatter always uses a Flutter-like style for commas-separated constructors, so trailing commas are no longer needed for point 1.

When it does split a construct, it always adds a trailing comma. This means that point 2 breaks reversibility. Because of that, and because many users don't want to decide for every construct whether it should split or not, the new tall formatter no longer does point 2. It will remove a trailing comma and collapse a construct if it decides to.

However, some users rely heavily on that feature and strongly prefer control over forcing argument lists and other constructs to split.

This PR restores that functionality with an opt-in configuration:

```
formatter:
  trailing_commas: preserve
```

When enabled, if a construct has a trailing comma, the formatter will always split it. Users that preferred that behavior from the old formatter can enable the option and continue to work that way.

This does not mean we are generally moving in the direction of a more configurable formatter. This was a behavior that the formatter already supported and we removed it to the detriment of some users' experience. We're restoring that existing behavior for those users who want it.

Fix #1652.

### DIFF
--- a/lib/src/config_cache.dart
+++ b/lib/src/config_cache.dart
@@ -130,19 +130,26 @@ final class ConfigCache {
           pageWidth = width;
         }
 
-        if (formatter case {'trailing_commas': String commas}) {
-          trailingCommas = switch (commas) {
-            'automate' => TrailingCommas.automate,
-            'preserve' => TrailingCommas.preserve,
-            // Silently ignore any unrecognized name.
-            _ => null,
-          };
+        if (formatter case {'trailing_commas': var commas}) {
+          switch (commas) {
+            case 'automate':
+              trailingCommas = TrailingCommas.automate;
+            case 'preserve':
+              trailingCommas = TrailingCommas.preserve;
+            default:
+              stderr.writeln(
+                'Warning: "trailing_commas" option should be "automate" or '
+                '"preserve", but was "$commas".',
+              );
+          }
         }
       }
-    } on PackageResolutionException {
-      // Silently ignore any errors coming from the processing the analyis
-      // options. If there are any issues, we just use the default page width
-      // and keep going.
+    } on PackageResolutionException catch (exception) {
+      // Report the error, but use the default settings and keep going.
+      stderr.writeln(
+        'Warning: Package resolution error when reading '
+        '"analysis_options.yaml" file:\n$exception',
+      );
     }
 
     // Cache whichever options we found (or `null` if we didn't find them).
--- a/lib/src/config_cache.dart
+++ b/lib/src/config_cache.dart
@@ -9,13 +9,14 @@ import 'package:pub_semver/pub_semver.dart';
 
 import 'analysis_options/analysis_options_file.dart';
 import 'analysis_options/io_file_system.dart';
+import 'dart_formatter.dart';
 import 'profile.dart';
 
 /// Caches the nearest surrounding package config file for files in directories.
 ///
 /// The formatter reads `.dart_tool/package_config.json` files in order to
 /// determine the default language version of files in that package and to
-/// resolve "package:" URIs in `analysis_options.yaml` files.
+/// resolve "package:" URIs in "analysis_options.yaml" files.
 ///
 /// Walking the file system to find the package config and then reading it off
 /// disk is very slow. We know that every formatted file in the same directory
@@ -40,13 +41,9 @@ final class ConfigCache {
   /// discovered that there is no surrounding package.
   final Map<String, Version?> _directoryVersions = {};
 
-  /// The previously cached page width for all files immediately within a given
-  /// directory.
-  ///
-  /// The width may be `null` if we formatted a file in that directory and
-  /// discovered that there is no surrounding analysis options that sets a
-  /// page width.
-  final Map<String, int?> _directoryPageWidths = {};
+  /// The previously cached configured options for all files immediately within
+  /// a given directory.
+  final Map<String, _FormatterOptions> _directoryOptions = {};
 
   final IOFileSystem _fileSystem = IOFileSystem();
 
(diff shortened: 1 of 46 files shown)


rrousselGit:
Could the option also be a flag on `dart format`?

This way the IDE could be configured to use `dart format --trailing-commas=preserve` and the CI could not the flag. Cf https://github.com/dart-lang/dart_style/issues/1652#issuecomment-2725901609

munificent:
> Could the option also be a flag on `dart format`?
> 
> This way the IDE could be configured to use `dart format --trailing-commas=preserve` and the CI could not the flag. Cf [#1652 (comment)](https://github.com/dart-lang/dart_style/issues/1652#issuecomment-2725901609)

Yes this PR adds command-line option syntax to enable it as well with the exactly the syntax you suggest here.

lib/src/config_cache.dart
```diff
@@ -89,23 +86,57 @@ final class ConfigCache {
...
+      if (optionsFile['formatter'] case Map<Object?, Object?> formatter) {
+        if (formatter case {'page_width': int width}) {
+          pageWidth = width;
+        }
+
+        if (formatter case {'trailing_commas': String commas}) {
+          trailingCommas = switch (commas) {
+            'automate' => TrailingCommas.automate,
+            'preserve' => TrailingCommas.preserve,
+            // Silently ignore any unrecognized name.
```
natebosch, line 137: Should we consider emitting a message on `stderr` for this case?

lib/src/front_end/piece_factory.dart
```diff
@@ -1604,6 +1625,14 @@ mixin PieceFactory {
     });
   }
 
+  /// Whether there is a trailing comma at the end of the list delimited by
+  /// [rightBracket].
+  bool hasPreservedTrailingComma(Token rightBracket) {
+    if (formatter.trailingCommas != TrailingCommas.preserve) return false;
```
natebosch, line 1631: [nit] Consider returning the boolean expression instead of putting it in a conditional with a literal return.

```dart
    return formatter.trailingCommas == TrailingCommas.preserve &&
        rightBracket.hasCommaBefore;
 ```

File: test/cli/trailing_commas_test.dart [hunk 3]
```diff
@@ -0,0 +1,160 @@
...
+import 'dart:convert';
+
+import 'package:dart_style/src/dart_formatter.dart';
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import '../utils.dart';
+
+void main() {
+  compileFormatter();
```
natebosch, line 14: [optional] Consider using `setUpAll(compileFormatter);` to avoid extra work when the suite is loaded without running any test cases, such as test sharding.

test/cli/trailing_commas_test.dart
```diff
@@ -0,0 +1,160 @@
...
+        '--stdin-name=foo/main.dart',
+      ]);
+
+      process.stdin.writeln(_unformatted);
+      await process.stdin.close();
+
+      // Preserves trailing commas.
+      expect(await process.stdout.next, 'var x = function(');
+      expect(await process.stdout.next, '  argument,');
+      expect(await process.stdout.next, ');');
```
natebosch, line 124: [optional] It only makes a small difference, but some failure scenarios will have better output if you check the stream queue behavior with expectations instead of only check the values in expectations.

```dart
      await expectLater(process.stdout, emits('var x = function('));
      await expectLater(process.stdout, emits('  argument,'));
      await expectLater(process.stdout, emits(');'));
```

or

```dart
      await expectLater(process.stdout, emitsInOrder([
        'var x = function(',
        '  argument,',
        ');',
      ]);
```

natebosch (approved):

lib/src/config_cache.dart
munificent: Honestly, I'm not sure. I'm of two minds about this. In general, the formatter tries to never "fail". It just does the best it can with what you give it. It's historically not been the kind of tool to ask for the user's help to run well. So silently ignoring mistakes in configs is consistent with that.

At the same time, it will likely confuse users if they do have a mistake in the analysis_options.yaml file and the formatter just silently does nothing.

Maybe the right way to thread the needle is to print a warning on stderr but continue anyway. I'll think about that some.

test/cli/trailing_commas_test.dart
munificent: The body of `compileFormatter()` itself calls `setUpAll()`. :)

test/cli/trailing_commas_test.dart
munificent: Done, here and everywhere.

Pingear:
An incredibly stupid solution with automatic addition / removal of commas.
I format the document to make it easy to read, because I have to reread the code often. Some lines can be as long as you like, but some must be short. Manipulating the space in the document helps you navigate the code quickly. The point of formatters is to help me with this.
What genius could have come up with the idea that a monotonous document can somehow be useful to the author of the code?
I had to throw away a day from my life, which I spent looking for a way to fix this.
And most importantly, no one asked for this.
Formatters should format according to the author's wishes, and not add code at their own discretion.
I love the dart language, this is my main language and I am grateful to the dart team for working on it, but this change is very bad, please do not pursue a policy of imposition, if you want to change something, just add an option for customization.

For those whose work has also turned into a constant irritation, here is a short instruction on how to roll back to the old version, without this stupid function.
Go to the directory where flutter is located.
Run the following commands

git checkout 3.27.4
flutter doctor
flutter precache

That is, we simply switch to the old version. I tried to switch to the new one, but something went wrong, perhaps the dart_style 3.1.0-wip version is not yet included in some build, so we will live on the outdated version until a stable version is released with a fix for this stupid decision.
After the new version is released, do not forget to switch to the stable version

git checkout stable
flutter upgrade

Happy coding!

maxfrees:
In which flutter version will this be available, I don't want to automatically remove the comma

quyenlv-unicloud:
@maxfrees Flutter 3.32.0 released with this, you can enjoy now 👍 

`analysis_options.yaml`
```
formatter:
  trailing_commas: preserve
```

PR #1689, opened 2025-03-28
Avoid splitting chains containing only properties.

munificent:
I was trying out the new formatter on a corpus after #1687 and saw a bunch of diffs like:

```dart
// Before:
Widget(
  name:
      some.property.chain,
)

// After:
Widget(
  name: some
      .property
      .chain,
)
```

While the new headline style looks nice for most method call chains, it looks weird if it's only just a series of properties. In particular, a single `foo.bar` looks really silly when split.

This PR bumps the split cost of a chain up if it only contains properties. That leads the solver to prefer splitting the surrounding construct when possible.

### DIFF
--- a/test/tall/invocation/chain_property.stmt
+++ b/test/tall/invocation/chain_property.stmt
@@ -51,3 +51,21 @@ avian
     .feline
     .piscine
     .method();
+>>> Prefer splitting an assignment over splitting a pure property chain.
+variable = avian.bovine.canine.equine.feline;
+<<<
+variable =
+    avian.bovine.canine.equine.feline;
+>>> Don't prefer splitting an assignment if there are methods in the chain.
+variable = avian.bovine.canine.equine.feline();
+<<<
+variable = avian.bovine.canine.equine
+    .feline();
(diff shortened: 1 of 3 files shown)


test/tall/invocation/chain_property.stmt
```diff
@@ -50,4 +50,22 @@ avian
...
     .feline
     .piscine
-    .method();
\ No newline at end of file
+    .method();
+>>> Prefer splitting an assignment over splitting a pure property chain.
+variable = avian.bovine.canine.equine.feline;
+<<<
+variable =
+    avian.bovine.canine.equine.feline;
```
natebosch, line 58: How about a case where the RHS is too long to fit in one line? Will it move the initial selector back to the previous line?

natebosch (approved):

test/tall/invocation/chain_property.stmt
munificent: Oh, good question! Yes, in that case it will split the property chain and put the first selector on the same line as the `=`. Added a test.

PR #1690, opened 2025-04-01
Eagerly split argument lists based on their contents.

### LINKED ISSUE #1660: Split argument lists that would otherwise fit
The most consistent negative feedback we've gotten from the new formatting style is that it often puts an argument list on one line because it fits the page width when the user feels the resulting code would look better if split. For example (based on #1652):

```dart
// Current behavior:
Material(
  color: Colors.transparent,
  child: InkWell(
    customBorder: const CircleBorder(),
    child: Padding(padding: const EdgeInsets.all(2), child: child),
    onTap: () {
      context.read<ScGalleryVm>().crop();
    },
  ),
)

// Preferred:
Material(
  color: Colors.transparent,
  child: InkWell(
    customBorder: const CircleBorder(),
    child: Padding(
      padding: const EdgeInsets.all(2),
      child: child,
    ),
    onTap: () {
      context.read<ScGalleryVm>().crop();
    },
  ),
)
```

Even though the `Padding(...)` call fits on one line, many users feel the latter code is easier to read and maintain.

The new formatter's configurable page width exacerbates this issue. If you set a wider page width, you'll get *even more* of these nested long argument list lines.

I'm filing this issue to be the sort of "meta-issue" that discusses this problem. **If you have examples of argument lists where you think the formatter should have split them and didn't, please include them here (as text, not screenshots!).** Then we can discuss proposed *solutions* to this problem as separate issues:

* #1652
* #1647

munificent:
This is similar to the existing rule where nested non-empty collection literals force the outer one to split. The rules are basically:

* If an argument list contains any named arguments and contains any other calls whose argument lists contain any named arguments (directly or indirectly), then split the outer one. We make an exception where a named argument whose expression is a simple number, Boolean, or null literal doesn't count as a named argument.

* If a list, set, or map literal is the immediate expression in a named argument and one of the above kinds of argument lists (directly or indirectly), then force the collection to split.

```dart
// Before:
Stack(children: [result, Semantics(label: indexLabel)]);

// After:
Stack(
  children: [
    result,
    Semantics(label: indexLabel),
  ],
);
```

I've uploaded a diff of this change applied to a corpus here:

https://github.com/munificent/temp-eager-split-sample/commit/b864a67965c4db736e8a067985446134c5e4d9d5

Fix #1660.

@dart-lang/language-team Since this change is more "heuristic-y" than most formatter changes, I'd appreciate some wider feedback from the team. Take a look at the linked diff and let me know what you think.

### DIFF
--- a/lib/src/front_end/expression_contents.dart
+++ b/lib/src/front_end/expression_contents.dart
@@ -51,19 +51,18 @@ class ExpressionContents {
   /// Begins tracking an argument list.
   void beginCall(List<AstNode> arguments) {
     var type = _Type.otherCall;
+    var hasNontrivialNamedArgument = false;
     for (var argument in arguments) {
       if (argument is NamedExpression) {
         type = _Type.callWithNamedArgument;
         if (!_isTrivial(argument.expression)) {
-          type = _Type.callWithNontrivialNamedArgument;
+          hasNontrivialNamedArgument = true;
           break;
         }
       }
     }
 
-    if (type == _Type.callWithNontrivialNamedArgument) {
-      _stack.last.nontrivialCalls++;
-    }
+    if (hasNontrivialNamedArgument) _stack.last.nontrivialCalls++;
 
     _stack.add(_Contents(type));
   }
@@ -84,8 +83,7 @@ class ExpressionContents {
     // will always be owned by the same call. There may be a named argument at
     // the very beginning of the line owned by the surrounding call which is
     // forced to split.
-    return (contents.type == _Type.callWithNamedArgument ||
-            contents.type == _Type.callWithNontrivialNamedArgument) &&
+    return (contents.type == _Type.callWithNamedArgument) &&
         contents.nontrivialCalls > 0;
   }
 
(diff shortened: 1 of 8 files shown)


natebosch:
I skimmed through the diff and I think it looks really good, all the changes looked like pretty clear improvements to me.

lib/src/front_end/expression_contents.dart
```diff
@@ -0,0 +1,203 @@
...
+  final List<_Contents> _stack = [_Contents(_Type.otherCall)];
+
+  /// Begins tracking an argument list.
+  void beginCall(List<AstNode> arguments) {
+    var type = _Type.otherCall;
+    for (var argument in arguments) {
+      if (argument is NamedExpression) {
+        type = _Type.callWithNamedArgument;
+        if (!_isTrivial(argument.expression)) {
+          type = _Type.callWithNontrivialNamedArgument;
```
natebosch, line 58: [optional] Only this method that assigns `type` makes a differentiation between `callWithNamedArgument` and `callWithNontrivialNamedArgument`.

Consider collapsing those enums to `callWithNamedArgument` and tracking a separate boolean here.

```dart
    var hasNontrivialNamedArgument = false;
    for (var argument in arguments) {
      if (argument is NamedExpression) {
        type = _Type.callWithNamedArgument;
        if (!_isTrivial(argument.expression)) {
          hasNontrivialNamedArgument = true;
          break;
        }
      }
    }

    if (hasNontrivialNamedArgument) {
      _stack.last.nontrivialCalls++;
    }
```

natebosch (approved):

lib/src/front_end/expression_contents.dart
munificent: Good call. I didn't notice that we don't actually need to store that in `_Contents`. Thanks!

PR #1694, opened 2025-04-01
Allow shape to propagate through AssignPieces.

### LINKED ISSUE #1536: Put multiple assignments on the same line
Rather than

```dart
void main() {
  DeprecationProcessingLogger deprecationLogger =
      logger = DeprecationProcessingLogger(
        logger ?? Logger.stderr(),
        silenceDeprecations: {...?silenceDeprecations},
        fatalDeprecations: {...?fatalDeprecations},
        futureDeprecations: {...?futureDeprecations},
        limitRepetition: !verbose,
      );
}
```

I'd prefer

```dart
void main() {
  DeprecationProcessingLogger deprecationLogger = logger =
      DeprecationProcessingLogger(
        logger ?? Logger.stderr(),
        silenceDeprecations: {...?silenceDeprecations},
        fatalDeprecations: {...?fatalDeprecations},
        futureDeprecations: {...?futureDeprecations},
        limitRepetition: !verbose,
      );
}
```

While the AST thinks of the second assignment as an expression, as a user I think of it as an addendum to the LHS, and so I expect it to bind more tightly there than to the value of the assignment.

### LINKED ISSUE #1545: `=>` closures formatting in named argument list
The new dart formatter wants to format this as follows:

```dart
void _showDialog(Widget child, BuildContext context) {
  showCupertinoModalPopup<void>(
    context: context,
    builder:
        (BuildContext context) => Container(
          height: 216,
          padding: const EdgeInsets.only(top: 6.0),
          // The Bottom margin is provided to align the popup above the system
          // navigation bar.
          margin: EdgeInsets.only(
            bottom: MediaQuery.of(context).viewInsets.bottom,
          ),
          // Provide a background color for the popup.
          color: CupertinoColors.systemBackground.resolveFrom(context),
          // Use a SafeArea widget to avoid system overlaps.
          child: SafeArea(top: false, child: child),
        ),
  );
}
```

This gives the closure a lot of prominence over the other parameters and also causes more indentation reducing the effective line length for code inside the closure.

```dart
void _showDialog(Widget child, BuildContext context) {
  showCupertinoModalPopup<void>(
    context: context,
    builder: (BuildContext context) => Container(
      height: 216,
      padding: const EdgeInsets.only(top: 6.0),
      // The Bottom margin is provided to align the popup above the system
      // navigation bar.
      margin: EdgeInsets.only(
        bottom: MediaQuery.of(context).viewInsets.bottom,
      ),
      // Provide a background color for the popup.
      color: CupertinoColors.systemBackground.resolveFrom(context),
      // Use a SafeArea widget to avoid system overlaps.
      child: SafeArea(top: false, child: child),
    ),
  );
}
```

This formatting would visually also be more in line with the formatting I'd expect if builder were a list argument:

```dart
void _showDialog(Widget child, BuildContext context) {
  showCupertinoModalPopup<void>(
    context: context, 
    builder: <Widget>[
      longNameForAWidgetToMakeSureThisDoesNotFit,
      longNameForAWidgetToMakeSureThisDoesNotFit2,
      longNameForAWidgetToMakeSureThisDoesNotFit3,
    ],
  );
}
```

@munificent did mention that there are cases were this kind of formatting for `=>` closures looks worse. I'm interested in seeing examples for that.

### LINKED ISSUE #1668: Unnecessary split between named argument and `=>` function expression
The new formatter has changed the location of added line breaks within a parameter list to immediately after the named parameter token. This has resulted in some significant readability issues compared to the previous formatter. Here are two examples:

#### Before

Image

#### After

Image

#### Before

Image

#### After

Image

### LINKED ISSUE #1679: Inconsistent formatting of shorthand lambda expressions when passed as a named argument
Consider
```dart
void something() => functionCall(
  'Hello world!',
  'Good night city!',
  'Goodbye, world!',
  'Afterwards',
);

class Something {
  void something() => functionCall(
    'Hello world!',
    'Good night city!',
    'Goodbye, world!',
    'Afterwards',
  );
}
void main() {
  final result = switch (value) {
    Something() => functionCall(
      'Hello world!',
      'Good night city!',
      'Goodbye, world!',
    ),
  };

  calling(
    () => functionCall(
      'Hello world!',
      'Good night city!',
      'Goodbye, world!',
      'Afterwards',
    ),
  );

  calling(
    function:
        () => functionCall(
          'Hello world!',
          'Good night city!',
          'Goodbye, world!',
          'Afterwards',
        ),
  );
}
```

The last call to: `calling(function:))` feels inconsistent with the other formatting.

Expected:
```dart
  calling(
    function: () => functionCall(
        'Hello world!',
        'Good night city!',
        'Goodbye, world!',
        'Afterwards',
      ),
  );
```

Rational:
- `function: () => `
- `something() =>`
- `Something() =>`
- `Something.something() =>`
All feel similar as in it's a label that points to an expression to be executed.

munificent:
Consider:

```dart
function(name: (param) => another(argument1, argument2, argument3));
```

Here we have a ListPiece for the `another(...)` call wrapped in an AssignPiece for the `=>` wrapped in an AssignPiece for the `name:`. When the inner ListPiece splits, it will be block-shaped. That in turn allows the surrounding AssignPiece to block format, giving:

```dart
(param) => another(
  argument1,
  argument2,
  argument3,
)
```

(If the ListPiece wasn't block-shaped, then the formatter would split after `=>`.)

An interesting question is then "what is the shape of that AssignPiece?" Prior to this PR, the answer was always "other". Regardless of the shape of the AssignPiece's RHS, a split AssignPiece always had shape other. That meant that the entire original example would end up like:

```dart
function(
  name:
      (param) => another(
        argument1,
        argument2,
        argument3,
      ),
);
```

Because the `(param) => ...` piece had shape "other", the AssignPiece for the named argument couldn't block format it, so it would split after the `:`.

That's not ideal. In practice with `=` and espectially with named arguments, users really want code on the same line as the argument name (and less indented) if at all possible.

This PR propagates the shape of the AssignPiece's contents out to its surrounding context. This means that a headline-shaped RHS causes the assignment to be headline-shaped, and likewise for block-shaped. That lets the above example be formatted as:

```dart
function(
  name: (param) => another(
    argument1,
    argument2,
    argument3,
  ),
);
```

Since `=`, `:`, and `=>` all use AssignPiece, this theoretically means that arbitrarily complex chains of those propagate the shape out. For example, chained assignments like this now work:

```dart
outer = inner = [
  element,
  element,
];
```

In practice, 99% of the time this comes into play, it's a named argument whose expression is a `=>` function. But those are very common for callbacks in Flutter code, so this makes a big difference.

Fix #1536.
Fix #1545.
Fix #1668.
Fix #1679.


test/tall/regression/0500/0500.unit
```diff
@@ -28,19 +28,18 @@
       env.initialize(
         names[i],
         getter: (TopLevelBinding binding, ExprCont ek, ExprCont k) {
-          binding.getter =
-              (TopLevelBinding _, ExprCont ek0, ExprCont k0) => ek0(
-                "Reading static variable '${binding.name}' during its initialization",
-              );
+          binding
+              .getter = (TopLevelBinding _, ExprCont ek0, ExprCont k0) => ek0(
```
natebosch, line 32: This one looks a little strange to me. Is there anything we can do to nudge it towards the formatting on line 36?

natebosch (approved):
Code changes look good. All the sample changes look good except for one edge case with assignment to a field.

I think on the net this is still an improvement if we have to accept the split before the field so we can land it without a fix for that.

test/tall/regression/0500/0500.unit
munificent: This one tends to get funny because the string literals inside the function literal doesn't fit in the page width. So the formatter goes a little wonky trying to cram as much in as it can. I think it's mostly a red herring.

PR #1700, opened 2025-04-03
Make the eager argument list splitting heuristic more conservative.

munificent:
Some folks on the Flutter team pointed out that the previous rule is too aggressive and splits even simple common code like:

```dart
Text('Item 1', style: TextStyle(color: Colors.white));
```

That's probably simple enough to stay on one line. So this tweaks the heuristic to allow that to remain unsplit. Where the old heuristic split any argument list with a named argument that contained a nested named argument, this requires there to be at least *three* named arguments.

It's sort of an arbitrary cutoff, but it seems to do a good job when run on a large corpus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,26 +151,27 @@
   to read spread across multiple lines even if they would otherwise fit on a
   single line (#1660). The rules are basically:
 
-  * If an argument list contains any named arguments and contains any other
-    calls whose argument lists contain any named arguments (directly or
-    indirectly), then split the outer one. We make an exception where a named
-    argument whose expression is a simple number, Boolean, or null literal
-    doesn't count as a named argument.
+  * If an argument list contains at least three named arguments, at least one
+    of which must be directly in the argument list and at least one of which
+    must be nested in an inner argument list, then force the outer one to split.
+    We make an exception where a named argument whose expression is a simple
+    number, Boolean, or null literal doesn't count as a named argument.
 
   * If a list, set, or map literal is the immediate expression in a named
-    argument and one of the above kinds of argument lists (directly or
-    indirectly), then force the collection to split.
+    argument and contains any argument lists with a named argument, then force
+    the collection to split.
 
   ```dart
   // Before:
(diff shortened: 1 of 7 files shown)


munificent:
You can see how this rule compares to the previous more aggressive on in this diff:

https://github.com/munificent/temp_dart_style_3.1.0_diff/commit/a675544aa49260d15ed4cd105fa98c31aec1e230

cc @justinmc and @ piinks

Piinks:
This looks great, thank you @munificent!

natebosch:
> That's probably simple enough to stay on one line.

I don't have any gauge on whether a change like this would push more authors to using the trailing comma config, but IIRC there were examples in the trailing comma threads where they _do_ want to split these cases. My personal preference leans towards the compact version, and I think it's the right choice given the trailing comma config backup option.

natebosch (approved):

munificent:
> I don't have any gauge on whether a change like this would push more authors to using the trailing comma config,

I worry about that a bit too. But I suspect (hope?) that that's a small enough minority of users that this won't move the needle.

I worry equally about angering users who want *denser* code. A somewhat worrisome aspect of eager splitting is that there's no way to opt out. No "dense mode", and no matter how wide you set your page width, if the formatter decides to eagerly split, it will.

Given that, I think it makes sense to err on the side of being a little more conservative with the eager splitting.

> but IIRC there were examples in the trailing comma threads where they _do_ want to split these cases.

True, but there were also cases there where a user wanted something to split that almost no one else would want. I think they are mostly outliers.

> My personal preference leans towards the compact version, and I think it's the right choice given the trailing comma config backup option.

SGTM, thanks!

justinmc (commented):
I'm late but I think this is a big improvement, thanks for following up on this. I even see one example where it fixed the weirdness with trailing dot operators:

Before
```dart
          ).copyWith(
             a: TextStyle(color: LinkUtils.hexToColor("015fff")),
```

After
```dart
         ),
       ).copyWith(a: TextStyle(color: LinkUtils.hexToColor("015fff"))),
```

PR #1703, opened 2025-04-04
Preserve trailing commas for enums with members when trailing_commas are preserved

### LINKED ISSUE #1678: Trailing commas for enhanced enums are not preserved
```dart
enum MyEnum {
  pending('pending'),
  awesome('awesome'),
  completed('completed'),
  ;

  const MyEnum(this.value);

  final String value;
}
```

This gets formatted into, even when trailing commas are preserved:
```dart

enum MyEnum {
  pending('pending'),
  awesome('awesome'),
  completed('completed');

  const MyEnum(this.value);

  final String value;
}
```

jellynoone:
This PR changes the formatter to have it maintain trailing commas of last enum constants when there are members.

So:
```dart
enum E { a, b, c,; int x; }
```

Becomes:
```dart
enum E {
  a,
  b,
  c,
  ;

  int x;
}
```
Instead of:

```dart
enum E {
  a,
  b,
  c;

  int x;
}
```

Fixes #1678 

Additionally, a similar style choice was presented in https://github.com/dart-lang/dart_style/issues/1660#issuecomment-2729244122

**Affected users**:
- Those passing the `trailing_commas: preserve` config option

**Benefits**:
- Maintains developer's intent
- Less churn when adding additional member cases
(Personally, I always add a trailing semicolon to avoid churn and to also make it even clearer this is an advanced enum with members.)

**Downsides**:
- 7 additional logical lines of code to maintain

**Additional considerations**:
I also considered implementing this for `enum E { a, b,; }`, however, that would require more drastic changes to the formatter as far as I can tell.

Additionally, in this case the trailing semicolon provides fewer benefits:
- churn when adding new members is the same if there was a trailing semicolon or not
- giving the enum other members also only amounts to new lines being added rather than changed so again, no real churn.

**Notes**
- One test was changed due to this being a change in behavior.
- This was surprisingly easy and enjoyable to implement.

### DIFF
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -626,14 +626,26 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
           builder.leftBracket(node.leftBracket);
 
           for (var constant in node.constants) {
+            var isLast = constant == node.constants.last;
+            var treatAsLast = isLast;
+            if (isLast && formatter.trailingCommas == TrailingCommas.preserve) {
+              treatAsLast = constant.commaAfter == null;
+            }
             builder.addCommentsBefore(constant.firstNonCommentToken);
             builder.add(
               createEnumConstant(
                 constant,
-                isLastConstant: constant == node.constants.last,
+                isLastConstant: treatAsLast,
                 semicolon: node.semicolon,
               ),
             );
+            // If this the last constant and wasn't treated as last, we need
+            // to append the ending semicolon.
+            if (isLast && !treatAsLast) {
+              if (node.semicolon case var token?) {
+                builder.add(tokenPiece(token));
+              }
+            }
           }
 
           // Insert a blank line between the constants and members.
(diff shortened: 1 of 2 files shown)


jellynoone:
@munificent I saw your recent changes guarding the formatter with a language version flag. Please advise if the changes here require any additional modifications.

FYI, I pulled in the new changes, and all ran successfully.

jellynoone:
@munificent Have you had a change to go over the PR, yet?

jellynoone:
@munificent Is there anything else I should do here?

jellynoone:
@munificent Any plans to review this in the near future? Should I add a changelog entry?

munificent:
Sorry for the long delay. After getting 3.8 out the door, I needed to take a break from working on the formatter and get caught up on other stuff.

I really like this change. Thank you for the PR!

We're moving to a model where style changes are language versioned. I'll take care of handling that. So I'll merge this into a branch instead of main and then apply the language versioning change on top of that before merging into main. I hope that's OK.

I'll also take care of adding this to the CHANGELOG.

jellynoone:
Thank you!

jellynoone:
> We're moving to a model where style changes are language versioned.

I saw this change happening. I considered adding the check but, I figured this was a non-breaking change that it might make sense to always apply since the only people affected are people running `trailing-commas: preserve` and are already explicitly adding a trailing enum comma. (so this seemed more like a bug fix rather than a feature / style change).

PR #1744, opened 2025-07-17
Smarter versioned tests

munificent:
Redo how versioned tests work.

The diff in the README.md and the doc comments in the .dart file changes should explain what's going on here. The basic idea is that now a test is either completely unversioned with only a single output, or completely versioned where all of the output sections have a version number.

I think that's easier to read than a test like:

```
<<<
some.code();
>>> 3.8
some.code();
>>>
some.code();
```

And you have to try to figure out what versions the last section applies to.

This commit also updates the tests to match the new style. This just means adding `3.8` to a bunch of output sections where that was previously implicit. The set of tested versions and their expectations are the same.

I also updated the test updater to support the new output styles.

This change also speeds up running the formatter tests by skipping running tests on language versions where the output is the same as a previous and subsequent version. Prior to this change, running tall_format_test.dart runs 8,674 tests. If you bump the highest supported language version from 3.9 to 3.10 (which I'm about to do for dot shorthands), that jumps to 11,556 because it runs every test at every language version.

With this change, it runs 5,940 tests when the maximum supported version is both 3.9 and 3.10. The number of tests being run only increases when there are actual new tests or new version-specific expectations.


test/tall/statement/switch_legacy.stmt
munificent, line 1: This test was deleted because it doesn't actually make sense.

It was testing how the tall style formatter would format a switch statement at language version 2.19. But the tall style formatter can't be used at any language version older than 3.7, so there's no way for a user to access this.

munificent:
Oops, don't review this just yet. Let me fix the short style test failure.

example/format.dart
```diff
@@ -116,8 +116,15 @@ Future<void> _runTest(
...
-  var expectedText = formatTest.outputs.first.code.textWithSelectionMarkers;
+  String expectedText;
+  switch (formatTest) {
+    case UnversionedFormatTest():
+      expectedText = formatTest.output.code.textWithSelectionMarkers;
+    case VersionedFormatTest():
+      // Pick the newest style for the expectation.
+      expectedText =
+          formatTest.outputs.entries.last.value.code.textWithSelectionMarkers;
+  }
```
natebosch, line 127: Can this be a switch expression?
```suggestion
  var expectedText = switch (formatTest) {
    UnversionedFormatTest() => formatTest.output.code.textWithSelectionMarkers,
    // Pick the newest style for the expectation.
    VersionedFormatTest() => formatTest.outputs.entries.last.value.code.textWithSelectionMarkers
  };
```

lib/src/testing/test_file.dart
```diff
@@ -301,20 +341,45 @@ final class FormatTest {
...
+
+/// A test whose expected formatting changes at specific versions.
+final class VersionedFormatTest extends FormatTest {
+  /// The expected output by version.
+  ///
+  /// Each key is the lowest version where that output is expected. If there are
+  /// supported versions lower than the lowest key here, then the test is not
+  /// run on those versions at all. These tests represent new syntax that isn't
+  /// supported in later versions. For example, if the map has only a single
+  /// entry whose key is 3.8, then the test is skipped on 3.8, run at 3.8, and
```
natebosch, line 361: Skipped on 3.7?

test/utils.dart
```diff
@@ -146,50 +146,123 @@ Future<void> testBenchmarks({required bool useTallStyle}) async {
...
+            // This output is unversioned, so test at the lowest and highest
+            // supported versions. If it formats the same on those, we assume
+            // it does for all of the intermediate versions too.
+            testedVersions = {
+              _testedTallVersions.first: output,
+              _testedTallVersions.last: output,
+            };
+          case VersionedFormatTest(:var outputs):
+            testedVersions = _versionedTestEntries(outputs);
+        }
```
natebosch, line 162: Can this be a switch expression?

test/utils.dart
```diff
@@ -146,50 +146,123 @@ Future<void> testBenchmarks({required bool useTallStyle}) async {
...
+
+        testedVersions.forEach((version, output) {
```
natebosch, line 164: [optional] Should we consider holding a preference for `for (MapEntry(key: final version, value: final output) in testedVersions.entries)` to match our guidelines around `Iterable.forEach(<functionLiteral>)`?

natebosch (approved):

test/utils.dart
munificent: Good call. Did it for both switches here.

test/utils.dart
munificent: If we do https://github.com/dart-lang/language/issues/2563, then yes. But right now, it's pretty verbose to use destructuring here. Or maybe if we put an extension on Map that returns and Iterable of `(K, V)` so you don't have to write `MapEntry`, `key:`, and `value:`.

lib/src/testing/test_file.dart
munificent: Heh, oops!

PR #1759, opened 2025-08-08
Synchronize the dart_style version in its own help output.

munificent:
The pubspec is already bumped to 3.1.3-wip.


munificent:
Interesting! Probably not worth it. I do want to automate this a little better. There is a grinder script, but it doesn't quite fit the current workflow where we roll into the Dart SDK before publishing. It needs some tweaks.

PR #1776, opened 2025-10-11
Add exclude patterns for files and directories

### LINKED ISSUE #864: Add ability to exclude files
I'm using `dartfmt` via `flutter format`, which I run it like this:
```
flutter format .
```
This formats all `.dart` files recursively, which is almost exactly what I want. Unfortunately, it also formats `.g.dart` files, so my generated files bounce between their generated state and a formatted version thereof.

It would be very useful if `dartfmt` could exclude files matching certain patterns. Something like:

```
dartfmt -e *.g.dart -e *.foo.dart .
```

(surfacing this via `flutter format` would of course be a separate issue)

ab22593k:
Add the `--exclude` option to the format command.

This option allows users to exclude specific files and directories from being formatted. The option accepts a list of patterns that can be file paths, directory names, or glob patterns.

- Added `excludePatterns` to `FormatterOptions`.
- Modified `FormatCommand` to include `--exclude` option in `argParser`.
- Updated `formatPaths` to check and exclude files based on provided patterns.
- Added tests to verify the exclude functionality.

Fixes: #864.

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/lib/src/cli/format_command.dart
+++ b/lib/src/cli/format_command.dart
@@ -154,6 +154,13 @@ final class FormatCommand extends Command<int> {
           'See dart.dev/go/experiments.',
       hide: !verbose,
     );
+    argParser.addMultiOption(
+      'exclude',
+      help:
+          'Exclude files and directories matching the given patterns.\n'
+          'Patterns can be file paths, directory names, or glob patterns.',
+      hide: !verbose,
+    );
 
     if (verbose) argParser.addSeparator('Options when formatting from stdin:');
 
@@ -297,6 +304,7 @@ final class FormatCommand extends Command<int> {
     var setExitIfChanged = argResults.flag('set-exit-if-changed');
 
     var experimentFlags = argResults.multiOption('enable-experiment');
+    var excludePatterns = argResults.multiOption('exclude');
 
     // If stdin isn't connected to a pipe, then the user is not passing
     // anything to stdin, so let them know they made a mistake.
@@ -324,6 +332,7 @@ final class FormatCommand extends Command<int> {
       summary: summary,
       setExitIfChanged: setExitIfChanged,
       experimentFlags: experimentFlags,
+      excludePatterns: excludePatterns,
     );
 
     if (argResults.rest.isEmpty) {
--- a/lib/src/cli/formatter_options.dart
+++ b/lib/src/cli/formatter_options.dart
@@ -54,6 +54,9 @@ final class FormatterOptions {
   /// See dart.dev/go/experiments for details.
   final List<String> experimentFlags;
 
+  /// Patterns for files and directories to exclude from formatting.
+  final List<String> excludePatterns;
+
   FormatterOptions({
     this.languageVersion,
     this.indent = 0,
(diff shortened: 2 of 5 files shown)


brunonahornygaudium (approved):

bruno-m-antunes-alb (approved):
LGTM

brunonahornygaudium (approved):

lib/src/cli/format_command.dart
```diff
@@ -296,7 +303,8 @@ final class FormatCommand extends Command<int> {
     var followLinks = argResults.flag('follow-links');
     var setExitIfChanged = argResults.flag('set-exit-if-changed');
 
-    var experimentFlags = argResults.multiOption('enable-experiment');
+    var experimentFlags = argResults['enable-experiment'] as List<String>;
+    var excludePatterns = argResults['exclude'] as List<String>;
```
munificent: You'll want to use `.multiOption()` here instead of the older deprecated `[]` API.

munificent (commented):
The code and tests look great. Thank you for the contribution!

Most users don't run `dart format` on the command line. Or, if they do, it's not the *only* way they run it. They will also often run it from their IDE where they don't easily control the command line arguments being passed.

Also, projects are often worked on by multiple users and they all need to use the same exclusions or they'll risk accidentally formatting the wrong files.

That suggests that file exclusions should canonically be stored in [the analysis_options.yaml](https://github.com/dart-lang/dart_style/wiki/Configuration#analysis-options) and read from there. Most IDEs invoking the formatter do so through its library API. That API doesn't let you pass in the entire `analysis_options.yaml` file text for the formatter to parse. Instead, every configuration option has to be parsed by the host and [passed separately](https://github.com/dart-lang/dart_style/blob/main/lib/src/dart_formatter.dart#L92-L95). This is not a great library API design, but I haven't had the time to change it since it is a breaking API change and requires coordinating with the IDE teams to migrate to a better API.

That's the main reason we haven't already implemented this feature. We could support passing exclusions *only* on the command line, but it will make the overall user experience worse: Everything works right when you invoke `dart format` directly and remember to pass in the exclusions, but the next time you hit Ctrl-S in your editor it may accidentally format a file you didn't intend.

Of course, you didn't have this context, but it puts this PR in a weird place. I don't want to put users in the position where they start relying on a feature that breaks their IDE experience or makes things harder for others on their project. But I also don't know when we'll have the time to make the API changes to make it easier to plumb new configuration like this from the IDEs to the formatter through its library API.

dawid-niedzwiecki:
Would love this for CI checks, excluding formatted files from `dart format --set-exit-if-changed .`. Especially with `analysis_options.yaml` support!

luciemdx:
Friendly ping, this would indeed be a great companion to --set-exit-if-changed on CI.

munificent:
Sorry for the long delay. (I was out on medical leave.) Overall, yes, I would like to add support for file exclusions. To do so, as I said earlier, I think we should store them in a config file, probably analysis_options.yaml. Then the various IDEs need to plumb that configuration in when they invoke the formatter.

I don't want to simply add another parameter to the `DartFormatter()` constructor for exclusions and require every IDE implementation to parse it out of the config file themselves and pass it along. If we do that, then before too long, there will be some *other* piece of configuration that we want the formatter to support and again every IDE will have to parse it and plumb it through. I want to a better API that lets IDEs get out of the way and lets the formatter find that itself.

But this is a big enough change and requires enough coordination with other teams that I really need to write a design doc and discuss it with the larger team to decide what the best approach is. I'll try to get that process started soon. I'm sorry that it's frustrating that I can't simply land this (quite nice) PR, but I'm trying to make sure that the whole Dart user experience is the best it can be in all the ways the various tools interact.

PR #1781, opened 2025-10-30
Add script to remove experiments from tests when they have launched.

lrhn:
Re: https://github.com/dart-lang/dart_style/pull/1779#issuecomment-3460959291

I made a script to remove experiments from tests, and update the min-SDK version.

I haven't figured out how to find the matching analyzer package and update that in pubspec as well.
May need to run `dart pub upgrade` after running this script, or manually updating the analyzer package
to a version that supports the new language version without experiment flags.

### DIFF
--- a/lib/src/testing/test_file.dart
+++ b/lib/src/testing/test_file.dart
@@ -16,7 +16,6 @@ final _experimentPattern = RegExp(r'\(experiment ([a-z-]+)\)');
 final _preserveTrailingCommasPattern = r'(trailing_commas preserve)';
 // Only supports two-digit hex numbers.
 final _unicodeUnescapePattern = RegExp(r'×([0-9a-fA-F]{2,4})');
-final _unicodeEscapePattern = RegExp('[\x0a\x0c\x0d\x80-\uFFFF]');
 
 /// Matches an output header line with an optional version and description.
 /// Examples:
@@ -133,13 +132,13 @@ final class TestFile {
       while (i < lines.length && !lines[i].startsWith('<<<')) {
         inputBuffer.writeln(readLine());
       }
-
-      var inputCode = _extractSelection(
-        _unescapeUnicode(inputBuffer.toString()),
+      var inputSource = inputBuffer.toString();
+      var inputCode = _parseTestSource(
+        inputSource,
         isCompilationUnit: isCompilationUnit,
       );
 
-      var input = TestEntry(description, inputComments, inputCode);
+      var input = TestEntry(description, inputComments, inputCode, inputSource);
 
       // Read the outputs. A single test should have outputs in one of two
       // forms:
@@ -191,8 +190,8 @@ final class TestFile {
           assert(outputText.endsWith('\n'));
           outputText = outputText.substring(0, outputText.length - 1);
         }
-        var outputCode = _extractSelection(
-          _unescapeUnicode(outputText),
+        var outputCode = _parseTestSource(
+          outputText,
           isCompilationUnit: isCompilationUnit,
         );
 
@@ -200,6 +199,7 @@ final class TestFile {
           outputDescription.trim(),
           outputComments,
           outputCode,
+          outputText,
         );
         if (outputVersion != null) {
           if (versionedOutputs.containsKey(outputVersion)) {
@@ -431,7 +431,12 @@ final class TestEntry {
 
   final SourceCode code;
 
-  TestEntry(this.description, this.comments, this.code);
+  /// Test file source text, including Unicode escapes and selection markers.
+  ///
+  /// Used when writing the test using [writeTo].
+  final String source;
+
+  TestEntry(this.description, this.comments, this.code, this.source);
 
   /// Writes description, comments and code.
   ///
@@ -447,7 +452,7 @@ final class TestEntry {
     for (var comment in comments) {
       output.writeln(comment);
     }
-    output.write(escapeUnicode(code.textWithSelectionMarkers));
+    output.write(source);
     if (!code.text.endsWith('\n')) output.writeln(); // Can that happen?
   }
 }
@@ -510,21 +515,42 @@ extension SourceCodeExtensions on SourceCode {
   }
 }
 
+/// Converts test source code to a [SourceCode].
+///
+/// Unescapes Unicode escapes (`×HH`, `×HHHH`) and extracts a selection from a
+/// `‹...›` range.
+SourceCode _parseTestSource(
+  String sourceText, {
+  bool isCompilationUnit = false,
+}) {
+  sourceText = _unescapeUnicode(sourceText);
+  return _extractSelection(sourceText, isCompilationUnit: isCompilationUnit);
+}
+
 /// Given a source string that contains ‹ and › to indicate a selection, returns
 /// a [SourceCode] with the text (with the selection markers removed) and the
 /// correct selection range.
+/// Only recognizes the first `‹...›` range.
 SourceCode _extractSelection(String source, {bool isCompilationUnit = false}) {
+  int? selectionStart;
+  int? selectionLength;
   var start = source.indexOf('‹');
-  source = source.replaceAll('‹', '');
-
-  var end = source.indexOf('›');
-  source = source.replaceAll('›', '');
-
+  if (start >= 0) {
+    var end = source.indexOf('›', start + 1);
+    if (end >= 0) {
+      source =
+          '${source.substring(0, start)}'
+          '${source.substring(start + 1, end)}'
+          '${source.substring(end + 1)}';
+      selectionStart = start;
+      selectionLength = end - start - 1;
+    }
+  }
   return SourceCode(
     source,
     isCompilationUnit: isCompilationUnit,
-    selectionStart: start == -1 ? null : start,
-    selectionLength: end == -1 ? null : end - start,
+    selectionStart: selectionStart,
+    selectionLength: selectionLength,
   );
 }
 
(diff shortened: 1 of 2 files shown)


lrhn:
(Will format and remove unused members on Monday.)

tool/update_sdk.dart
```diff
@@ -0,0 +1,628 @@
...
+  // General use of verbosity levels:
+  // 0: No extra output.
+  // 1+: Say what is done.
+  // 2+: Also say when no change is made (what is *not* done).
+  var verbose = 0;
+  var dryRun = false;
+  var printHelp = false;
+  File? experimentsFile;
+  Version? targetVersion;
+  for (var arg in args) {
```
munificent: This might be simpler using the args package (which dart_style already depends on), but it's up to you.

tool/update_sdk.dart
```diff
@@ -0,0 +1,628 @@
...
+         If provided, use that as SDK version in pubspec.yaml.
+         If not provided, uses most recent feature release version.
+PATH     Path to "experimental_features.yaml" or to an SDK repository containing
+         that file in "tools".
+         Will be searched for if not provided.
+
+-v       Verbosity. Can be used multiple times.
+-n       Dryrun. If set, does not write changed files back.
+-h       Show usage.
+''';
```
munificent: The args package will give you usage help text for free.

munificent (changes requested):
This is really cool! But I think it also re-invents some stuff that would be easier to reuse: arg parsing, test file parsing, and version comparison.

tool/update_sdk.dart
```diff
@@ -0,0 +1,628 @@
...
+      }
+      return output.join('\n');
+    }
+    return null;
+  });
+}
+
+// --------------------------------------------------------------------
+// Parse the `experimental_features.yaml` file to find experiments
+// and their 'enabled' version.
```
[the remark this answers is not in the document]
lrhn: All section headers are now above commented functions.

lib/src/testing/test_file.dart
```diff
@@ -336,6 +364,13 @@ sealed class FormatTest {
     if (input.description.isEmpty) return 'line $line';
     return 'line $line: ${input.description}';
   }
+
+  void writeTo(StringSink output) {
```
munificent: Nit: how about `sink` for the parameter here and in the subclasses, so it doesn't collide with the `output` instance variable?

lib/src/testing/test_file.dart
```diff
@@ -12,18 +13,19 @@ import '../source_code.dart';
...
 
 /// Matches an output header line with an optional version and description.
 /// Examples:
-///
-///    >>>
-///    >>> Only description.
-///    >>> 1.2
-///    >>> 1.2 Version and description.
-final _outputPattern = RegExp(r'<<<( (\d+)\.(\d+))?(.*)');
+/// ```plaintext
```
munificent: Style nit: @natebosch and I went back and forth a bit on how to style code in doc comments and eventually settled on the old Markdown indented style, so to be consistent, this should keep the old Markdown style.

lib/src/testing/test_file.dart
```diff
@@ -376,7 +431,30 @@ final class TestEntry {
...
+        ..write(' ')
+        ..writeln(description);
+    } else {
+      output.writeln();
+    }
+    for (var comment in comments) {
+      output.writeln(comment);
+    }
+    output.write(source);
+    if (!code.text.endsWith('\n')) output.writeln(); // Can that happen?
```
munificent: Nit: Make this a `TODO(rnystrom): ` comment.

(I don't know the answer either. 😬 )

lib/src/testing/test_file.dart
```diff
@@ -405,21 +515,42 @@ extension SourceCodeExtensions on SourceCode {
...
+  String sourceText, {
+  bool isCompilationUnit = false,
+}) {
+  sourceText = _unescapeUnicode(sourceText);
+  return _extractSelection(sourceText, isCompilationUnit: isCompilationUnit);
+}
+
 /// Given a source string that contains ‹ and › to indicate a selection, returns
 /// a [SourceCode] with the text (with the selection markers removed) and the
 /// correct selection range.
```
munificent, line 536: Tiny nit: Either put the next sentence on the same paragraph as the previous one, or insert a blank `///`  line to make it a separate paragraph.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+//   1+: Say what is done.
+//   2+: Also say when no change is made (what is *not* done).
+
+void main(List<String> args) {
+  try {
+    var (
+      int verbose,
+      bool dryRun,
+      File? experimentsFile,
+      Version? targetVersion,
```
munificent: No need for type annotations on the destructured fields since they'll be inferred from the return typre of `_parseArguments()`.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+_parseArguments(List<String> arguments) {
+  var argsParser = args.ArgParser()
+    ..addFlag(
+      'verbose',
+      abbr: 'v',
+      help: 'More verbose information during processing',
+      negatable: false,
+    )
+    ..addFlag(
+      'dryrun',
```
munificent, line 123: `dry-run`, here and below. (It's consistent with our other tools like `dart fix` which use `dry-run`.)

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+    // Only print usage if requested if no error.
+    stdout.writeln(usage);
+    stdout.writeln(argsParser.usage);
+    exit(0);
+  }
+  return (verbose, dryRun, experimentsFile, targetVersion);
+}
+
+class Updater {
+  // The `dart_style` package root.
```
munificent, line 185: `///` here and on other doc comments below.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+  final Directory root;
+
+  /// The target version that the `dart_style` package should use and support.
+  ///
+  /// Any experiment launched or discontinued before or in this version
+  /// should not be used as experiment flags.
+  final Version targetVersion;
+
+  // Verbosity level.
+  final int verbose;
```
munificent, line 195: It's OK, but it does feel a little weird for verbosity to be an `int` with only a couple of special blessed values. Ideally it would be an enum, but it's probably not worth it for a little tool script.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+    return buffer.toString();
+  });
+}
+
+// --------------------------------------------------------------------
+// Handle the experiments file.
+
+/// Parses the `experimental_features.yaml` file.
+///
+/// Finds the current version, all experiments their 'enabled' version.
```
munificent, line 429: I'm not sure what this sentence is trying to say.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+  if (currentVersion != null &&
+      targetVersion != null &&
+      targetVersion < targetVersion) {
+    throw UnsupportedError(
+      'Target version higher than actual version:'
+      ' $targetVersion > $targetVersion',
+    );
+  }
+  var features = yaml['features'] as y.YamlMap;
+  for (var MapEntry(key: name as String, value: info as y.YamlMap)
```
munificent, line 458: Nit: I would move the declaration of `result` down here.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+    String content;
+    if (existingContent != null) {
+      content = existingContent;
+    } else if (_cache.containsKey(path)) {
+      throw ConcurrentModificationError(path);
+    } else {
+      content = path.readAsStringSync();
+      _cache[path] = null;
+    }
+    var change = false;
```
munificent, line 523: Nit: `changed`.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+  // Modified file state.
+  final FileEditor files;
+
+  Updater(
+    this.root,
+    this.targetVersion,
+    this.experiments, {
+    this.verbose = 0,
+    this.dryRun = false,
+  }) : files = FileEditor(verbose: verbose - 1, dryRun: dryRun);
```
munificent, line 213: What's the `- 1` about here?

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+  // Check current directory. Script is run in the package root dir.
+  var cwd = Directory.current;
+  if (_isStylePackageDir(cwd)) return cwd;
+  // Check parent directory of script file,
+  // if run as `dart <pkgRoot>/tool/update_sdk.dart`.
+  var scriptDir = p.dirname(p.absolute(p.fromUri(Platform.script)));
+  var scriptParentDir = Directory(p.dirname(scriptDir));
+  if (_isStylePackageDir(scriptParentDir)) return scriptParentDir;
+  // Check ancestor directories of current directory,
+  // if run from, fx, inside `<pkgRoot>/tool/`.
```
munificent: `fx` -> `for example`.

tool/update_sdk.dart
```diff
@@ -0,0 +1,743 @@
...
+         If not provided, uses most recent feature release version.
+PATH     Path to "experimental_features.yaml" or to an SDK repository containing
+         that file in "tools/".
+         Will be searched for if not provided.
+''';
+
+/// Caught by [main] to flush output before actually exiting.
+Never exit(int value) {
+  // ignore: only_throw_errors
+  throw (exitCode: value);
```
munificent: Instead of ignoring, I'd rather you make a little `AbortException` class and throw that (and then catch it instead of doing a catch-all clause).

munificent (approved):
A bunch of small things here and there, most nits, then LGTM. Thanks for doing this!

lib/src/testing/test_file.dart
lrhn: The instance member in the subclass of `FormatTest` (not of `TestFile` itself) ...
Can rename, but "sink" is not a good name (it's just repeating the type, not saying what the parameter means).
I used "target" instead.

lib/src/testing/test_file.dart
lrhn: That's ... non-standard. It reduces the available line length, does not allow an info string like a code fence, and requires you to visually recognize whether the text is indented by 3 or 4 spaces.
Indented code blocks are strictly inferior to code fences.

This text is not Dart. Without a code fence tag, I can't tell that to DartDoc.

I'll make it consistent, but do consider switching the entire code to code fences instead.

lib/src/testing/test_file.dart
lrhn: Added new short first line, Chuck Norris style!
And more rewriting.

tool/update_sdk.dart
lrhn: Not what I would do.

"is there change? Yes!"
"did change occur? Yes!"
"Any change? YES!"

The word "changed" feels transitive. I'd ask what changed.

Might as well be precise and say what it actually means, so `wroteOutput`.

tool/update_sdk.dart
lrhn: I use `verbose` to decide the level of detail.
If I want a very verbose level of details for the script itself, I (heuristically) don't want that much information from a dependency.

I probably have a mapping from number to meaning in my head.
Something like:
0: Just do it, and just give me a summary. ("Done: N files changed.")
1: Tell me what you do. ("This file changed.")
2: Also tell me what you didn't do. ("This file was looked at, but didn't change.")
3: Trace everything. ("I'll now look at this file." ... "And now I'm done.")

When saying what I'm doing and not doing, and I'm using a separate feature as part of that, I treat that separate feature as a single operation, and just want it to say what it does. If it does nothing, I'll say that afterwards.

So, heuristically, `verbose - 1` for sub-components.

(It's all there for debugging anyway.)

PR #1787, opened 2025-11-14
Remove uses of internal analyzer implementation.

munificent:
In the very rare case that a user called formatStatement() and gave the formatter something longer than a single statement, we need to throw a FormatException but don't actually have a diagnostic from the analyzer to use.

To deal with that, it used to import some internal implementation libraries in order to conjure up a diagnostic. Instead, this has a local implementation of the necessary public analyzer classes.

This will avoid dart_style being broken with analyzer 9.0.0 when that internal code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.1.4-wip
+
+* Remove dependencies on analyzer internal implementation.
+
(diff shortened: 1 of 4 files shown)


stereotype441 (approved):
lgtm! Thanks, Bob!

PR #1795, opened 2026-01-08
Require analyzer 10.0.0, stop using deprecated nodes and properties.

scheglov:
This does not add full support for the primary constructors feature, only migrates to keep existing code working.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -1437,8 +1437,8 @@ mixin PieceFactory {
   /// If the type is an extension, then [onType] is a record containing the
   /// `on` keyword and the on type.
   ///
-  /// If the type is an extension type, then [representation] is the primary
-  /// constructor for it.
+  /// If the type has a primary constructor, e.g. an extension type, then
+  /// [primaryConstructor] is not `null`.
   void writeType(
     NodeList<Annotation> metadata,
     List<Token?> keywords, {
@@ -1446,7 +1446,7 @@ mixin PieceFactory {
     TypeParameterList? typeParameters,
     Token? equals,
     NamedType? superclass,
-    PrimaryConstructorDeclaration? extensionTypePrimaryConstructor,
+    PrimaryConstructorDeclaration? primaryConstructor,
     ExtendsClause? extendsClause,
     MixinOnClause? onClause,
     WithClause? withClause,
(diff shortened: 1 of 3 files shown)


pubspec.yaml
kevmoo: Might could just go w/ the release?

scheglov: I'm mostly looking to pull this into Dart SDK and google3.
To be able to remove deprecated APIs from the analyzer.
Given that this PR does not add full primary constructors support, publishing it might be less interesting.

FMorschel: It is interesting, so plugin authors can use the new `analyzer` version. Until this is published, the `analyzer_plugin`, which depends on the new `analyzer` version, can't ever be resolved. See https://github.com/dart-lang/dart_style/issues/1796

lib/src/front_end/piece_factory.dart
```diff
@@ -1441,12 +1441,12 @@ mixin PieceFactory {
...
     NodeList<Annotation> metadata,
-    List<Token?> keywords,
-    Token? name, {
+    List<Token?> keywords, {
+    Token? name,
     TypeParameterList? typeParameters,
     Token? equals,
     NamedType? superclass,
-    RepresentationDeclaration? representation,
+    PrimaryConstructorDeclaration? extensionTypePrimaryConstructor,
```
munificent: This will eventually be used for non-extension type primary constructors too, so I'd just call it `primaryConstructor`.

lib/src/short/source_visitor.dart
```diff
@@ -2365,6 +2372,40 @@ final class SourceVisitor extends ThrowingAstVisitor {
...
+    visit(node.constructorName);
+    visit(node.formalParameters);
+  }
+
+  @override
+  void visitPrimaryConstructorBody(PrimaryConstructorBody node) {
+    visitMetadata(node.metadata);
+    token(node.thisKeyword);
+
+    if (node.initializers.isNotEmpty) {
```
munificent: Can you add a TODO here to unify this with `visitConstructorDeclaration()`? The formatting for initializer lists should be the same for both, so I'll refactor it to use the same code at some point.

kevmoo:
Are we good to publish this? @munificent ?

munificent:
Yeah, that makes sense. It is annoying to have to send out a second PR to get reviewed that doesn't do anything except remove "-wip".

PR #1818, opened 2026-03-28
Fix IOFileSystemPath printing during package resolution exceptions

kevmoo:
The `IOFileSystemPath` object lacked a `toString()` method, leading to `Instance of 'IOFileSystemPath'` being rendered in error strings. When `PackageResolutionException` was thrown encountering package resolution errors while reading analysis options, the path output was masked.

This change adds `toString()` to `IOFileSystemPath` to output the correct underlying file system `.path`. A test is also added to `io_file_system_test.dart` preventing regressions.

### DIFF
--- a/lib/src/analysis_options/io_file_system.dart
+++ b/lib/src/analysis_options/io_file_system.dart
@@ -48,4 +48,7 @@ final class IOFileSystemPath implements FileSystemPath {
   final String path;
 
   IOFileSystemPath._(this.path);
+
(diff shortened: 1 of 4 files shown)


kevmoo:
This is a follow-up fix to https://github.com/dart-lang/dart_style/commit/0d56d8550440f19bb2a67b691859a22304486e02

kallentu (approved):

PR #1821, opened 2026-04-09
Require analyzer ^13.0.0, with breaking changes.

scheglov:
This change refers to not yet published `analyzer 13.0.0`, because I cannot land corresponding [analyzer CL](https://dart-review.googlesource.com/c/sdk/+/488624) (and then publish) without pulling into the Dart SDK a version of `dart_style` (and `test`, see https://github.com/dart-lang/test/pull/2622) that is compatible with this breaking change version of the analyzer. So, I want to land this PR with a version of `dart_style` that is compatible with `analyzer 13.0.0` (to be), and the pull this git commit into [the CL](https://dart-review.googlesource.com/c/sdk/+/488624).

I tested it locally with
```yaml
dependency_overrides:
  _fe_analyzer_shared:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/_fe_analyzer_shared
  analyzer:
    path: /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/analyzer
```
Obviously, GitHub bots will be red until `analyzer 13.0.0` is published.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -1475,6 +1475,10 @@ mixin PieceFactory {
   }
 
   /// Writes an assignment-like construct with a token on the left-hand side.
+  ///
+  /// This is used instead of [writeAssignment] when the left-hand side is a
+  /// [Token] (like the name of a named argument or record field) rather than
+  /// a full [AstNode].
   void writeTokenAssignment(
     Token leftHandSide,
     Token operator,
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -127,7 +127,7 @@ mixin PieceFactory {
   /// Writes a [ListPiece] for an argument list.
   void writeArguments(
     Token leftBracket,
-    List<Expression> arguments,
+    List<Argument> arguments,
     Token rightBracket,
   ) {
     // In 3.7, we don't support preserving trailing commas or eager splitting.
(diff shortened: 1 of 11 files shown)


scheglov:
Indication that this implementation passes tests.
image

lib/src/front_end/piece_factory.dart
```diff
@@ -1482,6 +1474,58 @@ mixin PieceFactory {
     pieces.add(AssignPiece(leftPiece, rightPiece));
   }
 
+  /// Writes an assignment-like construct with a token on the left-hand side.
+  void writeTokenAssignment(
```
kallentu, line 1482: Could you explain a bit further on why we couldn't use `writeAssignment`? This and that helper look really similar.

kallentu:
Also, I'll leave final approval to Bob now that he's back.

lib/src/front_end/piece_factory.dart
scheglov: I have added a doc comment to writeTokenAssignment explaining that it is used when the left-hand side is a Token (like the name of a named argument or record field) rather than a full AstNode.

scheglov: Done.

scheglov:
ping

lib/src/short/argument_list_visitor.dart
```diff
@@ -546,18 +546,16 @@ final class ArgumentSublist {
...
+    var named = <Argument>[];
     var inNamed = false;
     for (var argument in arguments) {
-      if (argument is NamedExpression) {
+      if (argument is NamedArgument) {
         inNamed = true;
       } else if (inNamed) {
         // Got a positional argument after a named one.
-        return [[], arguments];
+        return [<Argument>[], arguments];
```
munificent: This type argument shouldn't be needed. It should get filled in by inference from the function's return type.

munificent (approved):
LGTM! I really like the new AST structure. I'm happy to see `DefaultFormalParameter` gone. :)

lib/src/short/argument_list_visitor.dart
scheglov: Done.

PR #1823, opened 2026-04-10
Support primary constructors in short style.

kallentu:
A few SDK tests (`pkg/front_end/test/textual_outline/dartdevc/primary_constructors/abstract_class`) are crashing so I'm updating short style to handle primary constructors.

- Formatting of `;` bodies in classes, mixins, extensions, extension types, and enums.
- Refactored some of the visitors. `visitBlockEnumBody` was extracted from the enum declaration visitor. Added `NameWithTypeParameters` so we can visit the `namePart` for primary constructors and regular declarations.

### DIFF
--- a/lib/src/short/source_visitor.dart
+++ b/lib/src/short/source_visitor.dart
@@ -599,13 +599,7 @@ final class SourceVisitor extends ThrowingAstVisitor {
     modifier(node.mixinKeyword);
     token(node.classKeyword);
     space();
-    token(node.namePart.typeName);
-    visit(node.namePart.typeParameters);
-
-    var namePart = node.namePart;
-    if (namePart is PrimaryConstructorDeclaration) {
-      visit(namePart.formalParameters);
-    }
+    visit(node.namePart);
 
     visit(node.extendsClause);
     _visitClauses(node.withClause, node.implementsClause);
@@ -1024,8 +1018,7 @@ final class SourceVisitor extends ThrowingAstVisitor {
     builder.nestExpression();
     token(node.enumKeyword);
     space();
-    token(node.namePart.typeName);
-    visit(node.namePart.typeParameters);
+    visit(node.namePart);
     _visitClauses(node.withClause, node.implementsClause);
     space();
 
@@ -2384,6 +2377,12 @@ final class SourceVisitor extends ThrowingAstVisitor {
     visit(node.typeParameters);
     visit(node.constructorName);
     visit(node.formalParameters);
+  }
+
+  @override
+  void visitNameWithTypeParameters(NameWithTypeParameters node) {
+    token(node.typeName);
+    visit(node.typeParameters);
   }
 
   @override
--- a/lib/src/short/source_visitor.dart
+++ b/lib/src/short/source_visitor.dart
@@ -601,15 +601,24 @@ final class SourceVisitor extends ThrowingAstVisitor {
     space();
     token(node.namePart.typeName);
     visit(node.namePart.typeParameters);
+
+    var namePart = node.namePart;
+    if (namePart is PrimaryConstructorDeclaration) {
+      visit(namePart.formalParameters);
+    }
+
     visit(node.extendsClause);
     _visitClauses(node.withClause, node.implementsClause);
     visit(node.nativeClause, before: space);
-    space();
+    if (node.body is! EmptyClassBody) space();
 
     builder.unnest();
-    // TODO(scheglov): support for EmptyBody
-    var body = node.body as BlockClassBody;
-    _visitBody(body.leftBracket, body.members, body.rightBracket);
+    switch (node.body) {
+      case BlockClassBody body:
+        _visitBody(body.leftBracket, body.members, body.rightBracket);
+      case EmptyClassBody body:
+        token(body.semicolon);
+    }
   }
 
   @override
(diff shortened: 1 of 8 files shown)


natebosch:
Can you elaborate on what's failing? Do I understand correctly that there is a situation where we are trying to format code that uses new language features, but is configuraed at an old language version? I'm not sure it makes sense to try to format new syntax with old formatting rules, is it feasible for us to report analyzer errors instead of crashing?

kallentu:
```
/====================================================================================================================\
| pkg/front_end/test/textual_outline/dart2js/primary_constructors/abstract_class broke (Pass -> Fail, expected Pass) |
\====================================================================================================================/



type 'EmptyClassBodyImpl' is not a subtype of type 'BlockClassBody' in type cast

#0      SourceVisitor.visitClassDeclaration (package:dart_style/src/short/source_visitor.dart:611:26)
#1      ClassDeclarationImpl.accept (package:analyzer/src/dart/ast/ast.dart:4349:50)
#2      SourceVisitor.visit (package:dart_style/src/short/source_visitor.dart:3052:10)
#3      SourceVisitor.visitCompilationUnit (package:dart_style/src/short/source_visitor.dart:678:7)
#4      CompilationUnitImpl.accept (package:analyzer/src/dart/ast/ast.dart:5299:50)
#5      SourceVisitor.visit (package:dart_style/src/short/source_visitor.dart:3052:10)
#6      SourceVisitor.run (package:dart_style/src/short/source_visitor.dart:109:5)
#7      DartFormatter.formatSource (package:dart_style/src/dart_formatter.dart:250:24)
#8      DartFormatter.format (package:dart_style/src/dart_formatter.dart:114:12)
#9      TextualOutline.run (file:///b/s/w/ir/pkg/front_end/test/textual_outline_suite.dart:161:13)
#10     ChainContext.run.doStep.<anonymous closure> (package:testing/src/chain.dart:216:33)
#11     new Future.<anonymous closure> (dart:async/future.dart:260:40)
#12     _rootRun (dart:async/zone_root.dart:27:47)
#13     _CustomZone.run (dart:async/zone.dart:726:19)
#14     _CustomZone.runGuarded (dart:async/zone.dart:625:7)
#15     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:666:23)
#16     _rootRun (dart:async/zone_root.dart:35:13)
#17     _CustomZone.run (dart:async/zone.dart:726:19)
#18     _CustomZone.bindCallback.<anonymous closure> (dart:async/zone.dart:649:23)
#19     Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#20     _Timer._runTimers (dart:isolate-patch/timer_impl.dart:423:19)
#21     _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:454:5)
#22     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)


To re-run this test, run:

   dart --enable-asserts pkg/front_end/test/unit_test_suites.dart -p pkg/front_end/test/textual_outline/dart2js/primary_constructors/abstract_class

The test has outcome FormatterCrash, but was expected to have outcome(s) {Pass}. You might have to update the status file to the new outcome
There's a status entry looking something like

  dart2js/primary_constructors/abstract_class: Pass

which should be updated.

The status file is file:///b/s/w/ir/pkg/front_end/testcases/textual_outline.status.
```

From a flag flip CL I've been preparing (https://dart-review.googlesource.com/c/sdk/+/489241). The [test](https://github.com/dart-lang/sdk/blob/main/pkg/front_end/testcases/dart2js/primary_constructors/abstract_class.dart) seems to be crashing on short style for the `;` empty body.

I guess I'm not entirely sure what we do for short style for new language features. (also cc. @munificent)
From [this PR](https://github.com/dart-lang/dart_style/pull/1795), it looks like we implement the new nodes even on short style.

kallentu:
Anyways, uh, happy to scrap the CL or make changes. I'm not entirely sure what direction to lean towards though.

natebosch (approved):

munificent:
Catching up on this. There are two separate ways that language changes affect the formatter:

1. The analyzer team changes the AST classes in order to support new language features. If those are strictly new AST nodes, then the short style formatter usually doesn't have to be touched. It should never see code that uses those nodes. But if supporting a new language feature requires adjusting existing AST nodes (like adding a new little hierarchy for class bodies to support empty ones), then the short style formatter needs to be updated to handle the changed AST.

2. Actually formatting some new language feature's syntax. We should never need to support this in the short style. You can only get the short style on code whose language version is <3.7, so no new language features should ever appear in code that is short style formatted. The one possible edge case is that if a user explicitly enables an experiment to access the new feature in old code, but I thought even that was impossible. I believe the analyzer/parser only treats an experiment as enabled if the code in question is at the latest language version.

Given 1, it's not too surprising that we need to make some tweaks to the short formatter to handle things like empty class body AST nodes. I believe @scheglov made those changes when he changed the AST. But given 2, I'm very surprised that AST nodes for new syntax actually appear in code that is being formatted with the short style. That shouldn't be possible.

It looks like the problem is [here](https://github.com/dart-lang/sdk/blob/main/pkg/front_end/test/textual_outline_suite.dart#L159):

```dart
          // Default to the latest language version. If the test should be at
          // an older language version, it will contain a `// @dart=x.y`
          // comment, which takes precedence over this argument.
          result = new DartFormatter(
            languageVersion: DartFormatter.latestShortStyleLanguageVersion,
            experimentFlags: experimentFlags,
          ).format(result);
```

Notice that it's `latestShortStyleLanguageVersion` and not `latestLanguageVersion` which is what the comment would suggest. So the test suite is always formatting everything as if it were Dart 3.6 (!) even when it contains new syntax.

I'm guessing we used `latestShortStyleLanguageVersion` style here when I updated the API to require a language version because using the older version avoided having to reformat all of the golden files. It was OK at the time, but now that there are newer language features, we really should move to `latestLanguageVersion`.

Filed a bug: https://github.com/dart-lang/sdk/issues/63205

kallentu:
Ah, this makes so much more sense. Thank you for looking into that.

This CL probably shouldn't have made it in then, given that the fix we actually needed is a change with the test suites. Do we want to revert the changes?

munificent:
> Do we want to revert the changes?

I wouldn't bother. I think they're harmless.

PR #1825, opened 2026-04-12
Move directory formatting to a multi-threaded worker pool

kevmoo:
* Offload file I/O and UTF-8 decoding to worker isolates to improve performance.
* Implement a WorkerPool with batching and dynamic pool sizing (defaults to number of processors / 3).
* Only send formatted text back if it changed or if needed for output.
* Add test coverage for unexpected errors in worker isolates.
* Bump version to 3.1.9-wip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.1.9-wip
+
+* Move directory formatting to a multi-threaded worker pool. File I/O and
+  UTF-8 decoding are now performed in parallel on worker isolates.
+
 ## 3.1.8
 
 ### Style changes
--- a/lib/src/back_end/worker_pool.dart
+++ b/lib/src/back_end/worker_pool.dart
@@ -1,0 +1,296 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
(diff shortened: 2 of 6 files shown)


kevmoo:
For the Flutter repo

| Checkpoint | Commit | Min Time | Max Time | Average Time | Improvement vs Baseline |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **Before** (Baseline) | `f510a0a` | 17.958s | 19.548s | **18.396s** | - |
| **Final**  | `0d350de` | 9.274s | 10.373s | **9.772s** | **~46.9% faster** 🚀 |

munificent:
This is very interesting! I tried a similar approach a couple of years ago but it fizzled out when I couldn't get much of a performance improvement out of it. I expect that formatting should be nearly perfectly parallelizable: formatting a directory on a machine with four isolates and more than four cores should be about 4x as fast as doing it sequentially. But I was never able to get anywhere near that level of speed-up. I'm not sure why, but it may be that formatting was memory or GC bound in ways that span isolates.

Given that "format a big directory" is actually a fair rare user action and that maintaining concurrent code is famously hard, I decided to just leave things alone. The performance problems users complain about are ones where a single file contains code that is pathologically slow to format. Overall formatting speed on a big directory almost never comes up. That leads me to believe that it's best to keep the codebase simpler and easier to maintain, but I could be wrong if the speed up really does scale linearly with the number if isolates/cores.

kevmoo:
Outdated experiment; we can always go back to this later.

PR #1826, opened 2026-04-12
Move to a streamed model for files - mostly an experiment

kevmoo:


kevmoo:
Outdated experiment; we can always go back to this later.

PR #1828, opened 2026-04-20
Fix bug in list splitting optimization.

### LINKED ISSUE #1809: [Formatting issue]

```dart
void main() {
  // This is formatted nicely
  test('------------------ my long test name ------------------', () async {
    // My test code
  }, testOn: 'vm');

  // This is formatted differently
  test(
    '------------------ my long test name ------------------',
    () async {
      // My test code
    },
    testOn: 'vm && linux',
  );
}
```

Same as input.

```dart
void main() {
  // This is formatted nicely
  test('------------------ my long test name ------------------', () async {
    // My test code
  }, testOn: 'vm');

  // This is formatted differently
  test('------------------ my long test name ------------------', () async {
    // My test code
  }, testOn: 'vm && linux');
}
```

I'm surprised that the length of the `testOn` string affects how this is styled.

If I make the _long test name_ shorter then the `testOn` string can be longer 🤣

munificent:
If we can tell that a Piece must split into a certain state simply because of the size of its contents, it's helpful to figure that out early on before running the solver. Knowing that there are fixed splits before solving lets us break the piece tree into smaller individually formatted sections.

It's important that this optimization is conservative: it should never force a piece to split when the solver may actually find a way to not split it.

The most important piece to apply this optimization to is ListPiece since so much Dart code (especially Flutter UI code) is big nested trees of argument lists and collection literals which all use ListPiece.

The optimization is simple in most lists: if the character length of all of the elements is wider than the page, the list will definitely split. But some lists contain elements that may be block formatted, which complicates this:

```dart
  test('------------------ my long test name ------------------', () async {
    // My test code
  }, testOn: 'vm && linux');
```

Here, since the closure argument can be block formatted, we can't treat it like a normal element when seeing if the list should split. In particular, the elements before and after it ( `testOn: ...`) are no longer on the same line, so we shouldn't add their lengths when seeing if the list should be forced to split.

This regresses the flutter_scrollbar_test microbenchmark by 50%, which is a drag, but the other benchmarks are unchanged or slightly improved. Since this is fixing a correctness issue, I think it's worth the cost.

Fix #1809.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.10.0, dev]
+        sdk: [3.11.0, dev]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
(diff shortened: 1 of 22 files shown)


kallentu (approved):
Looks good, nice fix!

PR #1829, opened 2026-04-22
Handle unsupported language versions in formatter tests.

munificent:
It's rare but sometimes Dart removes support for syntax. That means a formatter test may need to be run at some version X but not be valid at X+1. Until recently, the only tests where that was true was the tests for the old constant expression syntax in Dart prior to 3.0. To handle that, there were some hacks where a versioned short style test only ran at that version. This is unlike other versioned tests where the version specifies the lower bound to run at, not the upper.

With primary constructors, we have more syntax that is no longer valid in a later Dart version: using `var` and `final` as modifiers. I expect that over time, we'll have more removed syntax like this.

To accommodate that, formatter tests can now directly specify an upper limit version where the test should no longer be run at. This removes the hack for short style and gives us a more explicit way to handle future syntax removals.

I also:

- Updated the short style switch test to use the new test syntax.

- Updated the `var` and `final` on parameter tests to use the new test syntax.

- Moved the somewhat hacky code for figuring out what language versions exist to a centralized `DartVersionHistory` class.

- Made `dart format --version --verbose` print the supported language versions.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [3.10.0, dev]
+        sdk: [3.11.0, dev]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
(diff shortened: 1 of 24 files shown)


munificent:
My plan is to wait until @scheglov has landed his changes and published dart_style 3.1.9 before I merge this in since everything is currently broken. (I made this PR by working against an older commit.)

kallentu (approved):
Awesome, thanks Bob. This looks great to me.

scheglov:
FYI, https://pub.dev/packages/analyzer/versions/13.0.0

PR #1831, opened 2026-04-28
Don't add a blank line before a comment at the end of a file or body.

### LINKED ISSUE #1644: [Tall style] New line shouldn't be inserted before comment following last block
Language version: 3.7
SDK version:  `3.7.0-323.2.beta` and `3.8.0-70.0.dev`
Page width: Default (80)

---

If I format a file with the following content, a new line is added before the comment.

```dart
void main() {
  print('Printing');
}
// Something.

```

Results in:

```dart
void main() {
  print('Printing');
}

// Something.

```

This doesn't happen for comments after earlier blocks. For example, the following doesn't change (as I'd expect).

```dart
void main() {
  f();
}
// Something.

void f() {
  print('Printing');
}
```

munificent:
In rare cases where you have a declaration with a braced body at the end of a block or file and there is a comment immediately after it, the formatter would place a blank line before the comment. It wouldn't do that if there was any other code after the comment.

For example, prior to this change, if you ran the formatter on:

```dart
main() {
  a() {
    ;
  }
  // Comment 1.

  b() {
    ;
  }
  // Comment 2.
}
```

It would produce:

```dart
main() {
  a() {
    ;
  }
  // Comment 1.

  b() {
    ;
  }

  // Comment 2.
}
```

Note how a blank line was added above `// Comment 2.` but not `// Comment 1.`. Now, it won't add a blank line before the last comment.

Fix #1644.


lib/src/front_end/ast_node_visitor.dart
```diff
@@ -107,7 +107,10 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
       // Add a blank line between directives and declarations.
       sequence.addBlank();
 
+      var needsBlank = false;
       for (var declaration in node.declarations) {
+        if (needsBlank) sequence.addBlank();
```
kallentu: Do we need this to be on a separate line to line 130? Could we `addBlank` in one if-statement?
Seems like setting the boolean twice (to true) doesn't hurt, but it's a little odd to me that we might need to call it twice, one after another.

kallentu (approved):

lib/src/front_end/ast_node_visitor.dart
munificent: Oh, good catch, thanks. Combined those.

PR #1834, opened 2026-04-29
Avoid running tests containing 'final' or 'var' on version 3.13.

kallentu:
These are the tests I thought I was fixing in https://github.com/dart-lang/dart_style/pull/1822. I updated them to actually avoid running on 3.13.

`test/tall/declaration/constructor_parameter.unit` was fixed in https://github.com/dart-lang/dart_style/commit/91f85cc453a27cbb0cfa852cf9d673b61bc92eb7

We're updating these tests for the primary constructors flag flip.


munificent (approved):
LGTM!

PR #1835, opened 2026-04-29
Allow block formatting "as", "is", and "is!" expressions.

### LINKED ISSUE #1542: Treat `as` expressions as block-formattable if the subexpression is
```dart
        final result = await buildDryRun(
          packageUri,
          createCapturingLogger(logMessages, level: Level.SEVERE),
          dartExecutable,
          linkingEnabled: false,
        );
```

->

```dart
        final result =
            await buildDryRun(
              packageUri,
              createCapturingLogger(logMessages, level: Level.SEVERE),
              dartExecutable,
              linkingEnabled: false,
            );
```

I'm not sure why the extra indentation is needed here. I expect this to be indented like it would be without an `await`

```dart
        final result = buildDryRun(
          packageUri,
          createCapturingLogger(logMessages, level: Level.SEVERE),
          dartExecutable,
          linkingEnabled: false,
        );
```

Is the await significant here? Or is it simply that the formatter deems the expression getting too big?

Source: https://github.com/dart-lang/native/blob/main/pkgs/native_assets_builder/test/build_runner/build_runner_cycle_test.dart

munificent:
Type test expressions are in a funny spot. They aren't formatted consistently with other binary operators. But they also don't allow block-formatting. The end result is weird formatting in contexts where block-formatting is possible:

```dart
      // A block-formatted expression:
      final buildConfig = await BuildConfig.build(
        outputDirectory: outputDirectory,
        packageName: await _packageName(),
        packageRoot: Directory.current.uri,
      );

      // A binary operator:
      final buildConfig =
          BuildConfig.build(
            outputDirectory: outputDirectory,
            packageName: await _packageName(),
            packageRoot: Directory.current.uri,
          ) +
          someLongExpression;

      // A type test today:
      final buildConfig =
          BuildConfig.build(
                outputDirectory: outputDirectory,
                packageName: await _packageName(),
                packageRoot: Directory.current.uri,
              )
              as BuildConfigImpl;
```

Daco requested that the formatter allow them to be block-formatted, like:

```dart
final buildConfig = BuildConfig.build(
  outputDirectory: outputDirectory,
  packageName: await _packageName(),
  packageRoot: Directory.current.uri,
) as BuildConfigImpl;
```

That's what this PR does. You can see the overall effect on a random corpus of code [here](https://gist.github.com/munificent/07f1dd403bd3e055e64042719853703b).

Fix #1542.

### DIFF
--- a/lib/src/piece/type_test.dart
+++ b/lib/src/piece/type_test.dart
@@ -1,0 +1,70 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../back_end/code_writer.dart';
+import 'piece.dart';
+
+/// A piece for an `as` or `is` expression.
+final class TypeTestPiece extends Piece {
+  /// Allow the expression to block split.
+  ///
+  /// Unlike most pieces that allow block splitting, the cost here isn't zero
+  /// because we would prefer to split at `=` if it lets the entire type test
+  /// expression fit on one line:
+  ///
+  ///     variable =
+  ///         function(argument) as Type;
+  static const State _blockSplitExpression = State(1);
+
+  /// The expression being tested or cast.
+  final Piece _expression;
+
+  /// The `as`, `is`, or `is!` operator.
+  final Piece _operator;
+
+  /// The type being tested against or cast to.
+  final Piece _type;
+
+  /// Whether the expression can be block formatted.
+  final bool _canBlockSplit;
+
+  TypeTestPiece(
+    this._expression,
+    this._operator,
+    this._type, {
+    bool canBlockSplit = false,
+  }) : _canBlockSplit = canBlockSplit;
+
+  @override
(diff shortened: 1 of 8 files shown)


lib/src/piece/type_test.dart
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
kallentu: Change copyright year to 2026.

kallentu (approved):

lib/src/piece/type_test.dart
munificent: Oops, good catch. I'd blame Gemini, but I have made that same mistake many times.

natebosch (approved):
I think the changes in the random corpus look like improvements to me.

When you merge can you include a clear description of the change in the commit message? The current one reads like a good justification for the change, but doesn't make it obvious exactly what block formatting these expressions implies. An example diff or similar would help.

munificent:
> When you merge can you include a clear description of the change in the commit message?

Good point, done. :)

PR #1836, opened 2026-05-01
Format the guard in an if-case to split if the pattern block splits.

### LINKED ISSUE #1596: If case when should prefer adding a new line to `when` and not to the inner variables [New style]
If you have a code like:

```dart
if (node.parent case MethodInvocation(:var target) when target is SimpleIdentifier?)
```

Today it formats like this:

```dart
if (node.parent case MethodInvocation(
  :var target,
) when target is SimpleIdentifier?)
```

In this case I think it should _prefer_ to add a new line before `when` rather than making a whole new line for `:var target`.

munificent:
The formatter allows block-formatting in if-case patterns, like:

```dart
if (expression case Thing(
  :var property,
  :var another,
)) {
  ...
}
```

It also allows this if there is a guard clause:

```dart
if (expression case Thing(
  :var property,
  :var another,
) when condition) {
  ...
}
```

But, in general, I don't like when code gets buried at the end of a line like that. Also, because we allow this and because block-formatting is usually lower cost than other kinds of splits, it means the formatter prefers this:

```dart
if (expression case Thing(
  :var property,
) when condition) {
  ...
}
```

Instead of:

```dart
if (expression case Thing(:var property)
    when condition) {
  ...
}
```

But I think the latter is strictly better. This PR tweaks the formatting rule so that an if-case can only block-format the pattern if there is no guard clause. That leads it to prefer the latter of the previous two examples.

In practice, this tweak turns out to affect very little code. I ran it on a huge corpus of pub packages and there were [only 32 changes](https://gist.github.com/munificent/6be922e612850e98e849af20c6688ce0). To my eye, those all look like improvements.

Fix #1596.

### DIFF
--- a/lib/src/front_end/piece_factory.dart
+++ b/lib/src/front_end/piece_factory.dart
@@ -900,27 +900,15 @@ mixin PieceFactory {
 
       var guardPiece = optionalNodePiece(caseClause.guardedPattern.whenClause);
 
-      if (style.blockFormatIfCaseWithGuard) {
-        pieces.add(
-          IfCasePiece3Dot12(
-            expressionPiece,
-            casePiece,
-            guardPiece,
-            canBlockSplitPattern:
-                caseClause.guardedPattern.pattern.canBlockSplit,
-          ),
-        );
-      } else {
-        pieces.add(
-          IfCasePiece(
-            expressionPiece,
-            casePiece,
-            guardPiece,
-            canBlockSplitPattern:
-                caseClause.guardedPattern.pattern.canBlockSplit,
-          ),
-        );
-      }
+      pieces.add(
+        IfCasePiece(
+          expressionPiece,
+          casePiece,
+          guardPiece,
+          canBlockSplitPattern: caseClause.guardedPattern.pattern.canBlockSplit,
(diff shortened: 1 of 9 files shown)


lib/src/front_end/piece_factory.dart
```diff
@@ -900,14 +900,27 @@ mixin PieceFactory {
...
-        IfCasePiece(
-          expressionPiece,
-          casePiece,
-          guardPiece,
-          canBlockSplitPattern: caseClause.guardedPattern.pattern.canBlockSplit,
-        ),
-      );
+      if (style.blockFormatIfCaseWithGuard) {
+        pieces.add(
+          IfCasePiece3Dot12(
```
natebosch: [nit] The `blockFormatIfCaseWithGuard` encapsulates the detail of the version, then the usage site repeats that detail.

Consider renaming one or the other so the level of abstraction is consistent here.

natebosch (approved):
I like the diff for places where the part of the pattern before `when` can stay on one line.

The other diffs which got taller don't look like an improvement to me, but I can imagine some folks would prefer it. It's also enough of an edge case that I don't think it's worth trying to add more heuristics.

kallentu (approved):

PR #1841, opened 2026-05-22
Move all tests to be explicitly versioned

munificent:
This PR is two related changes:

* **Update tool/update_tests.dart to handle versioning better.** Before, it would only update the version-specific sections for each test that were explicitly authored, but wouldn't split them itself. Now, it fully handles generating the set of version-specific output expectations that are needed for each test.

    After doing this, I ran it to update all of the tests. As expected, most were unchanged but a couple had unnecessary multiple output sections with identical output. A few others had different outputs at different versions that weren't captured by the tests and are now pinned down more precisely.

* **Remove support for unversioned tests.** This simplifies the test code and makes tests easier to read because it's clearer what versions the output apply to.

I split this into two commits to make the test changes easier to see. The first commit has the meaningful test changes. The test changes in the second commit are purely mechanical.


kallentu (approved):
Looks fine. Is the general guidance to always write version numbers for tests now? And if so, could we update the test writing docs to encourage that?

munificent:
> Is the general guidance to always write version numbers for tests now?

Yes.

> And if so, could we update the test writing docs to encourage that?

Oh, good catch! I meant to update the README but got distracted and forgot. Done.

PR #1842, opened 2026-05-26
Prefer to split call chains for single-element targets.

### LINKED ISSUE #1732: [Formatting issue] Formater seems to prefer splitting on arguments too much

```dart
                        GoRouter.of(context)
                            .go('/play/session/${level.number}');
```

```dart
                        GoRouter.of(
                          context,
                        ).go('/play/session/${level.number}');
```

Describe how you expected the formatter to format the code, ideally by showing what you had in mind as code.

```dart
                        GoRouter.of(context)
                            .go('/play/session/${level.number}');
```

**Anything else we should know?**

I personally think the original is better, as it shows a hierarchy.

I can imagine that the new style, where the argument is on its own line, is better for ui-as-code. But when the argument is a simple, relatively short variable, I think line-splitting on that makes the code less readable, not more.

This might be related or even duplicitous with https://github.com/dart-lang/dart_style/issues/1731 (though that one is a more obvious failure in my book, so I'm keeping them separate).

munificent:
One of the most challenging parts of formatting Dart code well is dealing with method call chains on targets that can themselves split. The formatter needs to decide whether to split the target and avoid splitting the call chain or vice versa. For example:

```dart
// Split target:
function(
  argument,
).method().another();

// Or split chain:
function(argument)
  .method()
  .another();
```

Heuristics for that choice that look great on some callsites often make others look worse. I've toyed with many variations over the years.

This adds a fairly simple rule based on the examples in (#1732): If the target of a call chain only has *one* element or argument, then prefer to split the call chain and keep the target together. So in the above example, if prefers the second output. If `function()` had multiple arguments, it would prefer the first.

An intuitive argument in favor of this rule is that when we split a comma-separated argument or element list, what feels most intuitive to readers is splitting *between* the elements. Then as a secondary effect, we also split around the delimiting brackets to. If there is only one element, then there's no way to split between them, only around them. That doesn't look as good, so this style change avoids that.

Pragmatically, I think the results of the new rule look good. You can see a diff of running this on a random assortment of pub packages [here](https://gist.github.com/munificent/7d39fa6cd7667e8b9d9458638753e4f3).

This is a larger change than most these days. I ran it on a large corpus of files from Pub:

```
Affected / total files : 3,511 / 93,664 (~3.7%)
Affected / total lines : 35,770 / 22,424,287 (0.13%)
```

For affected lines, I'm ignoring additions and counting only the number of deleted lines from running `git diff --stat`. Since the formatter is always replacing a chunk of lines with another, I think that's a good way to measure the number of original lines which are impacted.

Fix #1732

### DIFF
--- a/lib/src/front_end/chain_builder.dart
+++ b/lib/src/front_end/chain_builder.dart
@@ -65,7 +65,11 @@ final class ChainBuilder {
   ///       argument,
   ///     )
   ///         .method();
-  late final bool _allowSplitInTarget;
+  bool _allowSplitInTarget = false;
+
+  /// Whether the target of the chain is a call or collection with a single
+  /// argument or element (and we are in a style that treats that specially).
+  bool _isSingleElementTarget = false;
 
   /// The dotted property accesses and method calls following the target.
   final List<ChainCall> _calls = [];
@@ -132,6 +136,7 @@ final class ChainBuilder {
       indent: Indent.cascade,
       blockCallIndex: blockCallIndex,
       allowSplitInTarget: _allowSplitInTarget,
+      singleElementTarget: _isSingleElementTarget,
       is3Dot7: _visitor.style.is3Dot7,
     );
 
@@ -234,21 +239,18 @@ final class ChainBuilder {
       leadingProperties: leadingProperties,
       blockCallIndex: blockCallIndex,
       allowSplitInTarget: _allowSplitInTarget,
+      singleElementTarget: _isSingleElementTarget,
       is3Dot7: _visitor.style.is3Dot7,
     );
   }
 
   /// Given [expression], which is the expression for some call chain, traverses
-  /// the selectors to fill in the list of [_calls].
-  ///
-  /// Otherwise, it's a method chain, and this recursively calls itself for the
-  /// targets to unzip and flatten the nested selector expressions. Then it
-  /// initializes [_target] with the innermost subexpression that isn't a part
-  /// of the call chain. For example, given:
+  /// the selectors to fill in the list of [_calls] and initialize [_target].
+  /// For example, given:
   ///
   ///     foo.bar()!.baz[0][1].bang()
   ///
-  /// This returns `foo` and fills [_calls] with:
+  /// This initializes [_target] to `foo` and fills [_calls] with:
   ///
   ///     .bar()!
   ///     .baz[0][1]
@@ -314,8 +316,8 @@ final class ChainBuilder {
         _calls.add(ChainCall(piece, CallType.property));
 
       // Postfix expressions.
-      case FunctionExpressionInvocation():
-        _unwrapPostfix(expression.function, (target) {
+      case FunctionExpressionInvocation(:var function):
+        _unwrapPostfix(expression, function, (target) {
           return _visitor.pieces.build(() {
             _visitor.pieces.add(target);
             _visitor.pieces.visit(expression.typeArguments);
@@ -329,7 +331,7 @@ final class ChainBuilder {
         // with an index expression like:
         //
         //     object..[index].chain();
-        _unwrapPostfix(target, (target) {
+        _unwrapPostfix(expression, target, (target) {
           return _visitor.pieces.build(() {
             _visitor.pieces.add(target);
             _visitor.writeIndexExpression(expression);
@@ -337,7 +339,7 @@ final class ChainBuilder {
         });
 
       case PostfixExpression() when expression.operator.type == TokenType.BANG:
-        _unwrapPostfix(expression.operand, (target) {
+        _unwrapPostfix(expression, expression.operand, (target) {
           return _visitor.pieces.build(() {
             _visitor.pieces.add(target);
             _visitor.pieces.token(expression.operator);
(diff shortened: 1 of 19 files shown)


munificent:
@natebosch and @kallentu, this isn't a *huge* style change, but it's a big enough one that I'd like your thoughts on not just the implementation, but whether it's a good style change too. Thanks!

lib/src/piece/chain.dart
```diff
@@ -132,6 +132,7 @@ abstract base class ChainPiece extends Piece {
     int blockCallIndex = -1,
     Indent indent = Indent.expression,
     required bool allowSplitInTarget,
+    required bool singleElementTarget,
```
kallentu: Small nit, but `hasSingleElementTarget` instead?

kallentu (approved):
LGTM. I'm also in favor of this change. I think it reads more easily for my brain

lib/src/piece/chain.dart
munificent: I had `isSingleElementTarget` at first but that read kind of funny so I remove the verb based on [this guideline](https://dart.dev/effective-dart/design#consider-omitting-the-verb-for-a-named-boolean-parameter). But now that you mention it, I like `has` more. Done.

natebosch:
> often make others look worse

Do you have examples that you think look worse when we expand this behavior beyond single-argument calls? I tried removing the ` && target.hasSingleElement;` requirement from the condition and it didn't impact anything in this repo. I could take or leave diff on the contrived example I wrote. I'll look around for more code to try it on.

lib/src/front_end/chain_builder.dart
```diff
@@ -356,27 +358,56 @@ final class ChainBuilder {
...
       _calls.last.wrapPostfix(createPostfix);
     }
   }
+
+  void _initializeTargetProperties(Expression target) {
+    _allowSplitInTarget = target.canBlockSplit;
+
+    _hasSingleElementTarget =
+        _visitor.style.avoidSplittingSingleElementCallChainTargets &&
+        target.hasSingleElement;
```
natebosch, line 411: [optional] I've only checked a narrow corpus, but I like the way this looks if we do it no matter the number of elements.

```suggestion
    _hasSingleElementTarget =
        _visitor.style.avoidSplittingSingleElementCallChainTargets;
```

natebosch (approved):
Looks like an improvement to me, a pretty drastic one in some cases in your long diff.

natebosch:
Removing the single argument criteria for this change resulted in no diff here, `dart-lang/tools`, or `dart-lang/test`. It has some impact in `dart-lang/core`, and I think looks like neutral to positive changes.

https://gist.github.com/natebosch/cce83d2ba5015055f40069d6e0773eb8

munificent:
> I've only checked a narrow corpus, but I like the way this looks if we do it no matter the number of elements.

That's an interesting idea. There are a few axes we could consider tweaking the heuristic (and boy have I tried many combinations over the years already).

## Always preferring to split

The diff you have actually doesn't do exactly this. It tells the formatter to prefer splitting method chains regardless of what the target expression is at all. That's why, for example, your diff has:

```diff
-    var replaced = chars.replaceAll(
-      Characters(language),
-      Characters('Hangul!'),
-    );
+    var replaced = chars
+        .replaceAll(Characters(language), Characters('Hangul!'));
```

Here, the target is just `chars`, so there's no arguments at all. Way back in the early days of the rewrite, I believe I did toy with making call chains generally lean towards splitting the chain instead of the target. It looks nice in many places, but I think less nice in others. Some examples:

```dart
// Current:
  expect(
    [
      ...caseUnreachabilities,
      if (nonExhaustiveness != null) nonExhaustiveness,
    ].join('\n'),
    errors,
  );

// Always prefer splitting call chain:
  expect(
    [...caseUnreachabilities, if (nonExhaustiveness != null) nonExhaustiveness]
        .join('\n'),
    errors,
  );

// Current:
    buffer.writeln('    _controller = SmartContractController(');
    buffer.writeln('    abi: abi,');
    buffer.writeln('    contractAddress: contractAddress is String');
    buffer.writeln(
      '        ? SmartContractAddress.fromBech32(contractAddress)',
    );
    buffer.writeln('        : contractAddress as Address,');
    buffer.writeln('    networkProvider: networkProvider,');

// Always prefer splitting call chain:
    buffer.writeln('    _controller = SmartContractController(');
    buffer.writeln('    abi: abi,');
    buffer.writeln('    contractAddress: contractAddress is String');
    buffer
        .writeln('        ? SmartContractAddress.fromBech32(contractAddress)');
    buffer.writeln('        : contractAddress as Address,');
    buffer.writeln('    networkProvider: networkProvider,');
```

Most of the changes could go either way in my eyes. Also, it causes a bug where it confuses the formatter when a method chain is inside an assignment sometimes. For example:

```dart
// Current:
  List<Map<String, dynamic>> serializeAsList() => serializeAsMap().entries
      .map((entry) => (<String, dynamic>{entry.key: entry.value}))
      .toList();

// If we always set the split cost to zero:
  List<Map<String, dynamic>> serializeAsList() =>
      serializeAsMap().entries
          .map((entry) => (<String, dynamic>{entry.key: entry.value}))
          .toList();
```

Here, the second example is strictly worse than the first one. I spent some time trying to figure out what's going on, but I felt like I was going down a rabbit hole and stopped. I want to time box these changes pretty tightly to get them out in 3.13.

Also, if we take this approach, there are a *lot* of changes. A really large fraction of call chains get formatted differently, so we'd have to do a whole change process and migration. I don't want to do that now.

## Preferring to split on one or more arguments

I think the right way to implement the rule you suggest looks like:

```dart
// TODO: Rename.
  bool get hasSingleElement {
    var expression = this;
    while (true) {
      if (expression is ParenthesizedExpression) {
        expression = expression.expression;
      } else if (expression is AwaitExpression) {
        expression = expression.expression;
      } else if (expression is PostfixExpression) {
        expression = expression.operand;
      } else {
        break;
      }
    }

    return switch (expression) {
      MethodInvocation(:var argumentList) ||
      InstanceCreationExpression(:var argumentList) ||
      FunctionExpressionInvocation(
        :var argumentList,
      ) => argumentList.arguments.isNotEmpty, // <--
      ListLiteral(:var elements) ||
      SetOrMapLiteral(:var elements) => elements.isNotEmpty, // <--
      RecordLiteral(:var fields) => fields.isNotEmpty, // <--
      _ => false,
    };
  }
```

So we still distinguish targets that are non-empty delimited sequences versus other kinds of target expressions. We just don't limit it to single-argument ones.

[Here](https://gist.github.com/munificent/e0d5a877ac71364d8515b6d9c9456764) is a diff of that change relative to the current PR state. It's mostly a wash to me. I think most of the examples where the target is a collection look worse. The examples where it's an argument list are sometimes a little better but sometimes a little worse?

I don't think they are *enough* better to justify the churn, so I'm inclined to keep this heuristic tweak restricted to the single-argument case.

I still don't love all of the heuristics here. Sometimes I feel like there's an elegant formalism out there somewhere that would do the right thing in all sorts of complex cases, but even with lots of experimentation over the years, I still haven't found it. The [shape stuff](https://github.com/dart-lang/dart_style/pull/1687) got closer but was sorted of layered on top of the existing architecture instead of the rules being fully designed around shape constraints. If I had infinite time, I would try to rewrite the entire formatter around that and try to get rid of numeric scoring completely, but I can't see that happening this century.

If it's OK with you, I'll land this as-is (well, with the fix for unsplit property chains). I think the single-argument rule is narrow enough to not cause undue churn but improves the results significantly in places where it kicks in.

natebosch:
> It's mostly a wash to me. I think most of the examples where the target is a collection look worse. The examples where it's an argument list are sometimes a little better but sometimes a little worse?

I prefer the single line collection versions in that diff. Overall I like the changes with only a few exceptions where the extra indentation exaggerates the problem of deep nesting. I can imagine there may be folks happy with this current PR and wouldn't be happy with the alternative though, so no need to hold this one up.

> If it's OK with you, I'll land this as-is

SGTM

PR #1848, opened 2026-06-03
Handle overflowing solutions more gracefully.

### LINKED ISSUE #1802: Overflowing string literal earlier in expression forces worse formatting

See the `categories` setter call at the bottom here.

```dart
base mixin DartAnalyzerSupport
    on ToolsSupport, LoggingSupport, RootsTrackingSupport, FileSystemSupport
    implements SdkSupport, ProcessManagerSupport {
  @visibleForTesting
  static final resolveWorkspaceSymbolTool = Tool(
    name: 'resolve_workspace_symbol',
    description:
        'Look up a symbol or symbols in all workspaces by name. Can be used '
        'to validate that a symbol exists or discover small spelling '
        'mistakes, since the search is fuzzy.',
    inputSchema: Schema.object(
      properties: {
        ParameterNames.query: Schema.string(
          description:
              'Queries are matched based on a case-insensitive partial name '
              'match, and do not support complex pattern matching, regexes, '
              'or scoped lookups.',
        ),
      },
      description:
          'Returns all close matches to the query, with their names '
          'and locations. Be sure to check the name of the responses to ensure '
          'it looks like the thing you were searching for.',
      required: [ParameterNames.query],
      additionalProperties: false,
    ),
    annotations: ToolAnnotations(title: 'Project search', readOnlyHint: true),
  )..categories = [FeatureCategory.dart, FeatureCategory.flutter, FeatureCategory.lsp];
}
```

It won't change it (note that if you force a trailing comma for the `categories` setter it will reformat it to remove the trailing comma).

I expected it to use trailing commas for the list passed to the `categories` setter to keep lines under 80 chars.

```dart
base mixin DartAnalyzerSupport
    on ToolsSupport, LoggingSupport, RootsTrackingSupport, FileSystemSupport
    implements SdkSupport, ProcessManagerSupport {
  @visibleForTesting
  static final resolveWorkspaceSymbolTool = Tool(
    name: 'resolve_workspace_symbol',
    description:
        'Look up a symbol or symbols in all workspaces by name. Can be used '
        'to validate that a symbol exists or discover small spelling '
        'mistakes, since the search is fuzzy.',
    inputSchema: Schema.object(
      properties: {
        ParameterNames.query: Schema.string(
          description:
              'Queries are matched based on a case-insensitive partial name '
              'match, and do not support complex pattern matching, regexes, '
              'or scoped lookups.',
        ),
      },
      description:
          'Returns all close matches to the query, with their names '
          'and locations. Be sure to check the name of the responses to ensure '
          'it looks like the thing you were searching for.',
      required: [ParameterNames.query],
      additionalProperties: false,
    ),
    annotations: ToolAnnotations(title: 'Project search', readOnlyHint: true),
  )..categories = [
    FeatureCategory.dart,
    FeatureCategory.flutter,
    FeatureCategory.lsp,
  ];
}
```

**Anything else we should know?**

This appears to be related to the string literals, if I shorten them, then it will reformat the list to use trailing commas, but it also indents the entire constructor call.

This would then cause the string literals to be >80 chars, which can't be fixed without manually editing them.

I would still prefer it do that though because that is something I can fix, and it isn't obvious otherwise that this is the actual cause, it just seems like a formatter bug.

### LINKED ISSUE #1803: Prefer long lines in comments over code
A somewhat confusing situation happened when upgrading language version and reformatting with the tall style. When the formatter chooses to move the `() async {` part of the argument in a call to `test` to a new line changes the indentation of the body to add 1 extra level. If this pushes exactly 1 line of comment past the length limit it will go forward, but once this impacts 2 or more lines of comments it decides instead to keep the body unindented and breaks the line length goal elsewhere. Or maybe the heuristic is on total overflow more than number of lines?

A minimal repro:

Here the `() async {` extends past 80 characters

```dart
import 'package:test/test.dart';

void main() {
  test('XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX', () async {
    // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

    // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
  });
}
```

The reproduction case is sensitive to the length of the title and the lengths of the comments. This came up in a package with a long test where it was confusing to find the root cause. https://github.com/dart-lang/tools/pull/2324/changes/f55f3a20dbed7adf1276124080191ed7040cb1dc

Generally I think it would be less confusing to prioritize fitting code within the line length than comments, possibly reducing the weight of long comment lines to zero.

This is a low priority, I don't expect it comes up too often, and changing it won't eliminate the lint, just move it to a more actionable spot.

### LINKED ISSUE #1837: The formatter can split lines when a trailing comment would cause the entire line to go over the page width
It seems that the formatter chooses to count the length of a comment on the same line as code towards the total line length for formatting purposes?

I would expect the formatter to leave the comment as is, if the line of code preceding it is equal to or shorter than the configured page width.

Given the example below, where the comma at the end of the line of code is at 36 characters,
but the entire line, including the period at the end of the comment is 111 characters.

```dart
void main() {
  final knownCrossImports = [];

  test(
    skip: knownCrossImports.isEmpty, // [intended]: No message is needed when there are no known cross imports.
  );
}

void test({required bool skip}) {}
```

```dart
void main() {
  final knownCrossImports = [];

  test(
    skip: knownCrossImports
        .isEmpty, // [intended]: No message is needed when there are no known cross imports.
  );
}

void test({required bool skip}) {}
```

Describe how you expected the formatter to format the code, ideally by showing what you had in mind as code.

```dart
void main() {
  final knownCrossImports = [];

  test(
    skip: knownCrossImports.isEmpty, // [intended]: No message is needed when there are no known cross imports.
  );
}

void test({required bool skip}) {}
```

**Anything else we should know?**

This came up in https://github.com/flutter/flutter/pull/185493 due to the formatter chopping off the `[intended]` comment, which caused the skip test validation to fail.

The page width in flutter/flutter is set to 100.

This reproduces on 3.13.0-93.0.dev

munificent:
When no solution fits in the page width, prefer solutions where the overflowing characters are from string literals or comments, since those are the most common cause. We want the user to see them overflowing so they realize that's what they need to manually split.

You can see a diff of applying this to a corpus [here](https://gist.github.com/munificent/fe306269607bb3da6b5f44a1639090ae).

Fix #1802.
Fix #1803.
Fix #1837.


munificent:
This PR is relatively small (don't freak out by the diff, it's mostly comments and tests). But you wouldn't believe how many other approaches I had to try before I got this working.

lib/src/front_end/formatting_style.dart
```diff
@@ -81,6 +81,42 @@ final class FormattingStyle {
...
+  /// hanging past the page width so it's clear to the user where they need to
+  /// split things to get everything to fit. Also, this minimizes the format
+  /// churn when they do fix those strings or comments.
+  ///
+  /// To do that, the formatter distinguishes "soft" characters from other
+  /// kinds of code. "Soft" code is string literals, comments, or a few other
+  /// things that often follow a string literal or comment: `,`, `;`, or `() {`
+  /// for a trailing block-formatted lambda. When an overflowing line of code
+  /// ends in soft characters, the overflow cost of all of those characters is
+  /// collapsed to a single point of penalty instead of one per character.
```
kallentu: Why is the cost collapsed to 1? Is there a difference if it was 0?

lib/src/front_end/piece_writer.dart
```diff
@@ -106,8 +110,8 @@ final class PieceWriter {
 
     if (token.precedingComments != null) {
       // Don't append to the previous token if there is a comment after it.
-      _beginCodeToken(token);
-    } else if (_currentCode case var code?) {
+      _beginCodeToken(token, soft);
+    } else if (_currentCode case var code? when code.isSoft == soft) {
```
kallentu, line 115: Can you write a comment for this change here? Seems like we're splitting the pieces more granularly depending on if its soft or not.

kallentu (approved):

lib/src/front_end/formatting_style.dart
```diff
@@ -81,6 +81,42 @@ final class FormattingStyle {
...
+  /// that the strings or comments were the problem.
+  ///
+  /// What we want is for the formatter to leave those strings or comments
+  /// hanging past the page width so it's clear to the user where they need to
+  /// split things to get everything to fit. Also, this minimizes the format
+  /// churn when they do fix those strings or comments.
+  ///
+  /// To do that, the formatter distinguishes "soft" characters from other
+  /// kinds of code. "Soft" code is string literals, comments, or a few other
+  /// things that often follow a string literal or comment: `,`, `;`, or `() {`
```
natebosch, line 106: Should `async` be listed here? I think I recall overflowing `() async {` following long string literals.

natebosch (approved):

lib/src/front_end/formatting_style.dart
munificent: Good question. Added a sentence to the doc. If we did zero penalty, then if all of the overflowing characters were from strings or comments, then the formatter wouldn't realize there was any overflow at all. That in turn means these two formats would both have no overflow:

```dart
//                        page width -> |
veryLongLeftOperand + 'some overflowing string';

veryLongLeftOperand +
    'some overflowing string';
```

But we definitely want the formatter to prefer the second solution here (even though it has a point of higher *cost* from splitting the `+`) because that split actually does get the code to fit in the page width.

So we collapse all soft overflow to almost but not quite nothing, i.e. 1.

It's a heuristic and no heuristic is perfect. (We could, for example, collapse soft overflow to zero points if there is any other hard overflow too.) But I think this heuristic seems to do a pretty good job. I also tried another hueristic where we track soft overflow as a completely separate bit where there were both the number of characters of hard overflow, and a bool for whether or not there was any soft overflow. That was more complex and didn't seem to really improve anything.

lib/src/front_end/piece_writer.dart
munificent: > Seems like we're splitting the pieces more granularly depending on if its soft or not.

Right, that way we don't accidentally count a soft token as hard if it happens to follow other hard code or vice versa. Added a comment.

PR #1849, opened 2026-06-05
Move fixing a piece's state by page width to before solving.

### LINKED ISSUE #1847: Bug related to fixedStateForPageWidth() optimization leads to incorrect formatting
A key optimization the formatter performs is that if it can tell that the contents of a Piece are so large that the Piece will be forced to split regardless of any surrounding context, then it binds the Piece to the split state eagerly before solving gets involved.

The most important Piece implementing this optimization is ListPiece, which is used for collections and argument lists. Without this, large deeply nested collections and function calls (pervasive in Flutter) can be combinatorial to format. In theory, the optimzation is simple: if the total character width of all of the elements inside the list is greater than the page width or contain a hard newline, then it should definitely split.

However, "block formatting" makes that harder. Consider:

```dart
someFunction(veryLongArgument, () {;});
```

Here, there will be a hard newline after the `;` in the closure argument, so it seems like the ListPiece for the surrounding argument list must be forced to split like:

```dart
someFunction(
  veryLongArgument,
  () {
    ;
  },
);
```

But no, because this argument list can instead be block formatted like:

```dart
someFunction(veryLongArgument, () {
  ;
});
```

So knowing when the optimization can be applied and the Piece eagerly pinned to a state based on its contents is a little subtle. The [code for this](https://github.com/dart-lang/dart_style/blob/main/lib/src/piece/list.dart#L233-L277) has already had one bug. When investigating some other formatter changes, I ran into a few cases that indicate there is still an issue that I *believe* is related to this optimization.

Here is one example I found:

```dart
main() async {
  {
    await MethodChannelContainer()
        .onMethodChannelInvoke("reportCrash", <String, dynamic>{
          "time": nowTime,
          "errorValue": errorName,
          "reason": reason,
          "stacktrace": stacktrace,
        });
  }
}
```

That's what the formatter currently produces. I think it should produce:

```dart
main() async {
  {
    await MethodChannelContainer().onMethodChannelInvoke(
      "reportCrash",
      <String, dynamic>{
        "time": nowTime,
        "errorValue": errorName,
        "reason": reason,
        "stacktrace": stacktrace,
      },
    );
  }
}
```

If you disable the ListPiece optimization, that is what it outputs, which is why I think the bug relates to `fixedStateForPageWidth()`. I haven't been able to figure out what the actual bug is yet. And, unfortunately, the optimization is critical to performance so can't simply be disabled.

It's fairly rare that real-world code runs into this, so for now I'm just going to file this issue so I don't lose track of it.

munificent:
A key optimization the formatter performs is that if it can tell that the contents of a Piece are so large that the Piece will be forced to split regardless of any surrounding context, then it binds the Piece to the split state eagerly before solving gets involved. This radically speeds up solving because the solver doesn't have to explore a large number of *combinations* of piece states before finding the state for one piece.

Before this PR, that optimization was applied in the middle of building and formatting a solution. When writing an piece inline to a solution, we'd try to see if it had a fixed state based on its contents and the remaining page width at that point. Doing this optimization in the middle of writing the piece was good for performance because it gave us more information about the surrounding indentation. That in turn effectively reduced the page width which made it more likely the optimization would apply and successfully pick a state for the piece.

However, the process of exploring the solution tree is pretty complex and having some pieces get bound in the solution right in the middle of that process is brittle. In particular, it turns out there was at least one bug. When formatting a solution, we might eagerly bind some piece deep in the middle of the piece tree to a state based on its contents (the optimization). At that point, we would also increment the solution's cost to pay for that state.

Later, we might derive a new solution from that solution that formats part of its subtree separately. Separate formatting is another optimization where we can reuse formatted subtrees across multiple solutions. When we separately format, we add in the entire formatted output of a subtree. We also need to include the cost of any pieces split in that subtree.

These two optimizations didn't play nice together. In some cases, the eager piece binding would bind a piece to some state and pay its cost. Later, that piece would be part of a subtree that was separately formatted. When the subtree was merged, we would pay its cost again. The result is that when the eager piece binding optimization applied, sometimes a piece's state cost would be double counted.

That in turn would cause that solution to be not as desirable as it should be and sometimes other worse solutions would win. Looking through the diff, it mostly seems to manifest in code where you have a `.` method call whose argument list contains a large collection literal. The split list literal's cost would be double counted and instead some other solution that doesn't push the collection far enough to the right to trigger the optimization would win.

For example, currently the formatter produces:

```dart
main() async {
  {
    await MethodChannelContainer()
        .onMethodChannelInvoke("reportCrash", <String, dynamic>{
          "time": nowTime,
          "errorValue": errorName,
          "reason": reason,
          "stacktrace": stacktrace,
        });
  }
}
```

There is no reason to split before `.onMethodChannelInvoke`. It's just double-counting the cost of the split list literal which incorrectly bumps the cost of the solution that should win:

```dart
main() async {
  {
    await MethodChannelContainer().onMethodChannelInvoke(
      "reportCrash",
      <String, dynamic>{
        "time": nowTime,
        "errorValue": errorName,
        "reason": reason,
        "stacktrace": stacktrace,
      },
    );
  }
}
```

This PR fixes that by moving the "eagerly bind pieces based on contents" optimization out of the solver. Instead, we do one big traversal of the piece tree first and bind what we can. When we do, since the piece will now be bound to that state for all possible solutions, there's no need to count its cost. And thus we never double-count it (or even single-count it for that matter).

You can see the diff applied to a large corpus [here](https://gist.github.com/munificent/db66c4c2573d99cb9d25dbd0c1cb7b13). That's the entire diff from a very large corpus, so the amount of code affected is relatively small. Interestingly, it seems to be very often in method channel calls. I'm guessing because those often combine a call chain with a large collection literal.

This does regress performance. :( Doing the optimization earlier is beneficial in some ways because we'll only try to apply it to a given piece once instead of once per solution that happens to reach it. But it's also worse because the effective page width is always the full width of the file without taking any indentation for where the piece appears into account. Thus the optimization kicks in less frequently. Overall, it's a regression but not catastrophic:

Here's the benchmarks on my Macbook (higher percentages = faster):

```
block                             0.080    0.088    0.174    0.091     82.7%
chain                             0.704    0.721    0.759    0.724     95.5%
collection                        0.408    0.420    0.447    0.422     64.1%
collection_large                  1.759    1.797    1.936    1.806    101.2%
conditional                       0.084    0.085    0.105    0.087     80.0%
curry                             0.173    0.175    0.203    0.178     99.4%
curry_2                           0.346    0.356    0.398    0.360     99.5%
ffi                               0.159    0.162    0.185    0.166     99.2%
flutter_popup_menu_test           0.542    0.559    0.595    0.559     93.1%
flutter_scrollbar_test            1.996    2.017    2.055    2.019     48.8%
function_call                     1.781    1.822    1.904    1.821     99.6%
infix_large                       0.714    0.737    0.775    0.740    101.4%
infix_small                       0.182    0.185    0.222    0.189    107.5%
interpolation                     0.142    0.146    0.175    0.148     96.3%
interpolation_1516                0.125    0.126    0.141    0.128     95.5%
large                             4.046    4.145    4.362    4.149     93.3%
top_level                         0.149    0.155    0.181    0.157     97.3%
```

And here's the time formatting a giant corpus of pub code:

```
Before this PR: 233.91s user 30.02s system 109% cpu 4:01.20 total
After:          261.15s user 30.66s system 108% cpu 4:27.88 total
```

Fix #1847.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,36 @@
 
 ### Style changes
 
+* Fix a bug in an eager splitting optimization that would led the formatter to
+  prefer less desirable solutions (#1847).
+
+  Typically, the code affected by this bug is a call chain that contains an
+  argument list with a large collection literal, as in:
+
+  ```dart
+  // Before:
+  await MethodChannelContainer()
+      .onMethodChannelInvoke("reportCrash", <String, dynamic>{
+        "time": nowTime,
+        "errorValue": errorName,
+        "reason": reason,
+        "stacktrace": stacktrace,
+      });
+
+  // After:
+  await MethodChannelContainer().onMethodChannelInvoke(
+    "reportCrash",
+    <String, dynamic>{
+      "time": nowTime,
+      "errorValue": errorName,
+      "reason": reason,
+      "stacktrace": stacktrace,
+    },
+  );
+  ```
+
+  This change is language versioned and only affects code at 3.13 or higher.
+
 * Prefer to split call chains for single-element targets (#1732).
 
   When formatting a method call chain whose target can also split, the formatter
--- a/lib/src/back_end/code_writer.dart
+++ b/lib/src/back_end/code_writer.dart
@@ -154,7 +154,7 @@ final class CodeWriter {
   /// Increases the indentation by [indent] relative to the current amount of
   /// indentation.
   void pushIndent(Indent indent) {
-    if (_cache.is3Dot7) {
+    if (_cache.style.is3Dot7) {
       _pushIndent3Dot7(indent);
     } else {
       var parent = _indentStack.last;
@@ -191,7 +191,7 @@ final class CodeWriter {
   /// This is only used in a couple of corners of if-case and for-in headers
   /// where the indentation is unusual.
   void pushCollapsibleIndent() {
-    if (_cache.is3Dot7) {
+    if (_cache.style.is3Dot7) {
       _pushIndent3Dot7(Indent.expression, canCollapse: true);
     } else {
       pushIndent(Indent.controlFlowClause);
@@ -362,7 +362,7 @@ final class CodeWriter {
 
     // See if we can immediately bind it based on the page width and the piece's
     // contents.
-    if (isUnsolved) {
+    if (isUnsolved && !_cache.style.pinStateByPageWidthBeforeSolving) {
       // If the solution doesn't bind the piece already, we may be able to
       // eagerly bind it to a state knowing just the page width (minus any
       // leading indentation). If so, do that now. We do that here instead of
@@ -398,7 +398,7 @@ final class CodeWriter {
       );
 
       bool invalid;
-      if (_cache.is3Dot7) {
+      if (_cache.style.is3Dot7) {
         // If the child must be inline, then invalidate because we know it
         // contains some kind of newline.
         // TODO(rnystrom): It would be better if this logic wasn't different for
--- a/lib/src/back_end/solution.dart
+++ b/lib/src/back_end/solution.dart
@@ -146,6 +146,11 @@ final class Solution implements Comparable<Solution> {
   ///
   /// If it can, binds the piece to that state in this solution and returns
   /// `true`. Otherwise returns `false`.
+  ///
+  /// This code has a bug, but is kept for older language versions to avoid
+  /// unexpected formatting changes.
+  ///
+  /// See: https://github.com/dart-lang/dart_style/issues/1847
   bool tryBindByPageWidth(Piece piece, int pageWidth) {
     if (piece.fixedStateForPageWidth(pageWidth) case var state?) {
       _bind(piece, state);
--- a/lib/src/back_end/solution_cache.dart
+++ b/lib/src/back_end/solution_cache.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import '../front_end/formatting_style.dart';
 import '../piece/piece.dart';
 import 'solution.dart';
 import 'solver.dart';
@@ -23,12 +24,11 @@ import 'solver.dart';
 /// indentation. When that happens, sharing this cache allows us to reuse that
 /// cached subtree Solution.
 final class SolutionCache {
-  /// Whether this cache and all solutions in it use the 3.7 style solver.
-  final bool is3Dot7;
+  final FormattingStyle style;
 
   final _cache = <_Key, Solution>{};
 
-  SolutionCache({required this.is3Dot7});
+  SolutionCache(this.style);
 
   /// Returns a previously cached solution for formatting [root] with leading
   /// [indent] (and [subsequentIndent] for lines after the first) or produces a
--- a/lib/src/back_end/solver.dart
+++ b/lib/src/back_end/solver.dart
@@ -176,8 +176,10 @@ final class Solver {
     if (debug.traceSolver) {
       debug.unindent();
       debug.log(debug.bold('Solved $root to $best'));
-      debug.log(best.code.toDebugString());
-      debug.log('');
+      if (debug.traceSolverShowCode) {
+        debug.log(best.code.toDebugString());
+        debug.log('');
+      }
     }
 
     return best;
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -750,7 +750,7 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
       this,
       node.metadata,
       [node.extensionKeyword, node.typeKeyword],
-      namePart: node.primaryConstructor,
+      namePart: node.namePart,
       implementsClause: node.implementsClause,
     );
     builder.buildClassBody(node.body);
--- a/lib/src/front_end/formatting_style.dart
+++ b/lib/src/front_end/formatting_style.dart
@@ -85,6 +85,18 @@ final class FormattingStyle {
   /// different categories: `dart:`, `package:`, or relative.
   bool get separateDirectiveSections => _languageVersion >= _version3Dot13;
 
+  /// Whether to try to figure out a piece's state based on the page width
+  /// before running the solver or during.
+  ///
+  /// Initially, this ran during solving but that leads to some subtle bugs in
+  /// the solver. Performing it before solving is less effective for performance
+  /// but avoids those bugs.
+  ///
+  /// We language version this even though the old logic was never correct to
+  /// minimize unexpected churn.
+  bool get pinStateByPageWidthBeforeSolving =>
+      _languageVersion >= _version3Dot13;
+
   /// Whether there is a trailing comma at the end of the list delimited by
   /// [rightBracket] which should be preserved by this style.
   bool preserveTrailingCommaBefore(Token rightBracket) =>
--- a/lib/src/front_end/piece_writer.dart
+++ b/lib/src/front_end/piece_writer.dart
@@ -456,12 +456,16 @@ final class PieceWriter {
 
     Profile.begin('PieceWriter.finish() format piece tree');
 
-    var cache = SolutionCache(is3Dot7: style.is3Dot7);
+    var cache = SolutionCache(style);
     var solver = Solver(
       cache,
       pageWidth: style.pageWidth,
       leadingIndent: style.leadingIndent,
     );
+
+    if (style.pinStateByPageWidthBeforeSolving) {
+      _pinPiecesByPageWidth(rootPiece, style.pageWidth - style.leadingIndent);
+    }
 
     var solution = solver.format(rootPiece);
     var output = solution.code.build(source, style.lineEnding);
@@ -469,6 +473,21 @@ final class PieceWriter {
     Profile.end('PieceWriter.finish() format piece tree');
 
     return output;
+  }
+
+  /// Traverse the piece tree at [rootPiece] and attempt to pin pieces to states
+  /// if the [pageWidth] and the contents of the pieces are enough to
+  /// determine what state the piece will end up in.
+  void _pinPiecesByPageWidth(Piece rootPiece, int pageWidth) {
+    void traverse(Piece piece) {
+      if (piece.fixedStateForPageWidth(pageWidth) case var state?) {
+        piece.pin(state);
+      }
+
+      piece.forEachChild(traverse);
+    }
+
+    traverse(rootPiece);
   }
 
   /// Returns the number of characters past [position] in the source where the
--- a/lib/src/short/source_visitor.dart
+++ b/lib/src/short/source_visitor.dart
@@ -1171,7 +1171,7 @@ final class SourceVisitor extends ThrowingAstVisitor {
     space();
     token(node.typeKeyword);
 
-    visit(node.primaryConstructor, before: space);
+    visit(node.namePart, before: space);
 
     builder.startRule(CombinatorRule());
     visit(node.implementsClause);
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -9,7 +9,7 @@ environment:
   sdk: ^3.11.0
 
 dependencies:
-  analyzer: ^13.0.0
+  analyzer: ^13.1.0
   args: ^2.5.0
   collection: ^1.19.0
   package_config: ^2.1.0
--- a/test/tall/regression/1800/1847.unit
+++ b/test/tall/regression/1800/1847.unit
@@ -1,0 +1,333 @@
+>>>
+main() async {
+  {
+    await MethodChannelContainer()
+        .onMethodChannelInvoke("reportCrash", <String, dynamic>{
+          "time": nowTime,
+          "errorValue": errorName,
+          "reason": reason,
+          "stacktrace": stacktrace,
+        });
+  }
+}
+<<< 3.7
+main() async {
+  {
+    await MethodChannelContainer()
+        .onMethodChannelInvoke("reportCrash", <String, dynamic>{
+          "time": nowTime,
+          "errorValue": errorName,
+          "reason": reason,
+          "stacktrace": stacktrace,
+        });
+  }
+}
+<<< 3.13
+main() async {
+  {
+    await MethodChannelContainer().onMethodChannelInvoke(
+      "reportCrash",
+      <String, dynamic>{
+        "time": nowTime,
+        "errorValue": errorName,
+        "reason": reason,
+        "stacktrace": stacktrace,
+      },
+    );
+  }
+}
+>>> (indent 4) abrevva-3.0.4/lib/abrevva_method_channel.dart
+    final result = await _methodChannel
+        .invokeMethod<Map<dynamic, dynamic>?>('read', {
+          'deviceId': deviceId,
+          'service': service,
+          'characteristic': characteristic,
+          'timeout': timeout,
+        });
+<<< 3.7
+    final result = await _methodChannel
+        .invokeMethod<Map<dynamic, dynamic>?>('read', {
+          'deviceId': deviceId,
+          'service': service,
+          'characteristic': characteristic,
+          'timeout': timeout,
+        });
+<<< 3.13
+    final result = await _methodChannel.invokeMethod<Map<dynamic, dynamic>?>(
+      'read',
+      {
+        'deviceId': deviceId,
+        'service': service,
+        'characteristic': characteristic,
+        'timeout': timeout,
+      },
+    );
+>>> amplify_auth_cognito_dart-0.11.20/lib/src/model/hosted_ui/oauth_parameters.g.dart
+final BuiltSet<OAuthErrorCode> _$values =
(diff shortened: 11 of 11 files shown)


natebosch:
Do we know what's happening in

```diff
diff --git a/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart b/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
index 2898ca9b8..e886c82b6 100644
--- a/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
+++ b/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
@@ -128,15 +128,16 @@ Future<List<SubscriptionEvent<T>>> createObservedEventsGetter<T extends Model>(
   ModelType<T> modelType, {
   required int take,
   required EventType eventType,
-}) => Amplify.DataStore.observe(modelType)
-    .where((event) => event.eventType == eventType)
-    .distinct(
-      (prev, next) =>
-          prev.eventType == next.eventType &&
-          prev.item.modelIdentifier == next.item.modelIdentifier,
-    )
-    .take(take)
-    .toList();
+}) =>
+    Amplify.DataStore.observe(modelType)
+        .where((event) => event.eventType == eventType)
+        .distinct(
+          (prev, next) =>
+              prev.eventType == next.eventType &&
+              prev.item.modelIdentifier == next.item.modelIdentifier,
+        )
+        .take(take)
+        .toList();
```

natebosch:
The overwhelming majority of the diff looks like an improvement to me. Most of places where indentation is changing it's getting smaller. The above hunk stood out to me as doing the opposite.

There's also this unhappy side effect which looks strictly worse to me:

```diff
diff --git a/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart b/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
index 7958ffaaf..fd84ddf86 100644
--- a/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
+++ b/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
@@ -17098,7 +17098,8 @@ class FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse {
   @override
   bool operator ==(Object other) {
     return identical(this, other) ||
-        (other is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
+        (other
+                is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
             (identical(other.status, status) ||
                 const DeepCollectionEquality().equals(other.status, status)) &&
             (identical(other.success, success) ||
```

And these look worse to me

```diff
diff --git a/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart b/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
index 965a5a08e..9583e87b0 100644
--- a/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
+++ b/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
@@ -1560,30 +1560,12 @@ class _SellInOrderDetailsScreenState extends State<SellInOrderDetailsScreen> {
                                                                     MapNavigationHelper.navigateTo(
                                                                       targetLatitude:
                                                                           double.tryParse(
-                                                                            stockHistory.value.productOrderItem?[0].productRefOrValue?.place
-                                                                                ?.firstWhereOrNull(
-                                                                                  (
-                                                                                    element,
-                                                                                  ) =>
-                                                                                      element.role ==
-                                                                                      "pickupLocation",
-                                                                                )
-                                                                                ?.meta
-                                                                                ?.latitude,
+                                                                            stockHistory.value.productOrderItem?[0].productRefOrValue?.place?.firstWhereOrNull((element) => element.role == "pickupLocation")?.meta?.latitude,
                                                                           ) ??
                                                                           0,
                                                                       targetLongitude:
                                                                           double.tryParse(
```


I'll take a closer look at the code changes tomorrow, overall the corpus diff is an improvement so I'd be OK landing even with these changes that I see as regressions, but the latter two in this comment might be worth looking at separately.

natebosch (approved):

munificent:
> Do we know what's happening in
> 
> ```diff
> diff --git a/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart b/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
> index 2898ca9b8..e886c82b6 100644
> --- a/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
> +++ b/amplify_datastore-2.11.0/example/integration_test/utils/test_cloud_synced_model_operation.dart
> @@ -128,15 +128,16 @@ Future<List<SubscriptionEvent<T>>> createObservedEventsGetter<T extends Model>(
>    ModelType<T> modelType, {
>    required int take,
>    required EventType eventType,
> -}) => Amplify.DataStore.observe(modelType)
> -    .where((event) => event.eventType == eventType)
> -    .distinct(
> -      (prev, next) =>
> -          prev.eventType == next.eventType &&
> -          prev.item.modelIdentifier == next.item.modelIdentifier,
> -    )
> -    .take(take)
> -    .toList();
> +}) =>
> +    Amplify.DataStore.observe(modelType)
> +        .where((event) => event.eventType == eventType)
> +        .distinct(
> +          (prev, next) =>
> +              prev.eventType == next.eventType &&
> +              prev.item.modelIdentifier == next.item.modelIdentifier,
> +        )
> +        .take(take)
> +        .toList();
> ```

I did spend some time investigating it. I added a regression test for this funny corner case with a comment so that at least some of the current behavior is pinned down:

```
### TODO(rnystrom): It shouldn't split after `=>`, but there is some subtle bug
### around how overflow is used to order solutions. If we don't take overflow
### into account when comparing solutions, this is fixed, but lots of other
### things break. An unfortunately load-bearing bug.
```

Here's a much longer explanation: The basic idea of the solver is that we start with a solution that has as few splits as possible. Since splitting code never reduces the cost of a solution, that almost implies that we start with a lowest-possible-cost solution. Then we derive new solutions from that initial solution by taking some pieces, binding them to other states, and see how that effects things. The bargain that choosing a state for a piece usually makes is to increase the solution cost but reduce its overflow.

We explore solutions in order of increasing cost, so as soon as we find a solution with no overflow, we know it must be the lowest-possible cost solution that fits. The solver can thus immediately stop on that solution.

To implement this, the core of the solver is a priority queue that yields orders solutions and lets the solver explore them. But what happens if two solutions have the same cost? How are those ordered? I want there to be a total order for solutions so that the priority queue is reliably deterministic. The current answer is:

1. First, order by increasing cost.

2. Otherwise, order by increasing overflow.

3. Otherwise, order by the numeric piece states of bound pieces. There must be some difference here because we avoid enqueuing the same solution (i.e. same set of bound piece states) multiple times.

Step 1 matters, but the other 2 are kind of arbitrary. I honestly don't recall why I added comparing by overflow. Now that I look, the old formatter doesn't do that.

It's the overflow comparison that leads to this weird formatting. If you comment that out, then you get:

```dart
Future<List<SubscriptionEvent<T>>> createObservedEventsGetter<T extends Model>(
  ModelType<T> modelType, {
  required int take,
  required EventType eventType,
}) => Amplify.DataStore.observe(modelType)
    .where((event) => event.eventType == eventType)
    .distinct(
      (prev, next) =>
          prev.eventType == next.eventType &&
          prev.item.modelIdentifier == next.item.modelIdentifier,
    )
    .take(take)
    .toList();
```

Even better, not a single other test fails. So it seems like we *could* just remove that check. However, when I tested that on a big corpus, it does actually lead to a *lot* of changes (though it seems that the majority of them are in generated built_value code). Many are good like:

```dart
// Before:
      expect(
        () =>
            acmeConnectionAccount(client)!
                .createOrderForHttp(identifiers: const []),
        throwsA(isA<AcmeJwsException>()),
      );

// After:
      expect(
        () => acmeConnectionAccount(client)!
            .createOrderForHttp(identifiers: const []),
        throwsA(isA<AcmeJwsException>()),
      );

// Before:
  return predicate(
    (List<SharedTypeView> x) =>
        unorderedEquals(expectedTypes)
            .matches(x.map((t) => t.unwrapTypeView<Type>().type).toList(), {}),
    'interest set $expectedTypes',
  );

  return predicate(
    (List<SharedTypeView> x) => unorderedEquals(expectedTypes)
        .matches(x.map((t) => t.unwrapTypeView<Type>().type).toList(), {}),
    'interest set $expectedTypes',
  );
```

But there are also many more that are sort of neutral like:

```dart
// Before:
  Map<String, Object?> get _debugFields =>
      super._debugFields..['conditionTrue'] = _conditionTrue;

// After:
  Map<String, Object?> get _debugFields => super._debugFields
    ..['conditionTrue'] = _conditionTrue;

// Before:
            ElevatedButton(
              onPressed: () =>
                  Navigator.of(context).pushReplacementNamed('/routeB'),
              child: const Text('Goto Route B'),
            ),

// After:
            ElevatedButton(
              onPressed: () => Navigator.of(context)
                  .pushReplacementNamed('/routeB'),
              child: const Text('Goto Route B'),
            ),

// Before:
  UsernameConfigType get usernameType =>
      AuthenticatorFormState.of(context).usernameType;

// After:
  UsernameConfigType get usernameType => AuthenticatorFormState.of(context)
      .usernameType;
```

I believe the underlying issue is that some pieces have states whose split cost is zero. I ended up doing that in some places to get the formatter to pick one preferred state over another without instead increasing the cost of the other states because that would interfere with the surrounding formatting choices.

Most of this all stems from trying to balance the hanging mobile of formatting assignments (anything with `=`, `=>`, or `:`) whose RHS is some combination of a method call chain, argument list, or collection literal. I never came up with an elegant way to handle those that felt like the right formatting just composed naturally from the constituent parts.

Either way, I felt like also removing the overflow check would increase the amount of churn caused by this change too much. And I already spent so much time digging into this issue that I didn't want to rathole forever on it. I would like to revisit it at some point, but the window of getting changes into 3.13 is closing.

kallentu (approved):

munificent:
I dug into this one:

> ```diff
> diff --git a/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart b/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
> index 7958ffaaf..fd84ddf86 100644
> --- a/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
> +++ b/rentalworks-8.0.0/lib/generated_code/administrator.swagger.dart
> @@ -17098,7 +17098,8 @@ class FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse {
>    @override
>    bool operator ==(Object other) {
>      return identical(this, other) ||
> -        (other is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
> +        (other
> +                is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
>              (identical(other.status, status) ||
>                  const DeepCollectionEquality().equals(other.status, status)) &&
>              (identical(other.success, success) ||
> ```

The change here is clearly worse. Here's a simplified repro is:

```dart
// Before:
main() {
  {
    return identical_this__other_ ||
        (other is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
            (identical_other_msg__msg_ ||
                const_DeepCollectionEquality___equals(other_msg, msg)));
  }
}

// After:
{
  {
    return identical_this__other_ ||
        (other
                is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
            (identical_other_msg__msg_ ||
                const_DeepCollectionEquality___equals(other_msg, msg)));
  }
}
```

The split on `is` is completely unnecessary.

As far as I can tell, the old optimization of eagerly binding pieces based on the size of their contents during solving was masking a different bug around how the solver calculates expand pieces. This PR replaces that optimization with a new one that no longer masks the bug. The new optimization in this PR doesn't itself cause the bug. If I comment out the eager piece pinning completely (both the old and new ways), then the bug still happens.

#### Expand piece set

A critical thing the solver does for performance is limit which pieces it tries to expand (choose split states for) when generating new solutions from an existing partial solution. We want the set of expand pieces to be as small as possible to keep down the branchiness of the solution tree. Otherwise, solving gets too exponential.

An observation I made many years ago is that we only really need to worry about unbound pieces on the first overflowing line. Pieces that are entirely contained in non-overflowing lines before the first overflowing line don't cause any problems so can be left alone. Pieces on lines *after* the first overflowing line will likely have their indentation or other context changed when a piece on the first overflowing line is split differently, so we can defer worrying about those until after the solver explores splitting pieces on the first overflowing line.

So when a solution is being formatted, it keeps track of which pieces are writing code on overflowing lines. Once it finds the first overflowing line that contains some unbound pieces, it records those and then stops collecting more.

It looks like that observation is only *mostly* correct. In this code, the first overflowing line is:

```
        (other is FwStandardModulesAdministratorCreateNewSystemCreateNewSystemResponse &&
```

We do find an unbound piece on that line, for the `is` expression. So the only solution in the expand set from the initial empty solution is one that splits on the `is`. Once we do that, only later do we finally figure to split at `identical_other_msg__msg_ ||` which actually reduces the overflow. But since that was derived from the prior solution that splits on `is`, we keep that pointless split after `is`. :(

This problem has been around unnoticed forever (even the old formatter had a similar optimization), so it's probably rare that it actually gets triggered. I'll think more about whether there's something smarter we can do here. But for now, I think cases like this are rare enough to accept. (And in particular, it may be that the only way the expand piece set pruning optimization does the wrong thing is in cases where the code can't be made to fit? I'm not sure.)

munificent:
```diff
iff --git a/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart b/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
index 965a5a08e..9583e87b0 100644
--- a/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
+++ b/ventas_dealer-0.0.124/lib/view/stock_order/sellin_order_details_screen.dart
@@ -1560,30 +1560,12 @@ class _SellInOrderDetailsScreenState extends State<SellInOrderDetailsScreen> {
                                                                     MapNavigationHelper.navigateTo(
                                                                       targetLatitude:
                                                                           double.tryParse(
-                                                                            stockHistory.value.productOrderItem?[0].productRefOrValue?.place
-                                                                                ?.firstWhereOrNull(
-                                                                                  (
-                                                                                    element,
-                                                                                  ) =>
-                                                                                      element.role ==
-                                                                                      "pickupLocation",
-                                                                                )
-                                                                                ?.meta
-                                                                                ?.latitude,
+                                                                            stockHistory.value.productOrderItem?[0].productRefOrValue?.place?.firstWhereOrNull((element) => element.role == "pickupLocation")?.meta?.latitude,
                                                                           ) ??
                                                                           0,
                                                                       targetLongitude:
                                                                           double.tryParse(
```

Oh, this is a funny one.

Prior to this PR, the old optimization to force pieces to split by page width would force the call chain after `stockHistory` to split because its contents don't fit the tiny remaining page width since the old optimization knew the surrounding indentation.

The new optimization lacks that context. It only knows the overall page width of 80. Also, because of the `firstWhereOrNull()` call which contains a lambda which might block format, the new optimization doesn't eagerly force the call chain to split. So instead of relies on the solver to figure out whether to split it or not.

Comically, the solver correctly figures out that splitting that call chain doesn't actually help reduce overflow at all. This code is indented so far that the entire split call chain is past the page width. Here's what it looks like with a bunch of leading indentation stripped off to be easier to read:

```dart
// Before:
    MapNavigationHelper.navigateTo(
      targetLatitude:
          double.tryParse(
//              | <- 80 column page width
            stockHistory.value.productOrderItem?[0].productRefOrValue?.place
                ?.firstWhereOrNull(
                  (
                    element,
                  ) =>
                      element.role ==
                      "pickupLocation",
                )
                ?.meta
                ?.latitude,
          ) ??
          0,
      targetLongitude:
          double.tryParse(

// After:
    MapNavigationHelper.navigateTo(
      targetLatitude:
          double.tryParse(
//              | <- 80 column page width
            stockHistory.value.productOrderItem?[0].productRefOrValue?.place?.firstWhereOrNull((element) => element.role == "pickupLocation")?.meta?.latitude,
          ) ??
          0,
      targetLongitude:
          double.tryParse(
```

So the formatter correctly concludes that splitting the call chain doesn't actually reduce the number of overflowing characters, so it picks the shorter solution.

Really, this is just a weird example because the code is so pathologically indented. (In the full source file, this one expression is thousands of lines long.)

munificent:
Thank you for reviewing the diffs so thoroughly! It encourages me to really dig into them and helps keep me honest about understanding what the formatter is doing.

natebosch:
> doesn't actually reduce the number of overflowing characters, so it picks the shorter solution.

Should (or have) we consider making the heuristic weight long overflows as worse than multiple short overflows? The case is extreme, but it's also a pattern (deeply nested widget trees) that I suspect could be more prevalent in closed source Flutter widgets than in the corpus we can scrape. I think we should wait and see if we get feedback on this before making any changes, but I'll be curious to see if anyone brings it up.

PR #1857, opened 2026-06-09
Merge commit for Dart SDK main branch roll

munificent:
Merge commit with d7b366261228340b5ef99aa4152891e8ac7f34a8 (from #1855) as a parent so that that commit is reachable from main. That commit is rolled into the Dart SDK main branch, so we need to make sure it doesn't get GCed by Git.


munificent:
Oops, looks like there's a mistake I need to fix first (#1858). Closing this. I'll create a new PR once that's fixed.

PR #1858, opened 2026-06-09
Fix io_file_system_test.dart.

munificent:
I added a new mandatory named parameter to `readAnalysisOptions()` and somehow I missed one callsite in a test.

For this test, no file reads should fail, so we just provide it with a closure that fails the test if invoked.

### DIFF
--- a/test/analysis_options/io_file_system_test.dart
+++ b/test/analysis_options/io_file_system_test.dart
@@ -162,7 +162,12 @@ void main() {
       Future<String?> failingResolver(Uri uri) async => null;
 
       expect(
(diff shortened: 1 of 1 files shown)


munificent:
Ugh, after all of the shenanigans to create the roll CLs merge commits, I find a dumb one-line mistake. I'll update the [roll CL](https://dart-review.googlesource.com/c/sdk/+/510222) and [merge commit](https://github.com/dart-lang/dart_style/pull/1857) once this lands.

natebosch (approved):

PR #1860, opened 2026-06-10
Support a trailing comma in extension type representation clauses.

### LINKED ISSUE #1845: Put trailing commas in split representation clauses in 3.13
When [extension types](https://github.com/dart-lang/language/blob/main/accepted/3.3/extension-types/feature-specification.md) were first added to the language, they had their own unique grammar for the representation clause that was distinct from the parameter list grammar:

```
<representationDeclaration> ::=
  ('.' <identifierOrNew>)? '(' <metadata> <type> <identifier> ')'
```

When we added [primary constructors](https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md), we reworked the extension type grammar to now also use the same primary constructor syntax (with some semantic restrictions around arity):

```
<extensionTypeDeclaration> ::= // Modified rule.
     'extension' 'type' <primaryConstructor> <interfaces>?
     <memberedDeclarationBody>;
```

Perhaps unexpectedly, this does change the syntax. The original representation clause grammar does not allow a trailing comma before the `)`. A primary constructor does. The formatter had to [carefully avoid putting a trailing comma in a representation clause](https://github.com/dart-lang/dart_style/blob/main/lib/src/front_end/ast_node_visitor.dart#L816-L818) to avoid generating invalid syntax.

In Dart 3.13, that trailing comma is now allowed and for consistency with other comma-delimited constructors, we should probably put a trailing comma there too. Obviously, this must be language-versioned so that a trailing comma is only written in 3.13 or later files.

munificent:
In Dart 3.13, we changed the grammar for extension types to match primary constructors (with restrictions). A side effect of that is that where the old grammar (inadvertently) didn't allow a trailing comma in a split representation clause, the new grammar does (because it's just a formal parameter list now).

So in Dart 3.13 and later libraries, the formatter will write a trailing comma there, to be consistent with most other places in the language were we write a trailing comma.

Fix #1845.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,6 +226,9 @@
   Note how a blank line was added above `// Comment 2.` but not `// Comment 1.`.
   Now, it won't add a blank line before the last comment.
 
+* Write a trailing comma in split extension type representation clauses when in
+  a library whose language version allows it (#1845).
+
 ### Internal changes
 
 * Require `analyzer: ^13.1.0`.
--- a/lib/src/front_end/ast_node_visitor.dart
+++ b/lib/src/front_end/ast_node_visitor.dart
@@ -811,9 +811,11 @@ final class AstNodeVisitor extends ThrowingAstVisitor<void> with PieceFactory {
       return;
     }
 
-    // If all parameters are optional, put the `[` or `{` right after `(`.
+    // Extension type representation clauses prior to Dart 3.13 don't allow a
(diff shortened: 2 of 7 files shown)


munificent:
Here's an easy one. :)

kallentu (approved):

natebosch (approved):

PR #1871, opened 2026-07-08
Migrate off the grinder package and test that versions match.

munificent:
The formatter supports `--version` to print its version. Since the shipped package doesn't include the pubspec, we have a constant inside the codebase that mirrors the pubspec's version number.

Keeping that in sync is an annoying source of bugs. I've partially alleviated that by having a grinder script to update the version numbers in parallel.

This PR migrates that code to standalone scripts to remove the dependency on grinder (which I hear is no longer maintained).

I also updated the pre-publish script to validate that the CHANGELOG has an entry for the new version.

And I added a test that will fail if the two version numbers ever get out of sync.


lib/src/tool_utils.dart
natebosch, line 1: Why put this in `lib` if it's only used locally?

tool/ship.dart
```diff
@@ -0,0 +1,79 @@
...
+      print('Unexpected pre-release version $version. Don\'t ship!');
+      exit(1);
+  }
+
+  // Check the CHANGELOG.
+  var changelog = File('CHANGELOG.md').readAsLinesSync();
+  if (!changelog.contains('## $version')) {
+    print('Missing entry in CHANGELOG.md for $version. Don\'t ship!');
+    exit(1);
+  }
```
natebosch: It might also be worth trying to migrate to the team standard `package:firehose`

https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose

tool/ship.dart
```diff
@@ -0,0 +1,79 @@
...
+  String executable,
+  List<String> arguments, [
+  String? message,
+]) async {
+  if (message != null) print(message);
+  var process = await Process.start(executable, arguments);
+
+  // Pipe output so the user can see progress.
+  var stdoutFuture = stdout.addStream(process.stdout);
+  var stderrFuture = stderr.addStream(process.stderr);
```
natebosch:
```suggestion
  var process = await Process.start(executable, arguments, mode: .inheritStdio);
```

natebosch (approved):

lib/src/tool_utils.dart
munificent: Is there a better place to put it? It felt weird dropping it in `tool/` because it's not an entrypoint you can execute.

tool/ship.dart
munificent: I don't quite understand how it all works, but I believe dart_style is already using firehose.

Now that you mention it, I suppose that means I don't need to validate the CHANGELOG here, but it is nice having a little validation step before you commit and upload a PR.

tool/ship.dart
munificent: Nice, done.

lib/src/tool_utils.dart
natebosch: I don't think that's stopped us in the past but I don't have a specific example in mind. I think we've also used the `src/` prefix to distinguish entrypoints from non-`lib/` utility code.

One reason not to use `lib` is we would _not_ want to add new non-dev dependencies in the public if we find the need to add an `import` in this file in the future, but that's what an author is likely to do by default.

munificent: > I think we've also used the src/ prefix to distinguish entrypoints from non-lib/ utility code.

So `tool/src/tool_utils.dart`? That's a good idea. I clicked "merge" perhaps a little too eagerly, but I'll do that: https://github.com/dart-lang/dart_style/pull/1872

> One reason not to use `lib` is we would _not_ want to add new non-dev dependencies in the public if we find the need to add an `import` in this file in the future, but that's what an author is likely to do by default.

That's a good point, but don't you run into a similar problem with the `lib/src/testing/` pattern?

natebosch: > but don't you run into a similar problem with the `lib/src/testing/` pattern?

I don't think we recommend that pattern either. `test/utils.dart` and `test/src/utils.dart` are both patterns I think I've seen. I'd think a `lib/src/testing.dart` would typically exist if `lib/testing.dart` exports stuff from it.

munificent: Oh, TIL. (dart_style has a `lib/src/testing/` directory. I'll move that at some point.)

PR #1878, opened 2026-07-17
Make DartFormatter.lineEnding immutable.

### LINKED ISSUE #1337: DartFormatter class caches line endings from the first format which might produce incorrect results for subsequent formats
I'm not sure if this is strictly a bug, but it feels incorrect. I was investigating an issue at https://github.com/Dart-Code/Dart-Code/issues/4865 where sometimes lots of blank lines were being added when formatting documents.

It seems like there may be two issues:

1. The formatter changing a file using `\n` to `\r\n`
2. VS Code showing `\r\n` as if it's two newlines if the file is set to `LF` mode but contains CRLF.

The reason the first was occurring is that in the analysis server we are currently reusing the `DartFormatter` instance. This class caches the line ending on first format if it's not explicitly provided. This means if you format a document that uses CRLF and then one that uses LF, it will output CRLF for the second file:

```
final formatter = DartFormatter();
print(jsonEncode(formatter.format('int a;\r\n'))); // Outputs "int a;\r\n"
print(jsonEncode(formatter.format('int b;\n'))); // Outputs "int b;\r\n" but expected "int b;\n"
```

I'm going to remove the reuse of this instance from the server (we were discussing this because of language versions anyway), but I thought it was worth raising this because it's not clear if this is expected behaviour.

munificent:
This property was always intended to be semantically immutable but since the early days, it was exposed as a public writable field.

This retains the setter to avoid a compile-time breaking change in the off chance there is some code calling the setter, but invoking it does nothing.

This also fixes the weird behavior where if you omit a line ending (which I'm guessing all users do), it infers the line ending from the first format() call and then persists that to later calls. This is a weird subtle kind of mutability. In practice, it's harmless because users rarely reuse a DartFormatter instance to format code with different line endings. But it never felt like a good design.

This is technically a breaking change, but in a corpus of 2,000 pub packages, I couldn't find a single reference to an identifier named "lineEnding", so I suspect no code in the wild will actually break, which is why I only bumped the patch version.

Fix #1337.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,17 @@
   `lineEnding` always returns the value you passed in the constructor.
 
   This means you can use the same `DartFormatter` instance to format code with
-  different line endings and the first `format()` call won't affect the results
-  of later calls (#1337).
+  different line endings by not providing an explicit line ending. Each call
+  to `format()` will infer a separate line ending for only that call's code
+  (#1337).
 
   The setter for `lineEnding` is still available but is marked deprecated and
   does nothing. This is technically a **breaking change**, but I believe no code
-  in the wild actually reads the `lineEnding` field so it is harmless. In a
-  future major version release, the setter will be removed.
+  in the wild is affected by it. To be affected by the setter doing nothing,
+  code would have to call the setter and then call `format()`, or call the
+  `lineEnding` getter. I haven't found any code on pub that sets or gets
+  `lineEnding`, so I suspect this change is harmless. In a future major version
+  release, the setter will be removed.
 
 ## 3.1.12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+## 3.1.13-wip
+
+### API changes
+
+* `DartFormatter.lineEnding` is no longer mutable. If no line ending is
+  provided, then a line ending is inferred from the source as usual, but the
+  result is not stored back in the `lineEnding` field. Instead, accessing
+  `lineEnding` always returns the value you passed in the constructor.
+
+  This means you can use the same `DartFormatter` instance to format code with
+  different line endings and the first `format()` call won't affect the results
+  of later calls (#1337).
+
+  The setter for `lineEnding` is still available but is marked deprecated and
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+### API changes
+
+* `DartFormatter.lineEnding` is no longer mutable. If no line ending is
+  provided, then a line ending is inferred from the source as usual, but the
+  result is not stored back in the `lineEnding` field. Instead, accessing
+  `lineEnding` always returns the value you passed in the constructor.
+
+  This means you can use the same `DartFormatter` instance to format code with
+  different line endings and the first `format()` call won't affect the results
+  of later calls (#1337).
```
natebosch: Is this accurate? It's true that `format()` calls no longer impact the results of later calls, but I don't think it's true that you can use the same `DartFormatter` instance to format code with different line endings when the main change in the PR is making it immutable.

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+
+  The setter for `lineEnding` is still available but is marked deprecated and
+  does nothing. This is technically a **breaking change**, but I believe no code
+  in the wild actually reads the `lineEnding` field so it is harmless. In a
```
natebosch: Did you mean "writes" instead of "reads"? Reads shouldn't be impacted AFAICT.

natebosch (approved):

CHANGELOG.md
munificent: If you create an instance with `null` for the `lineEnding` (which I think basically everyone does) then you can reuse it because each `format()` call will infer a line ending independently of other calls. If you were to create a `DartFormatter` with an explicit line ending, then every call to `format()` will use that line ending, which I think is what you'd expect.

Reworded to make this clearer.

CHANGELOG.md
munificent: I meant reads but this wasn't written clearly. The only way you can tell that the setter now does nothing is to do something *after* that that would be affected by the setter's behavior. That would mean either calling the `lineEnding` *getter* and observing that the write was ignored or calling `format()` and seeing that it didn't use the new line ending.

Reworded to make that clearer.

PR #1892, opened 2026-08-20
Fix version_test to handle being run from within the Dart SDK.

munificent:
Normally, when you run a package's tests, the current working directory is the root directory of the package. Without using mirrors, tests don't have a way to locate a file relative to the test file itself. Instead, most just assume the CWD is the package root and find files relative to that.

However, when the Dart SDK's test runner runs tests, the CWD is usually the root of the SDK itself and not the package's root. I put some time into having the test runner set the working directory to the package root when it runs a package's tests, but that breaks a lot of other stuff: Currently in the test runner, the executable and arguments are all relative to the SDK root. I don't think it's a good idea to change all of those.

Instead, in version_test, I just do a little hack to try to detect if it looks like the test is being run by the test runner and tweak the path if so.

### DIFF
--- a/test/cli/version_test.dart
+++ b/test/cli/version_test.dart
@@ -5,6 +5,7 @@
 import 'dart:io';
 
 import 'package:dart_style/src/cli/formatter_options.dart';
+import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:test/test.dart';
 import 'package:yaml/yaml.dart';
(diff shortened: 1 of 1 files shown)


munificent:
This is blocking rolling the enum crash and style fixes into the SDK (and thus into the next stable patch release) (https://dart-review.googlesource.com/c/sdk/+/539080), so if you can take a look relatively quickly, I would appreciate it. Thanks!

kallentu (approved):

natebosch (approved):

PR #1905, opened 2026-09-17
Optimize piece state lookups during formatting.

davidmorgan:
_Provenance note: this issue and fix were discovered during unrelated work driven by me (Morgan) but done by an agent; the fix is by the agent, not by me. It was free to produce, please **reject it** if it is not useful, no explanation needed. The performance improvement is very likely real, this was discovered during wider benchmarking of dartfmt._

Most pieces in a piece tree remain unbound in State.unsplit throughout formatting. Previously, Solution.pieceStateIfBound walked the entire _pieceStates linked list from head to tail for every unbound piece, and CodeWriter._formatInline queried the solution multiple times per piece and once again for its parent piece.

- Add Piece.boundInAnySolution to short-circuit Solution.pieceStateIfBound, Solution.pieceState, and Solution.isBound in O(1) time for pieces that have never been bound in any solution.
- Store the current piece state in _FormatState so child pieces can look up their parent piece state in O(1) time without searching the solution.
- Consolidate the bound state check in CodeWriter._formatInline into a single lookup.

Speeds up formatting by 3% to 10% across the dart_style benchmark suite.


munificent:
This might be useful, but having a mutable field stored in `Piece` which is set during solving and interacts with other solvers is very architecturally spooky. The piece tree, once it gets to solving, should be considered completely immutable so that solvers can share it and work independently without there being weird artifacts left from previous solve attempts.

It might do the right thing and effectively allow sharing work between solves, but I'm not smart enough to reason about it.

